PR #5, opened 2026-03-19
feat: widen signer types for Keychain compatibility

amilz:
## Summary
- Replace `KeyPairSigner` with `MessagePartialSigner` for voucher signing (authorizers, `signVoucher`)
- Replace `TransactionSigner` with `TransactionPartialSigner` for server fee payer co-signing
- Properly decode transaction lifetime from wire bytes using `getCompiledTransactionMessageCodec` + `getTransactionLifetimeConstraintFromCompiledTransactionMessage` instead of type casts
- Validate signer return value and use `isTransactionPartialSigner` runtime guard

[Solana Keychain](https://github.com/solana-foundation/solana-keychain)'s `SolanaSigner` (Vault, Privy, Turnkey, AWS KMS, etc.) now works as a drop-in wherever signers are accepted. `KeyPairSigner` remains compatible since it extends both `MessagePartialSigner` and `TransactionPartialSigner`.

## Test plan
- [x] 39/39 unit tests pass
- [x] Typecheck clean (pre-existing zod inference errors only)
- [x] Demo app tested with fee payer co-signing

Closes TOO-195

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Solana payment method for the [Machine Payments Protocol](https://mpp.dev).
 - [Swig](https://build.onswig.com) smart wallet integration for on-chain spend limits
 
 **General**
-- Works with [ConnectorKit](https://www.connectorkit.dev) and `@solana/kit` keypair signers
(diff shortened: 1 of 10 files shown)


lgalabru (commented):
Beautiful, thanks for connecting these dots ❤️

PR #34, opened 2026-04-27
feat: :sparkles: payment-channels program-wire layer

TDemeco:
This PR lays the foundation for payment-channels session support in the SDK. Scope is deliberately narrow: stand up the feature flag that fences off the pre-v1 flow, pull in the Solana and upstream crates the session module will need, and carve out the SDK-owned program boundary (the code that speaks on-chain program bytes) along with the protocol wire types that cross the network. No runtime logic yet; those land in later parts.

The upstream `Moonsong-Labs/solana-payment-channels` program is consumed as a black box: pinned by full 40-char rev (currently `5e35626`, post-open-ix), with the Codama-generated `payment_channels_client` crate as a git dependency. The SDK owns only the hand-written byte-contract code for the three contracts the program speaks (voucher payload, ed25519 precompile instruction, Channel PDA) plus the off-chain orchestration helpers around them. The program source itself stays read-only reference, never vendored.

## Program boundary scaffold

A new `pull-mode` cargo feature gates the existing `multi_delegator` client behind a compile-time toggle; that pre-v1 pull-style flow still compiles when the feature is enabled but is no longer part of the supported v1 surface. The `rust/src/program/payment_channels/` module tree gets the six submodules (`voucher`, `state`, `ix`, `verify`, `splits_ext`, plus the Codama client consumed at call sites) that hold the byte contracts and verify helpers. Dependencies pulled in at this layer: `solana-client` for non-blocking RPC, `solana-sdk-ids`/`solana-ed25519-program`/`solana-compute-budget-interface` as atomic crates, `spl-token` plus `spl-token-2022` (the latter only for the rejection check, with default features kept on because v8 fails to compile without them), `spl-associated-token-account-client`, plus `borsh` and `solana-address` (with the `borsh` feature) as direct deps that were transitive before. The upstream pin's `edition = "2024"` raises the workspace rustc floor to 1.85.

## Program binary fetch + hash pinning

`scripts/fetch-program-binary.sh` builds the upstream program at the pinned SHA (read directly from `cargo metadata`, so `Cargo.lock` is the single source of truth) and verifies the resulting `.so` against a committed sha256. Toolchain pin is `REQUIRED_SOLANA_VERSION="3.1.13"`; two builds at different toolchain versions produce different bytes, so the hash check is meaningful only with the toolchain pin in place. Bootstrap is explicit (`--bootstrap`, recorded by the first maintainer at a given upstream SHA); empty hash files fail loud rather than auto-blessing whatever the first builder happened to have installed. Three justfile recipes cover the full flow: `fetch-program-binary`, `fetch-program-binary-bootstrap`, `verify-binary-hash`.

## The three byte contracts

`build_signed_payload(channel_id, cumulative_amount, expires_at) -> [u8; 48]` produces the voucher payload an authorized signer signs to authorize a `settle`. The body delegates to `borsh::to_vec` over the upstream `VoucherArgs` struct rather than hand-rolling offsets, so a future field reorder upstream moves both the program and the SDK in lockstep. The wire convention `expires_at == 0 means no expiry` matches what the program's clock comparison skips on.

`build_verify_ix(channel_id, cumulative_amount, expires_at, signer)` composes the 160-byte ed25519 precompile instruction by calling `solana_ed25519_program::new_ed25519_instruction_with_signature` (the canonical helper every Solana client uses for that layout) on top of `build_signed_payload`. The signer is a `VoucherSigner` trait with a fallible `sign_voucher_payload(&self, &[u8; 48]) -> Result<[u8; 64], SignerError>` so production paths through KMS/HSM/custodial wallets fit the same shape as in-process `ed25519_dalek::SigningKey`. `verify_voucher_signature(pubkey, signature, payload)` runs `verify_strict` off-chain so server code can fail fast before paying network fees on a doomed transaction.

`find_channel_pda(payer, payee, mint, authorized_signer, salt: u64, program_id) -> (Pubkey, u8)` derives the Channel PDA using the seeds the program consumes: literal `b"channel"`, then payer, payee, mint, authorized_signer, and salt as 8 little-endian bytes. Four L0 regression tests pin determinism, salt sensitivity, signer sensitivity, and canonical-bump round-trip via `Pubkey::create_program_address`.

## Channel state reader and RPC verify helpers

`ChannelView` wraps the upstream Codama-generated `Channel` struct so the rest of the SDK works in `solana_pubkey::Pubkey` (3.x) instead of upstream's `solana_address::Address` (2.x), centralizing the type bridge in one place rather than spreading it across every server handler.

A `Mismatch` enum carries twelve typed per-field variants (Deposit, Settled, Bump, Version, Status, GracePeriod, ClosureStartedAt, Payer, Payee, AuthorizedSigner, Mint, ClosureNotStarted) so callers can pattern-match on a `Status` mismatch differently from a `Mint` mismatch. The outer `VerifyError` carries `NotFound`, `Tombstoned`, `NotTombstoned { data_len }`, `Mismatch(#[from] Mismatch)`, and the transport catchalls `Rpc(String)` and `Decode(String)`.

Six async verify helpers cover the lifecycle: `verify_open` (with an `ExpectedOpenState` struct), `verify_topup`, `verify_settled`, `verify_closing`, plus the strict/broad pair `verify_tombstoned` and `verify_finalized_or_absent`. The strict tombstone helper rejects absence as `NotFound`; the broad helper accepts both tombstone and absence and is meant for the startup-recovery path with independent store evidence (the doc comment makes the caller contract explicit). RPC fetch uses `get_ui_account_with_config` (the deprecated `get_account_with_config` is gone in `solana-client` 4.x) plus `UiAccountData::decode`. AccountNotFound is matched on JSON-RPC code `-32004`, not the error string, so it stays stable across providers.

## L1 integration oracle (litesvm)

Two integration test files at `rust/tests/`. `session_l1_ed25519_oracle.rs` ships two LIVE tests: a positive path that submits a `build_verify_ix`-built transaction against the native ed25519 precompile inside litesvm, and a negative test that mutates `data[48]` (the first byte of the signature in the canonical 160-byte layout) and asserts rejection. `session_l1_open_oracle.rs` ships one LIVE test that drives the full open ix through the upstream Codama `OpenBuilder`, including SPL mint setup, payer ATA, channel PDA derivation, and the 13-account list, and asserts the channel account lands at the SDK-computed PDA with the canonical bump in `data[2]`. Both run today against the bootstrapped program binary.

`litesvm` and `litesvm-token` land as dev-deps; the `solana-pubkey-v2` aliased dev-dep bridges the cross-major boundary where `spl-associated-token-account-client` 2.x speaks `solana-pubkey` 2.x while the SDK uses 3.x. Two justfile recipes (`l1-oracle-ed25519` and `l1-oracle-open`) keep each path independently runnable.

## Wire-protocol narrowing

`rust/src/protocol/intents/session.rs` is rewritten from a sprawling speculative surface into a strict spec'd public API. `VoucherData`, `SignedVoucher`, `SigType`, `SessionRequest`, `MethodDetails`, `BpsSplit`, `Split` (a `#[non_exhaustive]` typed twin of `BpsSplit`), `wire_to_typed` / `typed_to_wire` boundary helpers, and the `SessionAction` envelope (`Open`, `Voucher`, `TopUp`, `Close` variants tagged `"action"` and rendered camelCase). `SessionAction::Voucher` flattens its inner `SignedVoucher` directly beside the action tag; `SessionAction::TopUp` serializes as `"action":"topUp"` (capital U) to match the upstream protocol doc.

`wire_to_typed` rejects three classes of malformed input at the wire boundary: non-canonical base58, wrong-length decodes (anything other than 32 bytes after decode), and `share_bps > 10_000`. The bps cap fits the function's existing "reject malformed wire input" role and spares every downstream caller from re-validating.

Two draft modules (`server/session.rs`, `client/session.rs`) get a three-line gating header (`#![allow(...)]` plus `#![cfg(any())]`) so they compile out under all configurations while their bodies stay in the diff history as reference for the eventual rewrite.

## Wire-shape contract pin

`rust/tests/session_l0_wire.rs` (~570 lines, 28 tests) locks down the wire contract from breakage in either direction. Ten roundtrip tests catch field drops and broken `serde(deserialize_with = ...)` attributes; twelve literal JSON shape fixtures (each comparing `serde_json::to_value` against a `serde_json::json!({...})` literal) catch field renames that roundtrips alone would not see; two JCS+base64url roundtrips drive `serialize_request` / `deserialize_request` on `SessionRequest` and every `SessionAction` variant. Three negative tests assert that an unknown `action` tag, a missing `action` tag, and a `VoucherData` missing a required field all fail to deserialize.

Every camelCase key the SDK emits has an explicit fixture (`channelId`, `cumulativeAmount`, `expiresAt`, `signatureType`, `shareBps`, `channelProgram`, `distributionSplits`, `minimumDeposit`, `challengeId`, `authorizedSigner`, `depositAmount`, `additionalAmount`, `feePayerKey`, `minVoucherDelta`, `ttlSeconds`, `gracePeriodSeconds`), and the variant tag values `"open"`, `"voucher"`, `"topUp"`, `"close"` are pinned. A future refactor that accidentally renames any of these fails at least one test.

## Implementation considerations worth flagging

**Splits canonicalization is deferred.** `splits_ext::canonical_preimage` and `splits_ext::distribution_hash` panic via `todo!()`. The upstream payment-channels program ships a fixed-amount distribution model at the current pin, but that is not the final shape; basis-point splits will replace it. Until upstream commits, any preimage layout the SDK invents in isolation is guaranteed to disagree with whatever the program eventually ships. The `Bps` placeholder type (`recipient: Pubkey`, `share_bps: u16`) reserves the eventual basis-point shape without computing anything.

**`verify_open` does not check `distribution_hash`.** Tied to the splits deferral above. When canonicalization ships, splits-aware verification lands as a separate `verify_distribution_hash` function that callers compose alongside `verify_open` rather than as an `Option<[u8; 32]>` field on `ExpectedOpenState`.

**`challenge_id` placement is an authored deviation from the protocol doc.** Upstream `docs/002-http-protocol.md` carries the challenge id in the envelope's `id` field; the SDK puts a `challenge_id` inside each of `OpenPayload`, `TopUpPayload`, and `ClosePayload` because the topup flow specifically needs an explicit challenge binding to reject replays. The protocol doc must be updated before v1 ships; the `TopUpPayload` doc comment documents the gap so a future maintainer reading the SDK does not assume the doc is current.

**`charge.rs` is not aligned with the new dep set.** `rust/src/client/charge.rs` still hand-rolls its own blocking RPC client, ComputeBudget instructions, and program IDs (lines 182-201, 501-527, 182-193). Migrating it onto the new atomic crates is deferred as tech debt; the inline comment on the new `[dependencies]` block tracks this as a follow-up.


TDemeco:
My bad, intended to open it to our fork. Closing

PR #39, opened 2026-05-15
fix(python): pass splits through pay decorator

EfeDurmaz16:
## Summary

Passes the Python `pay(..., splits=[...])` option through to `ChargeOptions`.

The lower-level Python server API already supports split payments through `ChargeOptions.splits`, and the SDK README documents split payments as a supported capability. The ASGI-style `pay` decorator accepted arbitrary options but did not forward `splits`, so decorator users could not include split payment metadata in generated challenges.

## Changes

- Forward `splits` from `pay(..., splits=...)` into `ChargeOptions`
- Add a middleware test that verifies the generated challenge includes `methodDetails.splits`

## Verification

- `PYTHONPATH=python/src python3 -m pytest python/tests/test_middleware.py`
- `python3 -m ruff check python/src/solana_mpp/server/middleware.py python/tests/test_middleware.py`
- `git diff --check`

### DIFF
--- a/python/src/solana_mpp/server/middleware.py
+++ b/python/src/solana_mpp/server/middleware.py
@@ -32,6 +32,7 @@ def pay(mpp_handler: Mpp, amount: str, **options: Any) -> Callable:
         external_id=options.get("external_id", ""),
         expires=options.get("expires", ""),
         fee_payer=options.get("fee_payer", False),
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
LGTM, thanks @EfeDurmaz16!

PR #40, opened 2026-05-15
feat(python): add charge max amount helper

EfeDurmaz16:
## Summary

Adds a small Python helper for validating that a charge request stays under a caller-defined maximum base-unit amount.

Other SDK surfaces already expose amount parsing / max-amount validation patterns for client-side spend caps. This gives Python users a simple equivalent primitive for policy-before-payment flows, especially agent clients that need to reject unexpected or oversized payment requirements before signing.

## Changes

- Add `validate_max_amount(request, max_amount)` to `solana_mpp.protocol.intents`
- Cover below-max, equal-max, above-max, invalid amount, and invalid max cases

## Verification

- `PYTHONPATH=python/src python3 -m pytest python/tests/test_intents.py`
- `python3 -m ruff check python/src/solana_mpp/protocol/intents.py python/tests/test_intents.py`
- `git diff --check`

### DIFF
--- a/python/src/solana_mpp/protocol/intents.py
+++ b/python/src/solana_mpp/protocol/intents.py
@@ -78,3 +78,19 @@ def parse_units(amount: str, decimals: int) -> str:
         raise ValueError(f"invalid amount: {amount}") from exc
 
     return str(val)
+
+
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
LGTM, thanks @EfeDurmaz16 💪

PR #41, opened 2026-05-15
docs(ts): align split limit comment

EfeDurmaz16:
## Summary

Aligns the shared TypeScript Solana charge schema comment with the actual split limit enforced by the SDK.

`server/Charge.ts` rejects more than 8 split entries and the public server parameter docs also say max 8. The shared schema comment still said max 32, which can mislead client/server users looking at the request schema.

## Changes

- Update the `methodDetails.splits` comment from max 32 to max 8

## Verification

- `pnpm --dir typescript install`
- `pnpm --dir typescript --filter @solana/mpp typecheck`
- `pnpm --dir typescript --filter @solana/mpp exec tsc --noEmit --pretty false`
- `git diff --check`


lgalabru (approved):
LGTM, thanks @EfeDurmaz16!

PR #42, opened 2026-05-15
fix(rust): guard session voucher overflow

EfeDurmaz16:
Summary
- Return an error from Rust session voucher increments when cumulative arithmetic would overflow.
- Keep the existing cumulative watermark unchanged on rejected overflow attempts.
- Add a focused unit test for the overflow path.

Why
- `sign_increment` previously added `self.cumulative + amount` directly before calling `sign_voucher`. In debug builds this can panic on overflow, and in release builds it can wrap before the monotonic voucher check sees the value.
- For payment sessions, overflow should be an explicit rejected state rather than arithmetic-dependent behavior.

Verification
- `cargo fmt --manifest-path rust/Cargo.toml --check`
- `cargo test client::session --manifest-path rust/Cargo.toml`
- `git diff --check`

### DIFF
--- a/rust/src/client/session.rs
+++ b/rust/src/client/session.rs
@@ -110,7 +110,14 @@ impl ActiveSession {
 
     /// Sign a voucher adding `amount` to the current cumulative.
     pub async fn sign_increment(&mut self, amount: u64) -> Result<SignedVoucher> {
-        self.sign_voucher(self.cumulative + amount).await
+        let next_cumulative = self.cumulative.checked_add(amount).ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


lgalabru:
I would avoid using sessions at this point, this method is being fully rewritten here: https://github.com/solana-foundation/mpp-sdk/pull/38

EfeDurmaz16:
Thanks for the context. Closing this one to avoid pushing on the session surface while #38 rewrites it. I'll focus follow-up contributions on smaller areas that do not conflict with that rewrite.

Issue #43, opened 2026-05-15
Python pull-mode transaction credentials should verify transfer intent before broadcast


EfeDurmaz16:
## Summary

The Python Solana server currently validates the challenge and transaction blockhash before broadcasting a pull-mode `type="transaction"` credential, then verifies recipient/amount/memo only after `send_raw_transaction`, confirmation, and `get_transaction`.

For agent payment flows, it would be safer if Python could reject a transaction whose decoded instructions do not match the expected charge request before submitting it to the network.

## Current behavior

In `python/src/solana_mpp/server/mpp.py`, `_verify_transaction` does this sequence:

1. Validate credential/challenge binding.
2. Decode the transaction enough to extract the recent blockhash.
3. Check the blockhash against the configured network.
4. Store replay marker.
5. `send_raw_transaction(raw_tx)`.
6. Confirm and fetch the parsed transaction.
7. Run `_verify_confirmed_transaction(tx, request, details)`, which checks expected SOL/SPL transfers and memos.

That means the strongest payment-intent checks happen after broadcast.

## Why this matters

For autonomous or delegated payers, policy should generally fail before execution. If a credential carries a transaction for the wrong recipient, wrong amount, wrong SPL mint, missing `externalId` memo, or unexpected split memo, the server should avoid broadcasting it when possible.

This aligns with payment-agent control-plane expectations: a challenge authorizes a specific payment intent, and settlement should not be attempted unless the submitted transaction matches that intent.

## Possible direction

A small, reviewable fix could be one of:

- Decode legacy/versioned Solana transactions locally and verify their transfer + memo instructions against `ChargeRequest` / `MethodDetails` before `send_raw_transaction`.
- If full local parsing is not reliable yet, fail closed for unsupported transaction shapes before broadcast and document which pull-mode credential shapes are accepted.
- Reuse the existing `_verify_parsed_sol_transfers`, `_verify_parsed_spl_transfers`, and `_verify_parsed_memo_instructions` logic by normalizing decoded local instructions into the same parsed shape used after confirmation.

## Acceptance criteria

- A wrong recipient/amount/mint/memo transaction is rejected before `send_raw_transaction` is called.
- Existing post-confirmation verification remains as a defense-in-depth check.
- Tests cover the pre-broadcast rejection path, ideally with a fake RPC client asserting `send_raw_transaction` was not invoked.

## Notes

This is intentionally issue-first because it touches payment execution semantics and the right local transaction parsing boundary may need maintainer input.

[...]

PR #44, opened 2026-05-15
fix(python): precheck SOL transaction intent

EfeDurmaz16:
Summary
- Decode legacy native SOL transaction credentials before broadcast.
- Verify local System Program transfers and Memo Program instructions against the charge request before `send_raw_transaction`.
- Add regression coverage for wrong recipient, wrong amount, and missing external-id memo proving the fake RPC does not broadcast.

Context
- Addresses the native SOL subset of #43. SPL/token transaction parsing remains a separate follow-up because it needs local token/ATA instruction normalization.

Verification
- `cd python && uv run ruff format src/solana_mpp/server/mpp.py tests/test_server.py`
- `cd python && uv run ruff check --select I,F401 src/solana_mpp/server/mpp.py tests/test_server.py`
- `cd python && uv run --extra dev pytest tests/test_server.py`
- `cd python && uv run --extra dev pyright src/solana_mpp/server/mpp.py tests/test_server.py`
- `git diff --check`

Full suite note
- `cd python && uv run --extra dev pytest` currently reports 207 passed / 5 failed. The failures are existing HTML rendering expectations in `tests/test_server_html.py` (`<!DOCTYPE html>` vs generated `<!doctype html>`, embedded HTML data expectations) and are unrelated to this SOL transaction precheck change.

### DIFF
--- a/python/src/solana_mpp/server/mpp.py
+++ b/python/src/solana_mpp/server/mpp.py
@@ -2,9 +2,9 @@
 
 from __future__ import annotations
 
-import logging
 import base64
 import json
+import logging
 from dataclasses import dataclass, field
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
Looking great thanks @EfeDurmaz16

PR #45, opened 2026-05-15
fix(python): precheck SPL transaction intent

### LINKED ISSUE #43: Python pull-mode transaction credentials should verify transfer intent before broadcast

EfeDurmaz16:
## Summary

- Extend Python pull-mode pre-broadcast verification to SPL Token and Token-2022 `TransferChecked` instructions.
- Normalize locally decoded token transfers into the existing parsed instruction shape and reuse the current transfer/memo verifiers.
- Add regression coverage for matching SPL transactions plus wrong recipient, wrong amount, and missing externalId memo before `send_raw_transaction`.

## Why

#44 added pre-broadcast checks for native SOL transactions. Token payments should get the same fail-before-execution behavior so agent/payment servers do not broadcast a signed transaction that does not match the issued charge intent.

Fixes #43.

## Verification

- `cd python && uv run ruff format src/solana_mpp/server/mpp.py tests/test_server.py`
- `cd python && uv run ruff check --select I,F401 src/solana_mpp/server/mpp.py tests/test_server.py`
- `cd python && uv run --extra dev pytest tests/test_server.py`
- `cd python && uv run --extra dev pyright src/solana_mpp/server/mpp.py tests/test_server.py`
- `git diff --check`

GitHub CI is green for this branch, including the current interop jobs.

## Notes

- `cd python && uv run --extra dev pytest` currently reports 218 passed / 5 failed on existing HTML rendering expectations in `tests/test_server_html.py` (`<!DOCTYPE html>` vs `<!doctype html>` and embedded HTML data expectations). The new `tests/test_server.py` coverage passes.

### DIFF
--- a/python/src/solana_mpp/server/mpp.py
+++ b/python/src/solana_mpp/server/mpp.py
@@ -19,6 +19,8 @@ from solana_mpp._types import PaymentChallenge, PaymentCredential, Receipt
 from solana_mpp.protocol.intents import ChargeRequest, parse_units
 from solana_mpp.protocol.solana import (
     MEMO_PROGRAM,
+    TOKEN_2022_PROGRAM,
+    TOKEN_PROGRAM,
     CredentialPayload,
     MethodDetails,
     default_rpc_url,
@@ -37,6 +39,7 @@ _SECRET_KEY_ENV_VAR = "MPP_SECRET_KEY"
 _CONSUMED_PREFIX = "solana-charge:consumed:"
 _SYSTEM_PROGRAM = "11111111111111111111111111111111"
 _SYSTEM_TRANSFER_INSTRUCTION = 2
+_TOKEN_TRANSFER_CHECKED_INSTRUCTION = 12
 
 
 def _build_expected_transfers(request: ChargeRequest, details: MethodDetails) -> list[tuple[str, int]]:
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
Thanks @EfeDurmaz16!

PR #47, opened 2026-05-15
fix(go): mark paid middleware responses uncacheable

EfeDurmaz16:
## Summary
- set `Cache-Control: no-store` on Go middleware 402 challenges and paid responses
- add `Vary: Authorization` so shared caches do not mix paid and unpaid variants
- cover no-auth, invalid-auth, and valid paid middleware paths

## Why
`Payment-Receipt` and authorization-bound paid responses should not be reusable from a shared cache without considering the `Authorization` header.

## Verification
- `cd go && GOCACHE=/tmp/go-build-cache go test ./server`
- `cd go && GOCACHE=/tmp/go-build-cache go test ./...`

### DIFF
--- a/go/server/middleware.go
+++ b/go/server/middleware.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	mpp "github.com/solana-foundation/mpp-sdk/go"
 )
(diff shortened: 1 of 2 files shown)


lgalabru (approved):
Thanks @EfeDurmaz16!

PR #49, opened 2026-05-15
fix(go): handle merged payment challenges

EfeDurmaz16:
## Summary
- parse Payment challenges out of multiple and merged `WWW-Authenticate` header values in Go
- let `PaymentTransport` retry when a non-Payment auth scheme appears before the Payment challenge
- add protocol and transport regression coverage for merged challenges

## Why
HTTP `WWW-Authenticate` can contain multiple challenges either as repeated header values or as one comma-separated value. The Go transport used `Header.Get` and parsed that single value directly, so a response such as `Bearer realm="api", Payment ...` returned the original 402 instead of building a payment credential.

## Verification
- `go test ./client -run TestTransport402RetryWithMergedWWWAuthenticate -count=1` failed before the fix
- `go test ./client -run TestTransport402RetryWithMergedWWWAuthenticate -count=1`
- `go test ./client ./protocol/core`
- `go test ./...`

### DIFF
--- a/go/client/transport.go
+++ b/go/client/transport.go
@@ -58,15 +58,12 @@ func (t *PaymentTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 		return resp, nil
 	}
 
-	wwwAuth := resp.Header.Get(mpp.WWWAuthenticateHeader)
-	if wwwAuth == "" {
+	challenges := mpp.ParseWWWAuthenticateAll(resp.Header.Values(mpp.WWWAuthenticateHeader))
+	if len(challenges) == 0 {
 		return resp, nil
 	}
 
-	challenge, err := mpp.ParseWWWAuthenticate(wwwAuth)
-	if err != nil {
-		return resp, nil
-	}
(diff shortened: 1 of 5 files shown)


go/client/transport.go
```diff
@@ -58,15 +58,12 @@ func (t *PaymentTransport) RoundTrip(req *http.Request) (*http.Response, error)
...
+	challenges := mpp.ParseWWWAuthenticateAll(resp.Header.Values(mpp.WWWAuthenticateHeader))
+	if len(challenges) == 0 {
 		return resp, nil
 	}
 
-	challenge, err := mpp.ParseWWWAuthenticate(wwwAuth)
-	if err != nil {
-		return resp, nil
-	}
+	challenge := challenges[0]
```
lgalabru, line 66: We may want to implement a less naive challenge selection method in the future (I believe already implemented in typescript / rust)

lgalabru (approved):
Thanks @EfeDurmaz16 !

PR #50, opened 2026-05-15
fix(python): precheck SPL transaction intent

EfeDurmaz16:
## Summary
- decode legacy Token/Token-2022 `TransferChecked` instructions before broadcasting pull-mode Python transactions
- run the existing SPL transfer and memo verifiers in the pre-broadcast path
- add a regression test proving a wrong SPL recipient is rejected before `send_raw_transaction`

## Why
Issue #43 was partially addressed for native SOL in #44, but SPL/token pull-mode credentials still skipped local intent checks before broadcast. For delegated payment flows, the server should reject locally-decodable token transactions whose recipient/amount/mint does not match the charge request before submitting them to the network.

## Verification
- `python/.venv/bin/pytest tests/test_server.py -k wrong_spl_recipient_before_broadcast -q`
- `python/.venv/bin/pytest tests/test_server.py -q`
- `python/.venv/bin/ruff check src/solana_mpp/server/mpp.py tests/test_server.py`
- `python/.venv/bin/pyright --pythonpath .venv/bin/python src/solana_mpp/server/mpp.py tests/test_server.py`

Note: full `python/.venv/bin/pytest -q` currently fails on existing `tests/test_server_html.py` HTML fixture expectations unrelated to this change.

### DIFF
--- a/python/src/solana_mpp/server/mpp.py
+++ b/python/src/solana_mpp/server/mpp.py
@@ -19,6 +19,8 @@ from solana_mpp._types import PaymentChallenge, PaymentCredential, Receipt
 from solana_mpp.protocol.intents import ChargeRequest, parse_units
 from solana_mpp.protocol.solana import (
     MEMO_PROGRAM,
+    TOKEN_2022_PROGRAM,
+    TOKEN_PROGRAM,
     CredentialPayload,
     MethodDetails,
     default_rpc_url,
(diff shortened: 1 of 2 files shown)


EfeDurmaz16:
Closing this as a duplicate of #45. #45 covers the same SPL pre-broadcast intent path more completely, including matching SPL transactions, wrong recipient, wrong amount, and missing memo coverage, and its CI is already green.

PR #56, opened 2026-05-16
fix(go): select supported charge challenge

EfeDurmaz16:
## Summary

Updates the Go `PaymentTransport` challenge selection so it does not blindly use the first parsed `WWW-Authenticate` Payment challenge.

The transport now selects the first supported `method=solana` + `intent=charge` challenge and leaves the original 402 response untouched when no supported challenge is present.

## Motivation

As MPP interop expands across languages and servers may advertise multiple payment methods/intents, the Go client should only build credentials for challenges it actually supports. This keeps Go aligned with the explicit challenge-selection behavior used elsewhere and avoids retrying with an unsupported challenge echo.

## Verification

- `gofmt -w go/client/transport.go go/client/transport_test.go`
- `git diff --check`
- `cd go && go test ./client ./protocol/core`
- `cd go && go test ./...`

### DIFF
--- a/go/client/transport.go
+++ b/go/client/transport.go
@@ -96,11 +96,15 @@ func (t *PaymentTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 
 func selectChargeChallenge(challenges []mpp.PaymentChallenge) (mpp.PaymentChallenge, bool) {
 	for _, challenge := range challenges {
-		if string(challenge.Method) == "solana" && challenge.Intent.IsCharge() {
+		if isSupportedChargeChallenge(challenge) {
 			return challenge, true
 		}
 	}
 	return mpp.PaymentChallenge{}, false
+}
+
+func isSupportedChargeChallenge(challenge mpp.PaymentChallenge) bool {
+	return challenge.Method == mpp.NewMethodName("solana") && challenge.Intent.IsCharge()
 }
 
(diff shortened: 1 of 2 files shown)


go/client/transport.go
```diff
@@ -91,6 +94,15 @@ func (t *PaymentTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	return t.base().RoundTrip(retry)
 }
 
+func selectChargeChallenge(challenges []mpp.PaymentChallenge) (mpp.PaymentChallenge, bool) {
+	for _, challenge := range challenges {
+		if string(challenge.Method) == "solana" && challenge.Intent.IsCharge() {
```
lgalabru: should we introduce a constant for "solana", or add helper for these checks `== "solana"`

EfeDurmaz16: adding a helper for it!

EfeDurmaz16:
Added a small helper for this PR's supported challenge check.

I kept the scope narrow here: the transport now uses `isSupportedChargeChallenge` for the Solana charge selection path touched in this PR. I’m leaving any broader cleanup/constant sweep across the Go SDK for a separate follow-up so this stays easy to review.

Verification:
- `go test ./client`
- `go test ./...`
- `git diff --check`

EfeDurmaz16:

PR #61, opened 2026-05-16
fix(python): reject malformed auth params

EfeDurmaz16:
## Summary

Tighten Python `WWW-Authenticate: Payment` auth-param parsing so malformed parameter lists fail closed instead of being partially accepted.

This catches cases such as:

- missing comma separators between auth params
- trailing non-param text after an otherwise valid challenge
- unterminated quoted values

## Why

The parser is used on the protocol boundary. Accepting a partial parse can hide producer bugs and make interop failures harder to diagnose across SDK implementations.

## Verification

- `PYTHONPATH=src uv run pytest tests/test_headers.py`
- `uv run ruff check src/solana_mpp/_headers.py tests/test_headers.py`
- `uv run ruff format --check src/solana_mpp/_headers.py tests/test_headers.py`
- `git diff --check`

Additional note: I also tried `PYTHONPATH=src uv run pytest tests`; it collected 220 tests and passed 215, but 5 existing `tests/test_server_html.py` expectations failed against the generated HTML output. Those failures appear unrelated to this header parser change.

### DIFF
--- a/python/src/solana_mpp/_headers.py
+++ b/python/src/solana_mpp/_headers.py
@@ -253,29 +253,46 @@ def _parse_auth_params(params_str: str) -> dict[str, str]:
     params: dict[str, str] = {}
     chars = list(params_str)
     i = 0
+    first = True
 
     while i < len(chars):
-        # Skip whitespace and commas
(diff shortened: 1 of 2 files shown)


EfeDurmaz16:

PR #62, opened 2026-05-16
fix(lua): reject malformed auth params

EfeDurmaz16:
## Summary

- make the Lua `WWW-Authenticate` auth-param parser reject missing comma separators
- reject trailing junk after a parsed parameter
- reject unterminated quoted auth-param values
- add a focused Lua core regression test for malformed MPP challenge headers

## Why

The Lua parser previously skipped commas and whitespace together before reading the next auth-param. That meant malformed challenge headers such as `id="..." realm="..."` could be accepted instead of failing early. It also accepted unterminated quoted values.

This brings Lua closer to the stricter behavior added for Python in the adjacent parser hardening work.

## Verification

- `mise x lua@5.4 -- bash -lc 'cd lua && lua -e "package.path = table.concat({\"./?.lua\",\"./?/init.lua\",\"./lua/?.lua\",\"./lua/?/init.lua\",package.path},\";\"); require(\"tests.core_spec\"); require(\"tests.test_helper\").run()"'`
- `mise x lua@5.4 -- bash -lc 'cd lua && luac -p mpp/protocol/core/headers.lua tests/core_spec.lua'`
- `git diff --check`

Note: full `cd lua && lua tests/run.lua` reaches the new core test successfully, then fails later in the existing HTML server spec because `mpp/server/html.lua` expects generated `payment_ui_js` assets that are absent in this local checkout.

### DIFF
--- a/lua/mpp/protocol/core/headers.lua
+++ b/lua/mpp/protocol/core/headers.lua
@@ -28,24 +28,65 @@ end
 local function parse_auth_params(input)
   local params = {}
   local i = 1
+  local first = true
+
   while i <= #input do
-    while i <= #input and input:sub(i, i):match('[%s,]') do
+    while i <= #input and input:sub(i, i):match('%s') do
       i = i + 1
     end
     if i > #input then
       break
     end
-    local eq = input:find('=', i, true)
-    if not eq then
+
+    if not first then
+      if input:sub(i, i) ~= ',' then
+        error('invalid auth parameter separator')
+      end
+      i = i + 1
+      while i <= #input and input:sub(i, i):match('%s') do
+        i = i + 1
+      end
+      if i > #input then
+        error('invalid auth parameter')
+      end
+    elseif input:sub(i, i) == ',' then
       error('invalid auth parameter')
     end
-    local key = input:sub(i, eq - 1):gsub('%s+$', '')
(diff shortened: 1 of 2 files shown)


EfeDurmaz16:
Closing to consolidate the Lua surface into the new stack: #97 (foundation), #99 (RFC conformance for auth params, canonical JSON, multi-challenge, RFC 3339), and the planned follow-up PR B for example servers and the interop adapter. The current open Lua PR count was confusing the review (Ludo's "rainbow" comment).

The strict malformed-auth-params behavior this PR added is not in #99 yet and is captured here so it can land as a small follow-up commit after #99 merges:

- `parse_auth_params` should raise `invalid auth parameter separator` when a comma is missing between consecutive auth-params (`id="x" realm="y"` should fail).
- raise `invalid auth parameter` when a parameter has an empty key (`,=,` or `=foo`).
- raise an explicit error when a quoted value is unterminated (the `closed` flag in the v1 patch).

The `lua/tests/core_spec.lua` fixture from this PR lifts verbatim into the follow-up pass. Tracking as a sub-item under audit v2 RFC track C row 4 (`auth-param token-form value`) plus row 12 (`auth-param NAME case`); both already in scope for a future cross-SDK strictness pass.

PR #64, opened 2026-05-16
fix(rust): reject malformed auth params

EfeDurmaz16:
## Summary

- make the Rust `WWW-Authenticate` auth-param parser reject missing comma separators
- reject key-without-value params instead of silently skipping them
- reject trailing junk after a parsed parameter
- reject unterminated quoted auth-param values

## Why

The Rust parser previously skipped commas and whitespace together before each auth-param, and skipped malformed params without `=`. That meant malformed challenge headers such as `id="..." realm="..."` or `id="...", badkey, realm="..."` could parse successfully.

This aligns Rust with the parser hardening work for the Python, Lua, and Go SDKs so malformed MPP challenge headers are rejected consistently across implementations.

## Verification

- `cd rust && cargo fmt --check`
- `cd rust && cargo test protocol::core::headers`
- `git diff --check`


EfeDurmaz16:

PR #67, opened 2026-05-19
fix(go): align SPL ATA creation with interop policy

EfeDurmaz16:
## Summary
- stop the Go SPL client from creating the primary recipient ATA in charge transactions
- preserve split ATA creation behavior and honor ataCreationRequired in Go split metadata
- enable the Go process client in the default interop matrix after verifying it against TypeScript and Rust servers

## Why
TypeScript and Rust servers reject fee-sponsored primary-recipient ATA creation because that owner is not authorized by the challenge ATA policy. The Go client was creating that ATA unconditionally, so Go -> TypeScript and Go -> Rust interop failed even though Go -> Go passed.

## Verification
- cd go && GOCACHE=/private/tmp/mpp-sdk-go-cache go test ./client ./protocol/...
- cd tests/interop && pnpm typecheck
- cd tests/interop && pnpm test

### DIFF
--- a/go/client/charge.go
+++ b/go/client/charge.go
@@ -125,23 +125,26 @@ func BuildChargeTransaction(
 				return protocol.CredentialPayload{}, err
 			}
 		}
-		addTransfer := func(owner solana.PublicKey, amount uint64) error {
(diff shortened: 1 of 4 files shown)


EfeDurmaz16:

PR #84, opened 2026-05-20
feat(kotlin): add charge client credential core

EfeDurmaz16:
## What
- add a Kotlin/JVM package for MPP charge client credential construction
- parse Payment charge challenges and serialize pull-mode Authorization credentials
- add an injected Solana signer abstraction with a raw Ed25519 memory signer for tests
- harden header parsing and keep test-only transaction helpers out of the main API
- run Kotlin tests and upload Jacoco coverage artifacts in CI

## Why
M1 includes Kotlin client-side support for MPP charge. This keeps the first Kotlin surface narrow: challenge parsing, credential construction, and signer abstraction boundaries without choosing an Android wallet implementation. There is no Kotlin server or interop adapter in this PR; that is intentional for the client-only M1 scope.

## Verified
- `gradle test`
- local Jacoco report generation via Gradle
- CI is rerunning after review-feedback fixes

PR #105, opened 2026-05-23
feat(kotlin): full charge intent support (client + Solana wire signing + interop adapter)

EfeDurmaz16:
## Summary

Adds a canonical-reference Kotlin client for the MPP charge intent on Solana, wire-compatible with the Rust spine and the TypeScript reference server, plus an Android Compose demo that runs the client end to end against Surfpool. Kotlin is client-only by design.

## Scope

- Charge-intent SDK at `kotlin/src/main/kotlin/com/solana/mpp/` split into `client/`, `protocol/`, `crypto/` to mirror the Rust layout
- Pure-Kotlin Base58 (Bitcoin alphabet, Ruby round-trip parity), URL-safe Base64, BouncyCastle-backed Ed25519 (RFC 8032 Test 1 vector), inline RFC 8032 §5.1.3 on-curve check
- Transaction codec verified byte-for-byte against the Rust golden vector at `rust/examples/golden_tx.rs`; SPL TransferChecked for Token and Token-2022, idempotent ATA-create, ComputeBudget, Memo, legacy plus v0
- Android Compose demo at `kotlin/examples/AndroidDemo/` wiring the client into a 402-replay flow with MWA-compatible signing surface and on-device run evidence
- Base58 plus multimult swap in `ae1c654` collapses two separate Base58 implementations onto one allocator path
- Interop adapter at `harness/kotlin-client/` registered as `kotlin-client`, on by default, following the env-var contract in `skills/pay-sdk-implementation/references/interop-harness.md`

## Files changed

- SDK: `kotlin/src/main/kotlin/com/solana/mpp/{client,protocol,crypto}/*.kt`
- Tests: `kotlin/src/test/kotlin/com/solana/mpp/**`
- Demo: `kotlin/examples/AndroidDemo/` plus `.github/workflows/android-demo.yml`
- Harness: `harness/kotlin-client/`, adapter registration in `harness/src/implementations.ts`
- Docs: `kotlin/README.md`, `kotlin/examples/AndroidDemo/README.md` (DEMO ONLY banner)

## Security highlights

- Ed25519 signer never sees private material outside `MemorySigner`; on-curve check is inline finite-field arithmetic with no reflection into BouncyCastle internals
- Canonical JSON (RFC 8785) for credential wire bytes so credential bytes are deterministic across language boundaries
- Demo signer secret is publicly known and the README states the demo must never be used with real funds

## Test evidence

- `gradle check`: clean on JDK 24 + Gradle 9.5.1 + Kotlin 2.3.20, line coverage above 90 percent via `jacocoTestCoverageVerification`
- Focused interop matrix `MPP_INTEROP_CLIENTS=kotlin MPP_INTEROP_SERVERS=typescript`: 2 of 2 scenarios green (`charge-basic`, `charge-split-ata`)
- AndroidDemo `:app:assembleDebug` clean locally and on the dedicated CI workflow
- On-chain settlement signature from Surfpool localnet emulator run: `bCcGzNrfXjk3abSE3RqRS2Fx6dF4uFxzuth9biphcBytS1Y6tdaPi93UD4gzZvoefY4oE8EoPEj4qMgdbc99fjJ`
- Codex r8: 0 P1, confidence 4/5; Greptile sweep on tip: no outstanding findings
- Codex r9 (tip `36b9902`): 0 P1, 0 P2, 1 P3 (justfile detekt fallback parity nit), confidence 4/5. Round-4 P2/P3 carryovers closed (kotlin interop default-disabled to match go/swift, CanonicalJsonTest EOF blank line stripped, Android demo `fillMaxWidth` import added, README quickstart imports + MWA version + Base58 wording synced to the post-multimult layout). Report at `notes/codex-review/pr-105-r9.md`.

## Closes / supersedes

- Closes #118
- Supersedes #84

## Reviewer notes

- Folder layout (`client/protocol/crypto`) mirrors `rust/src/{client,protocol,crypto}` per earlier review feedback
- Android demo replaces the previously-requested Solana Seeker example; it is the closest first-party Solana mobile surface available today
- Server adapters remain in TypeScript, Rust, Go, PHP, Ruby, Lua; this PR does not add a Kotlin server


kotlin/src/main/kotlin/com/solana/mpp/HttpClient.kt
```diff
@@ -0,0 +1,168 @@
...
+     * response body.
+     */
+    fun mppGet(url: String, headers: Map<String, String> = emptyMap()): Response {
+        val initial = execute("GET", url, headers, body = null)
+        if (initial.code != 402) {
+            return initial
+        }
+        val wwwAuthenticate = initial.header(WWW_AUTHENTICATE_HEADER)
+            ?: throw MppException.InvalidPaymentScheme
+        initial.close()
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 5603adb.

The 402 header read is now wrapped in Response.use { ... } so the body stream is closed deterministically on both branches (header present and header missing). Regression test doesNotLeakConnectionOnMissingChallengeHeader in HttpClientTest installs an OkHttp EventListener that counts connectionReleased events, sends a 402 with a non-empty body and no WWW-Authenticate header, and asserts the connection is released exactly once. Confirmed locally that the test fails against the pre-fix HttpClient.kt and passes after the use { } wrap.

EfeDurmaz16:
All three Greptile 4/5 concerns addressed in 8e818e1.

1. Ed25519 OnCurveCheck rewritten inline against RFC 8032 §5.1.3 finite-field arithmetic (no reflection into BouncyCastle internals, no JDK 17+ strong-encapsulation breakage). Pure BigInteger; ~70 LOC. The JDK 15+ EdECPublicKeySpec path was tried first but does not enforce on-curve (KeyFactory.generatePublic accepted 200 of 200 random 32 byte samples in a debug test), so the inline implementation was the only correct option. PdaTest still resolves the golden ATA 2xyhP3ZCpBqnjPe1oLzQCXqb5kMrJHY6sVkt7x4mBUz5 at bump 253 with bumps 255 and 254 correctly classified as on-curve and skipped.

2. tests/interop/kotlin-client Main.kt parseSecretKey now raises IllegalArgumentException when a JSON byte falls outside 0..255. New ParseSecretKeyTest (7 tests, JUnit Platform) covers 256 wrapping, -1 wrapping, wrong-size arrays, non-array input, and non-integer elements. Lifted parseSecretKey to internal so the test can exercise it.

3. .github/workflows/ci.yml test-kotlin job now runs `gradle check` instead of `gradle test jacocoTestReport`, so jacocoTestCoverageVerification (the 90 percent line-coverage gate declared in build.gradle.kts) is enforced in CI.

Manual DX gate green after the fixes: `gradle check` clean locally and the focused TS-to-Kotlin interop matrix still passes 2 of 2 charge scenarios on Surfpool localnet (charge-basic 6.1s, charge-split-ata 4.7s).

EfeDurmaz16:

- `2c6cfd1` kotlin (cherry-pick of `50436d4` from `local/m1-matrix-kotlin`, pushed to this PR branch): `ChargeRequest.description: String? = null` so kotlinx-serialization strict decoding accepts the new field added in PR #102; `MppHeaders.decodeChargeRequest` now catches `SerializationException` and surfaces it as `InvalidJson` with cause message attached; OkHttp interop adapter timeouts raised so surfpool warm-up does not trip the 10s default.

Kotlin slice for this PR: 37/37 green (5 clients consumed, kotlin-client x 7 servers all pass).

Round 1 vs round 2 delta: round 1 had kotlin-client failing every non-typescript server pair on strict-decoding the new `description` field. The cherry-pick above resolves it; remaining surfpool warm-up flakes on charge-symbol-usdc-localnet and charge-split-ata cleared with the OkHttp timeout bump.

ready for re-review.

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

.github/workflows/ci.yml
```diff
@@ -211,6 +211,31 @@ jobs:
...
 
+  test-kotlin:
+    name: Kotlin tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: "17"
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 8e818e1: CI now runs `gradle check`, which triggers `jacocoTestCoverageVerification` and enforces the 90% line-coverage gate. Resolving.

.github/workflows/ci.yml
```diff
@@ -211,6 +211,31 @@ jobs:
...
+      - uses: gradle/actions/setup-gradle@v6
+        with:
+          gradle-version: "9.5.1"
+      - name: Run tests with coverage
+        working-directory: kotlin
+        run: gradle test jacocoTestReport
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 8e818e1: CI now runs `gradle check` per the suggestion. Resolving.

kotlin/src/main/kotlin/com/solana/mpp/crypto/Ed25519.kt
```diff
@@ -0,0 +1,224 @@
...
+        m.isAccessible = true
+        m
+    }
+
+    fun validate(bytes: ByteArray): Boolean {
+        if (bytes.size != 32) return false
+        val result = method.invoke(null, bytes, 0)
+        return result as Boolean
+    }
+}
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 8e818e1: `OnCurveCheck` no longer reflects into BouncyCastle internals; the on-curve test is now pure java.math.BigInteger, JDK-portable and stable across BC versions. Resolving.

kotlin/src/main/kotlin/com/solana/mpp/HttpClient.kt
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 5603adb: the 402 branch now uses `initial.use { ... }` so the response body is closed deterministically even when `WWW-Authenticate` is missing. Resolving.

kotlin/src/main/kotlin/com/solana/mpp/client/Charge.kt
```diff
@@ -0,0 +1,380 @@
...
+        signer: SolanaSigner,
+        request: ChargeRequest,
+        blockhashProvider: BlockhashProvider,
+        computeUnitLimit: Int = DEFAULT_COMPUTE_UNIT_LIMIT,
+        computeUnitPrice: Long = DEFAULT_COMPUTE_UNIT_PRICE,
+    ): String {
+        val totalAmount = request.amount.toLongOrNull()
+            ?: throw MppException.InvalidTransaction("Invalid amount: ${request.amount}")
+        val splits = request.methodDetails.splits ?: emptyList()
+        if (splits.size > MAX_SPLITS) {
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 48dda98: negative split amounts are rejected up front with `MppException.InvalidTransaction`, and the total uses `Math.addExact` so overflow surfaces as a structured exception too. Resolving.

kotlin/src/main/kotlin/com/solana/mpp/client/HttpClient.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    private fun post(payload: JsonObject): JsonObject {
+        val body = json.encodeToString(JsonObject.serializer(), payload)
+            .toRequestBody("application/json".toMediaType())
+        val request = Request.Builder().url(url).post(body).build()
+        okHttp.newCall(request).execute().use { response ->
+            val text = response.body?.string()
+                ?: throw MppException.InvalidTransaction("empty RPC response")
+            val parsed = json.parseToJsonElement(text)
+            if (parsed !is JsonObject) {
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 49c17dd: `parseToJsonElement` is now wrapped and any `SerializationException` surfaces as `MppException.InvalidTransaction`. Resolving.

tests/interop/kotlin-client/src/main/kotlin/com/solana/mpp/interop/Main.kt
```diff
@@ -0,0 +1,145 @@
...
+    val element = Json.parseToJsonElement(raw)
+    if (element !is JsonArray) {
+        error("MPP_INTEROP_CLIENT_SECRET_KEY must be a JSON array of bytes")
+    }
+    val bytes = ByteArray(element.size)
+    for ((index, value) in element.withIndex()) {
+        val int = value.jsonPrimitive.intOrNull
+            ?: value.jsonPrimitive.longOrNull?.toInt()
+            ?: error("non-integer byte at index $index in secret key")
+        bytes[index] = int.toByte()
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 5d9208d: the secret-key parser now validates 0..255 explicitly and errors out before `Int.toByte()` wraps. Resolving.

EfeDurmaz16:
## Final status (tip 5d9208d)

- Greptile: 5/5
- CI: all green
- Manual DX: Kotlin walkthrough complete, 2 on-chain devnet signatures captured (27qD6v5b..., 4sFKfHjp...)
- Codex review: applied earlier in cycle (Ed25519 OnCurveCheck rewritten against RFC 8032 §5.1.3 plus two further concerns in 8e818e1), no open items
- Outstanding inline notes: none

Ready for review.

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
## Final status (tip 8908ce1)

- Greptile: 5/5
- CI: all green
- Latest round: 1 audit fix shipped (kt-* justfile recipes added for Kotlin parity with the dedicated language workflow)
- Manual DX: Kotlin walkthrough captured locally, on-chain signatures in mpp-sdk-self-learning/notes/verify-2026-05-24/
- Codex review round 3: 0 P1, 0 P2; no intentional-open inline threads remain
- Ready for re-review.

kotlin/README.md
```diff
@@ -0,0 +1,196 @@
...
+│   ├── Base58.kt           # Bitcoin-alphabet base58 codec
+│   ├── Base64Url.kt        # url-safe base64 (no padding)
+│   ├── Charge.kt           # full charge build pipeline
+│   ├── Ed25519.kt          # BouncyCastle signing + PDA derivation
+│   ├── Headers.kt          # WWW-Authenticate / Authorization parser
+│   ├── HttpClient.kt       # OkHttp client with 402 retry + JSON-RPC
+│   ├── Instructions.kt     # SystemProgram, SPL, ATA, compute budget
+│   ├── Models.kt           # exceptions, challenge + credential schema
+│   ├── SolanaSigner.kt     # signer abstraction + MemorySigner
+│   └── Transaction.kt      # legacy + v0 Solana wire codec
```
lgalabru: Can we structure this folder? Rust is the right inspiration source.

rust/crates/mpp/examples/golden_tx.rs
```diff
@@ -0,0 +1,45 @@
+// Golden-vector generator for the Kotlin transaction codec parity tests.
+// Builds a deterministic legacy SystemProgram transfer transaction with
+// a fixed seed and fixed blockhash, then prints the bincode-serialized
+// bytes as a hex string. The Kotlin codec must produce the same bytes.
+use ed25519_dalek::SigningKey;
```
lgalabru, line 5: Instead of this can we have an integrated example, with a quick demo app or something for solana seeker?

kotlin/README.md
EfeDurmaz16: Restructured kotlin/ to mirror rust/ (com.solana.mpp.{client,protocol,crypto} packages, src/test/kotlin mirrored, examples/). Updated the tests/interop/kotlin-client adapter imports to match. gradle check (incl. 90% Jacoco gate) passes. 80e6a35.

rust/crates/mpp/examples/golden_tx.rs
EfeDurmaz16: Filed as #114. Demo Android app (Solana Seeker target) lands as a separate M2 deliverable so this PR stays focused on the client SDK + interop adapter. The restructure in 80e6a35 opens up kotlin/examples/AndroidDemo/ as the home for it; the issue carries concrete acceptance criteria (./gradlew assembleDebug green, end-to-end charge against https://402.surfnet.dev, on-chain receipt verifiable on Solana Explorer, Mobile Wallet Adapter integration as the preferred Seeker path with MemorySigner fallback documented, dedicated CI smoke build).

kotlin/examples/README.md
```diff
@@ -0,0 +1,12 @@
...
+Sample clients exercising the `solana-mpp-kotlin` library.
+
+- `ChargeClient/` — minimal JVM CLI that performs an `mpp/charge/pull`
+  against a 402-protected endpoint. Build via the existing `gradle build`
+  in the parent project; the example is source-only today.
+
+Planned (M2): `AndroidDemo/` — Compose UI app targeting the Solana
+Seeker dev kit, end-to-end charge intent flow against
+`https://402.surfnet.dev`. Tracked as a separate deliverable to keep
+the SDK PR focused.
```
lgalabru: remove this file?

kotlin/README.md
```diff
@@ -0,0 +1,203 @@
+<p align="center">
+  <img src="https://github.com/solana-foundation/mpp-sdk/raw/main/assets/banner.png" alt="MPP" width="100%" />
+</p>
+
+# solana-mpp-kotlin
+
+Charge stablecoins (USDC, USDT, PYUSD, ...) for any HTTP endpoint, in Kotlin.
```
lgalabru: Consume stablecoin gated endpoints etc
This lib is client only.
Can we fix the swift readme too in this PR?

kotlin/examples/README.md
EfeDurmaz16: Done in a861da8. The top-level README was stale (described AndroidDemo as 'Planned' even though the demo ships in this PR with its own self-contained README). ChargeClient is a single Main.kt that does not need a directory-level index.

EfeDurmaz16:
Kotlin pivot landed. Two commits:

- `7817dc1` feat(kotlin/client): expose buildUnsignedChargeTransaction for external signers — additive SDK helper that returns unsigned legacy Solana wire bytes. `buildChargeTransaction` delegates to it then signs locally, so both paths share one composition step. Regression test asserts unsigned bytes equal locally-signed bytes once the signature slot is zeroed.
- `4f87097` feat(kotlin/examples): adopt mobile-wallet-adapter for real wallet flow — AndroidDemo now uses `com.solanamobile:mobile-wallet-adapter-clientlib-ktx:2.0.0` and signs transactions via `walletAdapter.transact { signTransactions(...) }`. The deterministic `DemoSigner` and the demo-only warning banner are gone. README documents real wallet (Phantom/Solflare/Backpack) and the mock-mwa-wallet path for emulator testing.

Two of the originally-scoped swaps are deferred with reasoning:

- **rpc-core**: skipped. The demo's two RPC calls (`getLatestBlockhash`, `sendTransaction`) are covered by the existing `JsonRpcClient` (120 LOC, MockWebServer tests). Adopting rpc-core would add a beta dep and a driver-config layer for no LOC reduction or correctness gain. Worth revisiting if the SDK ever needs full RPC client surface (account info, program subscriptions, etc.).
- **web3-solana curve-check swap**: skipped. The library's public surface exposes `ProgramDerivedAddress.find` (whole bump-loop + curve-check together) but does not appear to expose an atomic `SolanaPublicKey.isOnCurve` we could swap in isolation. Replacing the whole `Pda.findProgramAddress` is wire-affecting (ATAs could shift if their derivation differs by even one bump iteration) and was out of scope per the wire-byte-compat constraint. Worth a follow-up after reading web3-core source to confirm what's actually public.

`Transaction.kt`, `Instructions.kt`, `Models.kt`, `CanonicalJson.kt`, `Base58.kt`, `Base64Url.kt`, and `HttpClient.kt` stay hand-rolled to preserve wire-byte compatibility with the interop harness. Jacoco 90% line gate still passing.

Tests:
- `gradle test :jacocoTestCoverageVerification` green
- `gradle :app:assembleDebug` (AndroidDemo) green

EfeDurmaz16:
End-to-end MWA flow proven against `solana-mobile/mock-mwa-wallet` on an Android emulator (`mpp-demo` AVD, API 34) talking to local surfpool and the Python iOS-demo merchant on `:3004`.

Two demo-side bugs surfaced and are fixed in 25cd067:

1. `ConnectionIdentity.iconUri` rejected absolute URIs. MWA spec requires `icon` to be relative to `identity_uri`; `mobile-wallet-adapter-clientlib`'s `AuthorizationRequest.Builder` enforces this with `IllegalArgumentException: If non-null, iconRelativeUri must be a relative Uri`. Switched to `Uri.parse("icon.png")`.
2. `NetworkOnMainThreadException` on the merchant replay. `okhttp3.Response.body.string()` reads lazily; the original code closed the response on the Main coroutine after `withContext(IO)` returned. Moved both the body read and header extraction inside the `withContext(IO)` block, plus added `Log.e` so future failures show a stack trace instead of `(no message)`.

After the fixes, the full flow works end to end:

- Connect Wallet -> mock-mwa authorize prompt -> wallet pubkey `5mn2QRmLtgVNbaY3Y7T2hTbcHJRhqktuTrEe2yK7pbWi`
- Pay -> `sign_transactions` intent -> wallet signs -> demo replays GET with `Authorization: Payment ...` -> merchant returns 200 with fortune
- On-chain signature: `5anzHR9SygbcVK12t6gyvjkAUaM3qX97TDErSFbcUGNTBb99XuJyUJEsqsceqPMKyVxSpxso8vMQQ8vb1JNFN3n2`
- `solana confirm -v <sig> --url http://localhost:8899` returns the transaction executed in slot 422146326, fee payer `5mn2QRm...pbWi`, instructions include ComputeBudget + SPL Token `transferChecked` against mint `EPjFWdd5...DTtv1v` (USDC). Recent blockhash `SURFNETxSAFEHASH...` confirms it was settled on the local surfpool fork.

Evidence committed in 50a9a33:

- `kotlin/examples/AndroidDemo/docs/e2e-1-launch.png`
- `kotlin/examples/AndroidDemo/docs/e2e-2-authorized.png`
- `kotlin/examples/AndroidDemo/docs/e2e-3-success.png`
- `kotlin/examples/AndroidDemo/docs/e2e-signature.txt`

MWA spec sanity check: `MobileWalletAdapter.transact { authorize(); signTransactions(...) }` flows entirely through `mobile-wallet-adapter-clientlib-ktx`, which owns the JSON-RPC framing, encrypted session, and method payloads. The demo does not bypass the library, so MWA 2.0 conformance is inherited.

Funding note: mock-mwa generates a fresh random keypair on first authorize. The pubkey was funded post-authorize via `surfnet_setAccount` (1 SOL) and `surfnet_setTokenAccount` (1000 USDC) before tapping Pay. For automation, future runs would benefit from a merchant flag or helper script that funds the wallet account once it is announced.

kotlin/README.md
EfeDurmaz16: Reframed both READMEs to lead with 'consume stablecoin-gated HTTP endpoints' and call out the client-only scope explicitly. swift/README.md updated in the same commit (5081df434242868517e88ada6b7b626679239bc8). solana-foundation/pay-kit@5081df434242868517e88ada6b7b626679239bc8

kotlin/src/main/kotlin/com/solana/mpp/crypto/Base58.kt
```diff
@@ -0,0 +1,83 @@
+package com.solana.mpp.crypto
```
lgalabru, line 1: I'm surprised we still have these files after pulling the solana web3 lib

EfeDurmaz16: Good catch. The Solana Mobile lib was referenced from the README and the AndroidDemo example, not from the SDK proper, which is why these files were still hand-rolled.

I just pushed ae1c654 that swaps Base58 over to the multimult Base58 codec from Solana Mobile (the same one web3-solana and mobile-wallet-adapter-clientlib pull in). Net 62 lines of hand-rolled BigInteger code removed, public API kept, tests + 90% coverage gate still green.

I held off on swapping the rest in this PR because doing it properly is more invasive than it looks:

1. web3-solana SolanaSigner is suspend-based (signPayload, signTransaction). Adopting it cascades through Charge.kt and HttpClient.kt and flips the public Kotlin API to coroutines. That's a directional decision I'd rather make in a follow-up PR.
2. web3-solana ships AssociatedTokenProgram.createAssociatedTokenAccount but no idempotent variant. The MPP charge spine uses ATA create idempotent (discriminator 0x01, 6 accounts, no rent sysvar), so swapping that one breaks the wire-byte parity test against the Rust spine. Either we contribute idempotent upstream or keep that single builder hand-rolled.
3. web3-solana has no in-process JVM Ed25519 signer. The abstract Ed25519Signer expects the host to provide signing (MWA on Android). The interop harness needs to load raw 32-byte keypair seeds, so BouncyCastle stays on JVM regardless.

Happy to do the deeper migration (Message.Builder + program builders + suspend signer) in a follow-up. Want me to open that as a separate PR after this one lands, or expand scope here?

lgalabru: Ok, let's track this gap with an issue, I'll try to get team kotlin to move to a pattern using a kotlin version of solana-keychain.

EfeDurmaz16: Tracked at https://github.com/solana-foundation/pay-kit/issues/135 with the five concrete blockers documented (idempotent ATA missing upstream, suspend signer API break, JVM Ed25519, Message.Builder ordering edge case, coverage gate math). Will revisit once the solana-keychain kotlin port lands.

PR #106, opened 2026-05-23
feat(python): full charge intent support (interop adapter + L1-L11 hardening + RFC F1-F6 + B34 + structured error codes)

EfeDurmaz16:
## Summary

Brings Python SDK to full charge-intent parity with Rust, Ruby, PHP, and Lua. Closes the L8 confirmation-timeout double-pay window on Python (broadcast then consume then await), lands the audit-v2 hardening surface, and registers the Python server as a default-on participant in the cross-language interop harness.

Also folds in #115 (now closed as superseded): adds `docs/security/fee-payer-drain.md` next to the canonical Python spine because the implementation in this PR is the reference enforcement of the doc's four cross-SDK invariants (server-context fee-payer key, SOL drain reject, SPL `transferChecked` drain reject, B34 signature-credential reject under `feePayer=true`).

Closes #115.

## Python SDK scope

- **Full charge intent**: pull-mode settler with L8 broadcast then consume then await ordering (closes the same double-pay window #96 / #102 closed on Rust / Ruby / PHP / Lua, but for Python).
- **Interop adapter**: `tests/interop/python-server/main.py` plus default-on registration in `tests/interop/src/implementations.ts`. 13 / 13 cross-language scenarios pass against TypeScript and Rust clients.
- **L1-L11 hardening**:
  - L1 mainnet canonical slug, mainnet-beta alias
  - L2 reject Solana memo v1 (`Memo1Uhk...`)
  - L4 explicit replay store required, `FileReplayStore` added, fail-closed on corrupted file
  - L6 canonical structured error codes in 402 problem+json
  - L8 broadcast then consume then await ordering (headline)
  - L11 reject CR or LF in header parameter values
- **RFC F1-F6**:
  - F1 accept token-form auth-param values per RFC 7235
  - F2 RFC 8785 UTF-16 code-unit key sort
  - F3 ECMA-262 7.1.12.1 `ToString` numbers (`1e21`, `0.000001`, `-0.0`, reject `NaN` / `Inf`)
  - F4 lone-surrogate reject per RFC 8785 section 3.2.2
  - F5 multi-challenge `WWW-Authenticate` (quote-aware split)
  - F6 strict RFC 3339 grammar for challenge expiry, fail closed
- **B34**: reject signature credential under `feePayer=true` before any RPC call.
- **Structured error codes**: canonical L6 codes in 402 problem+json bodies via `payment_required_response` builder; route-mismatch gets a distinct code.

## Security doc (folded from #115)

`docs/security/fee-payer-drain.md` documents the fee-payer drain attack pattern as a cross-SDK invariant. Four invariants every language port MUST enforce:

1. Server-context fee-payer key (never trust client-supplied `feePayerKey` from the credential).
2. Reject System Program `Transfer` instructions sourced from the fee-payer account.
3. Reject SPL `transferChecked` instructions sourced from the fee-payer ATA.
4. B34 reject of signature credential under `feePayer=true` before any RPC call.

This PR contains the reference Python implementation of all four (`python/src/solana_mpp/server/mpp.py`).

## Flask example

`python/examples/simple-server/app.py` mirrors the Ruby Sinatra example, with clean MPP middleware isolation, environment-driven configuration, and a single protected route. Matches the structure Ludo asked for in the inline review.

## Coverage

`python` package: 92.80% line coverage, gate enforced at 90% via `pytest --cov-fail-under=90` in `.github/workflows/python.yml`. 442 unit tests pass.

## Test results

- Python unit tests: 442 passed, coverage 92.80%.
- Python interop matrix (Python server with TS + Rust clients): 13 / 13 passed.
- Greptile confidence score: 5 / 5.

## Spec links

- HTTP Payment Authentication Scheme: https://paymentauth.org
- Solana Charge Intent spec PR: https://github.com/tempoxyz/mpp-specs/pull/188


tests/interop/python-server/main.py
```diff
@@ -0,0 +1,348 @@
...
+        currency=mint,
+        decimals=int(optional_env("MPP_INTEROP_DECIMALS", "6")),
+        network=network,
+        rpc_url=rpc_url,
+        secret_key=secret_key,
+        realm=optional_env("MPP_INTEROP_REALM", "MPP Interop"),
+        fee_payer_signer=fee_payer,
+        store=MemoryStore(),
+        rpc=rpc,
+    )
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in b9546ed: dropped the boot-time SolanaRpc entirely. Mpp now boots with rpc=None and the do_GET handler constructs + aclose()'s a fresh SolanaRpc per request, scoped to that request's asyncio.run() event loop. The CI failure on 26337787977 had the same root cause class (workflow built TS package AFTER the harness install copied @solana/mpp into node_modules with no dist/); both are fixed in the same commit.

python/src/solana_mpp/_canonical_json.py
```diff
@@ -0,0 +1,198 @@
...
+
+def _encode(value: Any) -> str:
+    if value is None:
+        return "null"
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, str):
+        return _encode_string(value)
+    if isinstance(value, int) and not isinstance(value, bool):
+        return _encode_number(float(value)) if isinstance(value, bool) else str(value)
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 95945b4: collapsed to a single isinstance(int) guard with a comment explaining why the bool-subclass-of-int gotcha does not bite (the bool branch above returns first).

python/src/solana_mpp/_rpc.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class SolanaRpc:
+    """Minimal async JSON-RPC client for the Solana RPC API."""
+
+    def __init__(self, endpoint: str, timeout: float = 30.0) -> None:
+        self._endpoint = endpoint
+        self._timeout = timeout
+        self._client = httpx.AsyncClient(timeout=timeout)
+        self._id = 0
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in f18345b: switched to itertools.count(). The asyncio.Lock in dd6ff27 was per-loop and the interop adapter creates a fresh asyncio.run() loop per request, so different loops could collide. itertools.count() increments atomically under the GIL and is loop-agnostic.

python/src/solana_mpp/store.py
```diff
@@ -16,22 +39,115 @@ async def put_if_absent(self, key: str, value: Any) -> bool: ...
...
-        if key in self._data:
-            return False
-        self._data[key] = value
-        return True
+        async with self._lock:
+            if key in self._data:
+                return False
+            self._data[key] = value
+            self._flush()
+            return True
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 12d66cb: all three mutators (put, delete, put_if_absent) now build a candidate next_data dict, flush it to disk, and only commit to self._data on flush success. A failed flush propagates the IOError and leaves the previously committed state untouched. New TestFileReplayStore::test_flush_failure_rolls_back_in_memory_state and test_failed_put_keeps_previous_committed_state pin the rollback invariant.

EfeDurmaz16:
Both Greptile follow-up concerns addressed in 3a0ca5c.

1. `_co_sign_with_fee_payer` now calls `_assert_signature_slot` to verify `idx < num_required_signatures` before the splice. An out-of-range index raises `PaymentError(code=invalid-payload)` so a client that places the fee-payer pubkey outside the required-signer block sees a deterministic 402 instead of a corrupted transaction. `TestCoSignSplitBounds` pins the reject path (fee-payer in readonly-unsigned block) plus a happy-path positive control.

2. `tests/interop/python-server/main.py` 402 responses now emit `Content-Type: application/problem+json` per RFC 7807 section 3. The `_send_json` helper now lets the 402 callsite override the default content type via `extra_headers`. New `python/tests/test_interop_adapter.py` spawns the adapter as a subprocess, hits `/paid` without auth, and asserts the 402 content type plus the L6 canonical body shape.

Quality gates after the fix: pytest 303 passed, ruff check + ruff format clean, pyright clean. Focused TS-to-Python (4/4) and Rust-to-Python (2/2) interop matrices still green locally against Surfpool 1.1.1.

EfeDurmaz16:

- `ee1c69f` python: compute-budget cap mapping plus splits sum/length guards before signing, with 8 new pytest tests covering both guards and the canonical compute-cap path.

Python slice for this PR: 2/2 on the targeted matrix slice; full python-server x 5-client coverage all green.

Round 1 vs round 2 delta: round 1 surfaced two python-server failures on compute-cap drift versus typescript/rust references and missing splits validation. The commit above fixes both root causes and locks them with unit tests so regressions surface before the matrix runs.

ready for re-review.

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

tests/interop/python-server/main.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed. The interop adapter no longer constructs a long-lived SolanaRpc / httpx.AsyncClient at boot. Each do_GET handler builds a fresh SolanaRpc inside its per-request asyncio.run loop and aclose()s it in a finally block, so the AsyncClient is always bound to the loop it runs in. The Mpp handler now boots with rpc=None and the per-request client is plugged in just-in-time. Commit 25475e8 (full chain: b9546ed, dd6ff27, 25475e8). Verified by typescript+rust clients passing against the python server in the interop matrix.

python/src/solana_mpp/store.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in commit 12d66cb. FileReplayStore.put_if_absent (and put / delete) now build the next dict, flush to disk, and only swap self._data on success. If _flush raises, the in-memory state is unchanged, so a subsequent get reports the key as absent and a fresh instance loading from disk agrees. Regression covered by python/tests/test_store.py::test_flush_failure_rolls_back_in_memory_state and ::test_failed_put_keeps_previous_committed_state.

tests/interop/python-server/main.py
```diff
@@ -0,0 +1,382 @@
...
+    if not isinstance(splits, list):
+        print("MPP_INTEROP_SPLITS must decode to a JSON array", file=sys.stderr)
+        sys.exit(2)
+
+    # Fee-payer keypair is optional in the harness but always present when
+    # the scenario uses server-side fee sponsorship.
+    fee_payer_bytes = _decode_keypair_env("MPP_INTEROP_FEE_PAYER_SECRET_KEY")
+    from solders.keypair import Keypair
+
+    fee_payer = Keypair.from_bytes(fee_payer_bytes)
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in commit 33de5ff. The fee-payer keypair is now read with os.environ.get; absence is a no-op (no sys.exit), fee_payer_signer is only passed to Config when the env var is present, and the challenge therefore only advertises feePayer when the signer is configured. Verified by the full interop matrix (no MPP_INTEROP_FEE_PAYER_SECRET_KEY scenarios pass against the python server).

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
Note for Greptile: commit a3c6fea addresses both observations from the last review:

1. `FileReplayStore._flush` is now invoked through `_flush_off_loop` which trampolines through `asyncio.to_thread`, so all three async mutators (`put`, `delete`, `put_if_absent`) hand the synchronous `os.fsync` to the default executor instead of stalling the event loop.

2. The interop adapter no longer mutates `self.mpp._rpc` directly. `Mpp` now exposes a `using_rpc` async context manager that serializes the per-request RPC swap under a per-instance `asyncio.Lock` and always restores the prior value on exit, including when the body raises. `tests/interop/python-server/main.py` routes the fresh `SolanaRpc` through it.

Both items were the only non-blocking observations in the last 4/5 summary; please re-evaluate. @greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

python/src/solana_mpp/server/mpp.py
```diff
@@ -347,10 +468,468 @@ def _decode_legacy_payment_instructions(transaction_b64: str) -> list[dict[str,
...
+                 if amt == amount and _verify_ata_owner(destination, rcpt, mint, program_id)),
+                -1,
+            )
+            if match_idx == -1:
+                raise PaymentError(
+                    "unexpected Token Program transfer in payment transaction",
+                    code="invalid-payload",
+                )
+            remaining_transfers.pop(match_idx)
+            continue
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 5bf71d9. The allowlist now mirrors the rust spine for both transfer-like branches when `methodDetails.feePayer` is true:

- SPL `transferChecked` branch reads `accounts[0]` (source ATA) and `accounts[3]` (authority). It rejects with `code=invalid-payload` if `authority == fee_payer_pubkey` (covers the explicit case you described) or if `_verify_ata_owner(source_ata, fee_payer_pubkey, mint, program_id)` matches (covers any third-party authority signing over a source ATA derived from the fee-payer).
- System Program `Transfer` branch reads `accounts[0]` (source) and rejects if it equals `fee_payer_pubkey`. Same shape gap closed for SOL; the prior allowlist only failed drain attempts when the destination differed from the recipient.

Both branches resolve `fee_payer_pubkey` from `methodDetails.feePayerKey` rather than `account_keys[0]`, matching rust's `expected_fee_payer` invariant.

Tests added in `python/tests/test_server.py::TestFeePayerSourceDrainProtection`:
- `test_sol_drain_with_fee_payer_as_source_is_rejected`
- `test_spl_drain_with_fee_payer_ata_as_source_is_rejected`
- `test_spl_token_2022_drain_with_fee_payer_ata_as_source_is_rejected`
- `test_legitimate_spl_payment_with_fee_payer_cosign_is_accepted`

@greptileai review

EfeDurmaz16:
@greptileai review

tests/interop/python-server/main.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in dd6ff27 (with b9546ed): the harness now constructs a fresh `SolanaRpc` inside each request handler instead of sharing one across event loops. Resolving.

python/src/solana_mpp/_canonical_json.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 95945b4: the unreachable `_encode_number` branch was dropped; the live int path is the single `str(value)` return. Resolving.

python/src/solana_mpp/_rpc.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in f18345b: `_id` was replaced with `itertools.count`, whose `next` is atomic in CPython and works across event loops where an `asyncio.Lock` would not. Resolving.

python/src/solana_mpp/store.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 12d66cb: `put_if_absent` now flushes (`_flush_off_loop(next_data)`) BEFORE assigning `self._data = next_data`, so a fsync failure leaves both in-memory and on-disk state unchanged. Resolving.

tests/interop/python-server/main.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 33de5ff: `MPP_INTEROP_FEE_PAYER_SECRET_KEY` is read via `os.environ.get`; when absent the adapter skips `fee_payer_signer` and does not advertise `feePayer=true`. Resolving.

python/src/solana_mpp/server/mpp.py
```diff
@@ -347,10 +468,468 @@ def _decode_legacy_payment_instructions(transaction_b64: str) -> list[dict[str,
...
+                 if rcpt == destination and amt == lamports),
+                -1,
+            )
+            if match_idx == -1:
+                raise PaymentError(
+                    "unexpected System Program transfer in payment transaction",
+                    code="invalid-payload",
+                )
+            remaining_transfers.pop(match_idx)
+            continue
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 5bf71d9 and 4e5a58e: System Program transfers whose source equals the server-derived fee-payer pubkey are rejected with `payment_invalid` before co-sign. The fee-payer pubkey used for drain detection comes from `Mpp._fee_payer_signer.pubkey()`, not from client-controlled `methodDetails.feePayerKey`. Resolving.

python/src/solana_mpp/server/mpp.py
```diff
@@ -657,42 +1293,71 @@ async def _verify_transaction(
...
+        # diagnostic.
+        # Pass the raw signature string straight through. The previous
+        # ``Signature.from_string(signature)`` call sat between the
+        # durable consume marker (above) and the get_transaction call;
+        # if that parse ever raised (malformed RPC response, future
+        # solders API change), the consume would be durable but no
+        # receipt would be issued, stranding the user. ``get_transaction``
+        # already calls ``str(signature)`` internally on the wire, so the
+        # conversion is redundant work on the post-consume critical path.
+        await self._rpc.await_confirmation(signature)
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in c96bcc1: `Mpp.__init__` now validates that `Config.rpc` exposes `send_raw_transaction`, `await_confirmation`, and `get_transaction`, so an incompatible client fails at construction instead of after the transaction has been broadcast and consumed. Resolving.

python/src/solana_mpp/server/mpp.py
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in 5bf71d9: `transferChecked` instructions whose authority equals the fee-payer pubkey OR whose source ATA is owned by the fee-payer are rejected with `payment_invalid`, mirroring the Rust spine. Resolving.

EfeDurmaz16:
## Final status (tip 0ed4bd2)

- Greptile: 5/5
- CI: all green
- Manual DX: Python walkthrough complete, 1 positive on-chain devnet signature (2dfi9DgV...) plus 2 attack rejects confirmed (DRAIN + SLOT)
- Codex review: applied earlier in cycle (3a0ca5c `_co_sign_with_fee_payer` slot guard + a3c6fea `FileReplayStore._flush` fsync), no open items
- Outstanding inline notes: none

Ready for review.

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
@greptileai review

EfeDurmaz16:
## Final status (tip c59daf1)

- Greptile: 5/5
- CI: all green
- Latest round: 5 fixes shipped (Config.rpc interface validation at init, middleware module-level import surfacing, happy-path test coverage, ruff format pass, module-level _RpcResponse extraction + L8 verifier docstring contract)
- Manual DX: Python walkthrough captured locally, on-chain signatures in mpp-sdk-self-learning/notes/verify-2026-05-24/
- Codex review round 3: 0 P1, 0 P2; no intentional-open inline threads remain
- Ready for re-review.

python/README.md
```diff
@@ -1,54 +1,278 @@
...
+
+### ASGI / Starlette middleware
+
+```python
+from solana_mpp.server.middleware import pay
+
+@app.get("/paid")
+@pay(mpp, amount="0.001", description="Paid endpoint")
+async def handler(request, credential, receipt):
+    return {"ok": True}
```
lgalabru, line 104: Can we follow the structure of Ruby here?
Would love to see a `example/simple-server/app.py` or equivalent, and a very clean flask example (the Sinatra ruby example is a good inspiration) - clean middleware isolation, etc.

python/README.md
```diff
@@ -1,54 +1,278 @@
...
+  <img src="https://github.com/solana-foundation/mpp-sdk/raw/main/assets/banner.png" alt="MPP" width="100%" />
+</p>
 
-Python implements Solana MPP helpers for server and client-side `charge` flows.
+# solana-mpp
 
-This package provides:
+Charge stablecoins (USDC, USDT, PYUSD, USDG, CASH) for any HTTP endpoint,
+in Python. Implements the Solana payment method for the
+[Machine Payments Protocol](https://mpp.dev).
```
lgalabru: Generally speaking, let's link to https://paymentauth.org

python/README.md
```diff
@@ -1,54 +1,278 @@
...
+structured `application/problem+json` body with the L6 canonical
+error code (`payment_invalid`, `signature_consumed`, ...) on any 402.
+
+## Running the example
 
 ```bash
 cd python
-uv sync --extra dev
+pip install -e ".[dev]"
+python examples/payment_link_server.py
```
lgalabru: Flask example would be more concrete.

python/README.md
EfeDurmaz16: Added `python/examples/simple-server/app.py` (Flask) mirroring the Ruby Sinatra example shape: `GET /health` is free, `GET /paid` is gated by an `@mpp_charge(amount=..., description=...)` decorator that inspects the `Authorization: Payment` header, returns a 402 with a signed `WWW-Authenticate: Payment` challenge on miss, and attaches the on-chain `Payment-Receipt` header on hit.

Verified locally:

```
$ curl -i http://127.0.0.1:8765/paid
HTTP/1.1 402 PAYMENT REQUIRED
WWW-Authenticate: Payment id="...", realm="Python Flask Example", method="solana", intent="charge", request="...", expires="..."
{"type": "https://paymentauth.org/problems/payment-required", ...}

$ curl -i http://127.0.0.1:8765/health
HTTP/1.1 200 OK
{"ok":true}
```

Shipped in 38b2217. The older `payment_link_server.py` stays as the interop adapter; the README now reframes the Running section around the Flask example.

python/README.md
EfeDurmaz16: Replaced the inline `mpp.dev` reference in `python/README.md` with `https://paymentauth.org` (the spec home). Shipped in 38b2217. The same pattern will land on PHP / Ruby / Lua / Go / Kotlin / Swift READMEs through each respective PR for cross-SDK consistency, so the spec home is the single link from every language entry point.

python/README.md
EfeDurmaz16: Same Flask example shipped at `python/examples/simple-server/app.py` (response to comment 3298536441). Verified locally with `curl -i /paid` returning 402 + Payment challenge and `/health` returning 200. Shipped in 38b2217.

.github/workflows/python.yml
```diff
@@ -0,0 +1,117 @@
...
+          pip install -e ".[dev]"
+      - name: Lint with ruff
+        working-directory: python
+        run: ruff check src tests
+      - name: Type check with pyright
+        working-directory: python
+        run: pyright
+      - name: Run tests with coverage
+        working-directory: python
+        # Coverage gate is 80% pending follow-up backfill of payment_page.py
```
lgalabru: Can we bump to 90?

EfeDurmaz16: Bumped python.yml gate to 90% line coverage (M1 baseline, matching Ruby and PHP). Dropped the obsolete 80% rationale comment and the `--cov-branch` flag (branch coverage gate tracked separately in #108). Added 7 new tests covering the v0 (VersionedTransaction) cosign and allowlist paths to clear the line gate. 4710309

python/examples/simple-server/app.py
```diff
@@ -0,0 +1,138 @@
...
+@app.get("/paid")
+@mpp_charge(amount=AMOUNT, description="Paid endpoint")
+def paid():
+    return jsonify(ok=True, message="thanks for paying!")
+
+
+if __name__ == "__main__":
+    host = os.environ.get("HOST", "127.0.0.1")
+    port = int(os.environ.get("PORT", "8000"))
+    app.run(host=host, port=port)
```
lgalabru: could we split this file? app, config, middleware + follow flask best practices?
and rename the parent folder (should be `examples/flask`)
And remove the Sinatra related comment :)

python/src/solana_mpp/protocol/solana.py
```diff
@@ -47,10 +62,13 @@
 
 def default_rpc_url(network: str) -> str:
     """Return the default RPC endpoint for a Solana network."""
-    if network == "devnet":
+    canonical = _canonical_network(network)
```
lgalabru: canonical -> network

python/examples/simple-server/README.md
```diff
@@ -0,0 +1,36 @@
...
+curl -i http://127.0.0.1:8000/paid
+# HTTP/1.1 402 Payment Required
+# WWW-Authenticate: Payment realm="Python Flask Example", ...
+```
+
+## Environment
+
+`HOST`, `PORT`, `MPP_RPC_URL`, `MPP_NETWORK`, `MPP_CURRENCY`,
+`MPP_PAY_TO`, `MPP_SECRET_KEY`, `MPP_AMOUNT`. Defaults match the Ruby
+Sinatra example so cross-language clients can hit either server with
```
lgalabru: no need to mention Sinatra here

python/examples/payment-links/server.py
```diff
@@ -2,6 +2,7 @@
 
 Listens on port 3004. Requires Surfpool on localhost:8899.
```
lgalabru, line 3: examples/payment_link_server.py -> examples/payment-links/server.py

python/src/solana_mpp/server/mpp.py
```diff
@@ -41,8 +44,48 @@
...
+# in PHP and Ruby. A high split count balloons the transaction size and
+# the per-recipient ATA verification cost, so we reject early at the
+# pre-broadcast stage.
+MAX_SPLITS = 8
+
+# Legacy Solana memo program (v1). MPP charge transactions MUST use memo v2
+# (``MEMO_PROGRAM`` from :mod:`solana_mpp.protocol.solana`). v1 had a different
+# instruction shape and is rejected to match the L2 lock landed on PHP fde0efb
+# and mirrored in Ruby, Rust, Lua.
+_MEMO_V1_PROGRAM = "Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo"
```
lgalabru, line 73: Remove?

python/src/solana_mpp/protocol/solana.py
EfeDurmaz16: Done in f4801a4 — both  and  now reassign the parameter (`network = _canonical_network(network)`) instead of binding a separate `canonical` local. https://github.com/solana-foundation/pay-kit/pull/106/commits/f4801a4

python/examples/simple-server/README.md
EfeDurmaz16: Dropped in 441734a, and the lead paragraph reference is also gone after the folder rename in a9fd6ea. https://github.com/solana-foundation/pay-kit/pull/106/commits/441734a

python/examples/simple-server/app.py
EfeDurmaz16: Done in a9fd6ea — folder is now `python/examples/flask/`, `app.py` is a `create_app(mpp, settings)` factory, env-driven `Config` lives in `config.py`, the `@mpp_charge` decorator moves to `middleware.py`, and the Sinatra mirror callouts are gone. https://github.com/solana-foundation/pay-kit/pull/106/commits/a9fd6ea

python/examples/payment-links/server.py
EfeDurmaz16: Moved in 0aea30a via `git mv` to `python/examples/payment-links/server.py`. Updated `python/README.md` and the Swift iOSDemo merchant-server doc-comment that pointed at the old path. https://github.com/solana-foundation/pay-kit/pull/106/commits/0aea30a

python/src/solana_mpp/server/mpp.py
EfeDurmaz16: Keeping this one — `_MEMO_V1_PROGRAM` is load-bearing for the L2 v1-reject guard. It is referenced on lines 257, 478, and 867 to reject any `Memo1Uhk...` instruction in a submitted transaction (the legacy non-UTF-8 memo program), which mirrors the Rust spine's reject set. Dropping the constant would silently allow v1 memo instructions through validation.

EfeDurmaz16:
Addressed all 5 review comments:

- `f4801a4` refactor(python/protocol): rename `canonical` local to `network` in solana helpers
- `441734a` docs(python/examples): drop sinatra reference from simple-server env section
- `a9fd6ea` refactor(python/examples): split flask example into app/config/middleware (folder also renamed to `examples/flask/`)
- `0aea30a` refactor(python/examples): move `payment_link_server.py` to `payment-links/server.py`
- `_MEMO_V1_PROGRAM` constant in `server/mpp.py` is kept — explained inline (load-bearing for the L2 v1-reject guard, used on lines 257/478/867)

`pytest --cov=src/solana_mpp --cov-fail-under=90 tests/` green: 444 passed, 92.81% coverage.

PR #134, opened 2026-05-26
test(harness/x402): x402 exact interop harness, intent + matrix

EfeDurmaz16:
## Summary

Adds the x402-exact interop harness in one mergeable unit: scenario contracts and env-var protocol, plus the full client/server cross-pair matrix and negative scenarios. Consolidates the prior split between #132 (intent) and #133 (matrix) for easier review.

### Harness intent + contracts
- `harness/src/intents/x402-exact.ts`, scenario catalogue + canonical-codes alignment
- `harness/src/contracts.ts`, `InteropScenario` extension for `paymentVersion`, `paymentMode`, `expectedHeader`
- `harness/src/x402-pair-policy.ts`, `allowedPair` helper (data-driven matrix)
- Env-var contract: `X402_INTEROP_TARGET_URL`, `X402_INTEROP_RPC_URL`, etc. (see `harness/README.md`)

### Reference fixture
- `harness/src/fixtures/typescript/exact-{client,server,shared}.ts`, TS stub fixture (challenge framing only, no real settlement)

### Test matrix (3 tiers)
- `harness/test/x402-exact.compat.test.ts`, 17 compat tests, no RPC, runs everywhere
- `harness/test/x402-exact.e2e.test.ts`, full e2e cross-pairs, Surfpool-gated
- `harness/test/x402-exact.live.matrix.test.ts`, live-matrix, keypair + Surfpool gated
- `harness/test/x402-exact.negative.test.ts`, network-mismatch + cross-route-replay
- `harness/test/cross-server-scenarios.test.ts`, symmetric portability including ts-x402 to ts-x402

## Security highlights
- Replay bypass closed: drops the `issued.size > 0` guard that let fresh-server scenarios bypass deduplication.
- Network-mismatch scenario now actually creates a network mismatch (distinct envs on client vs server).
- Cross-route-replay credential dispatch matches the Rust spine's `charge_request_mismatch`.

## Test evidence
- 17/17 compat tests pass without RPC.
- Cross-pair matrix exercises 7 SDKs by 7 SDKs where both client and server roles exist.
- Codex r8: 0 P1.

## Supersedes
- #132 (x402-harness-intent)
- #133 (x402-harness-matrix)

PR #141, opened 2026-05-27
feat(lua): PayKit umbrella for OpenResty / Kong / APISIX (closes #140)

### LINKED ISSUE #140: Lua - Interface for PayKit
# solana-pay-kit (Lua) — SDK Design Guidelines

Working notes for the surface that unifies **x402** and **MPP** behind a
single Lua-idiomatic API. Sibling to `ruby/DESIGN.md`, `python/DESIGN.md`,
`go/DESIGN.md`, and `php/DESIGN.md`; same model, Lua / OpenResty idioms.

## Intent

- One module, one surface, two protocols underneath (x402, MPP).
  Operators shouldn't have to care which protocol settled a request
  unless they ask.
- **OpenResty-first.** Lua's primary HTTP server-side environment is
  LuaJIT inside nginx (OpenResty, Kong, APISIX). The lib is built for
  that substrate; plain-Lua use is supported via pluggable transports
  but is not the headline.
- **Kong-friendly out of the box.** A ready-to-install Kong plugin ships
  in the same repo and is a thin shim over the core lib — installing
  the plugin should never require knowing Lua.
- **Drop into existing `access_by_lua_*` blocks.** No special directives,
  no DSL, no patched nginx build. The lib is a regular Lua module.

## Naming

LuaRocks package and module path follow the OpenResty community split
— `lua-resty-<name>` package, `resty.<name>` import:

| Surface                  | Name                                          |
|--------------------------|-----------------------------------------------|
| LuaRocks (core lib)      | `lua-resty-pay-kit`                           |
| Module path              | `local pay_kit = require("resty.pay_kit")`    |
| LuaRocks (Kong plugin)   | `kong-plugin-pay-kit`                         |
| Kong plugin id           | `pay-kit` (the value in `KONG_PLUGINS=...`)   |
| LuaRocks (APISIX plugin) | `apisix-plugin-pay-kit`                       |
| Sub-modules              | `resty.pay_kit.signer`, `resty.pay_kit.kms`, … |

Rationale: `lua-resty-` package prefix is the OpenResty community
convention (`lua-resty-openidc`, `lua-resty-http`, `lua-resty-redis`,
`lua-resty-jwt`); the `resty.*` module namespace signals "designed for
OpenResty's cosocket / `ngx.*` runtime." Keeping it shorter than
`solana-pay-kit` matches what every other resty lib already does on
LuaRocks. Bare `pay_kit` (no `resty.` prefix) is also a valid alias for
pure-Lua use, but `resty.pay_kit` is the canonical path.

The current `mpp` LuaRocks package is the MPP-only predecessor; this
design supersedes it and consolidates both schemes. Migration via a
shim module that maps `require("mpp")` → `require("resty.pay_kit")`
with a `ngx.log(ngx.WARN, ...)` deprecation notice, removed after one
minor release.

## Inspirations

- **[lua-resty-openidc](https://github.com/zmartzone/lua-resty-openidc)**
  — the reference for "drop into `access_by_lua_block` and gate a
  route." We borrow the *single-entry-point shape*
  (`openidc.authenticate(opts)` → `pay_kit.require_payment(name)`),
  the `lua_shared_dict` pattern for replay protection, and the
  `(result, err)` return convention.
- **[lua-resty-http](https://github.com/ledgetech/lua-resty-http)** — the
  reference for resty client design: instance pattern
  (`local httpc = http.new()`), connection-pool-aware, cosocket-native.
  Our Solana RPC client follows the same shape.
- **[Kong plugin development guide](https://developer.konghq.com/custom-plugins/)**
  — canonical `schema.lua` + `handler.lua` layout, phase methods
  (`access`, `header_filter`, `log`), `PRIORITY` + `VERSION` constants,
  `kong.ctx.plugin` for per-request plugin state, `kong.ctx.shared`
  for cross-plugin state. The pay-kit Kong plugin follows this
  template literally.
- **[APISIX plugin model](https://apisix.apache.org/docs/apisix/plugin-develop/)**
  — sibling to Kong, also OpenResty-based, also `access(conf, ctx)`
  phase methods. Same lib backs both plugins; the only difference is
  the wrapping shim.
- **[lua-resty-jwt](https://github.com/SkyLothar/lua-resty-jwt)** — the
  `jwt:verify(secret, jwt_str)` shape and the way it carries HMAC
  secrets through the call. We mirror the discipline.
- **[resty.openssl](https://github.com/fffonion/lua-resty-openssl)** —
  modern OpenSSL bindings for OpenResty. Supports Ed25519 sign/verify
  and HMAC-SHA256. The crypto substrate we build on; the existing
  `mpp` package's `luasodium` dependency moves here.

## Vocabulary

Same shared vocabulary as the other four languages, with the Lua /
OpenResty runtime mapping called out explicitly:

| **operator**    | Merchant identity: recipient + signer + fee-payer flag.              |
| **signer**      | Ed25519 key source — `demo`, `bytes`, `file`, `from_env`, `kms.gcp`, …|
| **amount**      | The base amount a gate charges, before any `fee_on_top`.             |
| **total**       | What the customer pays: `amount + sum(fee_on_top)`. Derived.         |
| **price**       | Value object returned by `pay_kit.usd(...)`: number + denom + settlement. |
| **fee_within**  | Fee taken *out of* the amount. `pay_to` recipient nets less.         |
| **fee_on_top**  | Fee *added to* the amount. Customer pays more; `pay_to` nets full.   |
| **payment**     | Proof submitted by the client to pass a gate. Stored in `ngx.ctx.pay_kit_payment`. |
| **scheme**      | Protocol used to prove payment: `"x402"`, `"mpp"`.                   |
| **accept**      | Ordered preference list — applies to both schemes and stablecoins.   |
| **denom**       | The fiat unit a price is quoted in: `"USD"`, `"EUR"`.                |
| **settlement**  | The on-chain asset that actually transfers: `"USDC"`, `"USDT"`.      |

Lua-runtime mapping:

| Concept (cross-language) | Lua / OpenResty equivalent                              |
|--------------------------|---------------------------------------------------------|
| Boot-time `configure()`  | `init_by_lua_block` (master) or `init_worker_by_lua_block` (per worker) |
| Per-request state        | `ngx.ctx.pay_kit_payment`, `ngx.ctx.pay_kit_gate`       |
| Cross-worker shared state | `ngx.shared.pay_kit_replay` (declared as `lua_shared_dict`) |
| Halting a request        | `ngx.exit(ngx.HTTP_PAYMENT_REQUIRED)` (called by the lib) |
| Middleware insertion     | `access_by_lua_block` / `access_by_lua_file`            |
| Kong equivalent          | Kong plugin's `:access(conf)` phase method              |

### Boot-time configuration

Single function, table-based options — the universal Lua / OpenResty
config shape. Goes in `init_by_lua_block` (runs once at master start;
state survives all workers via Lua module caching). Production form:

```nginx
http {
  lua_package_path  '/path/to/lua-resty-pay-kit/?.lua;;';
  lua_shared_dict   pay_kit_replay  10m;   # MPP nonce replay store

  init_by_lua_block {
    local pay_kit = require("resty.pay_kit")
    local signer  = require("resty.pay_kit.signer")

    local ok, err = pay_kit.configure({
      network     = "solana_devnet",
      accept      = { "x402", "mpp" },           -- preference order
      stablecoins = { "USDC", "USDT" },          -- preference order
      rpc_url     = os.getenv("PAY_KIT_RPC_URL"), -- nil → default per network

      operator = {
        recipient = os.getenv("PAY_KIT_OPERATOR_RECIPIENT"), -- nil → signer.pubkey
        signer    = signer.from_env("PAY_KIT_OPERATOR_KEY"), -- nil → signer.demo()
        fee_payer = true,
      },

      -- OPTIONAL — when set, PayKit POSTs to this URL's /verify and /settle.
      -- When nil, PayKit runs x402 locally using rpc_url + operator.signer.
      x402 = {
        facilitator_url = os.getenv("PAY_KIT_X402_FACILITATOR_URL"),
        scheme          = "exact",
      },

      mpp = {
        realm                    = "MyApp",
        challenge_binding_secret = os.getenv("PAY_KIT_MPP_CHALLENGE_BINDING_SECRET"),
        expires_in               = 300,            -- seconds
      },
    })
    if not ok then ngx.log(ngx.ERR, "pay_kit.configure: ", err); os.exit(1) end

    -- Gates registered alongside configure — see "Gates" below.
    pay_kit.gate("report",  { amount = pay_kit.usd("0.10") })
    pay_kit.gate("paywall", { amount = pay_kit.usd("0.50") })
  }

  server {
    location = /report {
      access_by_lua_block {
        require("resty.pay_kit").require_payment("report")
      }
      proxy_pass http://backend;
    }
  }
}
```

Near-zero config — for smoke tests / examples, only `network` is
required and everything else falls back to the in-process demo signer:

```lua
require("resty.pay_kit").configure({ network = "solana_localnet" })
```

There is no top-level `pay_to` and no `x402.facilitator_secret_key` —
both cascade from `operator`. See **Operator** below.

`pay_kit.configure(opts)` returns `(true, nil)` on success or
`(nil, err)` on validation failure (mixed denoms, missing required
fields, demo signer on mainnet). Calling it twice in the same worker
returns an error — configuration is immutable after init. For dynamic
gates (price depends on the request), see **Gates** below.

### Operator — merchant identity in one place

An **operator** bundles the three things a merchant brings to the protocol:

- `recipient` — where settled funds land. Default `pay_to` for every gate.
- `signer` — the Ed25519 keypair used to sign x402 facilitator
  challenges, and (if `fee_payer = true`) to pay Solana network fees.
- `fee_payer` — whether the operator's signer also pays Solana network
  fees on settlement transactions.

```lua
operator = {
  recipient = "Cs2zdfUNonRdRGsiZUQQLdTxzxVvJZmgiX2mpLYKuEqP",
  signer    = signer.file("/etc/paykit/operator.json"),
  fee_payer = true,
}
```

#### Defaults

The operator table is optional. If you never set `opts.operator`, you
get:

```lua
{
  recipient = nil,                  -- → resolved to signer.pubkey at configure() time
  signer    = signer.demo(),        -- → hard-coded demo keypair
  fee_payer = true,
}
```

`signer.demo()` returns the package-shipped demo keypair (constant,
base58 pubkey: `AyNAa2VPe2t5pgg8M61iE6kqMudkV98zsT4rkAZuU6tj`). The
library calls `ngx.log(ngx.WARN, ...)` once at configure when the demo
signer is in use, and returns `(nil, "pay_kit: demo signer cannot be
used on solana_mainnet")` if combined with `network = "solana_mainnet"`.

#### Lua "nil-as-no-opinion" pattern

Lua's equivalent of Ruby's "nil-as-no-op setter convention" and Go's
"zero-value-as-no-opinion" is **nil-as-no-opinion in option tables**.
Setting a field to `nil` is the same as leaving it out — the configure
resolver fills in defaults. So `os.getenv` reads compose cleanly:

```lua
operator = {
  recipient = os.getenv("PAY_KIT_OPERATOR_RECIPIENT"),     -- nil → signer.pubkey
  signer    = signer.from_env("PAY_KIT_OPERATOR_KEY"),     -- nil → signer.demo()
  fee_payer = true,
}
```

`os.getenv` returns `nil` for unset vars, `signer.from_env` returns
`nil` for unset vars; both compose into the defaulting logic without
`if env then ... end` guards in user code.

#### Signer backends

Factories split by execution model. Local/in-process signers live in
`resty.pay_kit.signer` and ship in v1. Remote enclave signers (KMS,
Vault, HSMs) are reserved under `resty.pay_kit.kms` but **not part of
the initial release** — they're sketched here so the module path is
locked in before anyone writes against `signer.gcp_kms(...)` and we
have to rename later.

Both halves return tables conforming to the same duck-type contract:

```lua
-- A signer is any table with these three keys:
{
  pubkey    = function(self) return "<base58 address>" end,
  sign      = function(self, message) return "<64-byte signature>", err end,
  fee_payer = function(self) return true end,
}
```

`sign(self, message)` may yield (cosocket I/O) for KMS-backed signers;
local signers never yield. Callers don't have to care — OpenResty
cosockets make sync-style code cooperate with the event loop
automatically. No async/await syntax needed.

```lua
local signer = require("resty.pay_kit.signer")

-- === Local — synchronous, no I/O on :sign. Ships in v1. ===

signer.demo()                                          -- hard-coded demo keypair
signer.bytes({ 1, 2, 3, --[[ ..., 64 ]] })             -- raw 64-byte secret (table of ints)
signer.json("[1,2,3,...,64]")                          -- Solana-CLI JSON-array format
signer.base58("4PzYg...wuFNQT")                        -- Phantom / Solflare export
signer.hex("abcd1234...")                              -- 128-char hex string
signer.file("/etc/paykit/operator.json")               -- 64-byte JSON array on disk
signer.from_env("PAY_KIT_OPERATOR_KEY")                -- env var (auto-detects format)
signer.generate()                                      -- fresh ephemeral keypair (tests only)

-- === Remote enclave — async, cosocket I/O on :sign. FUTURE WORK. ===
-- Sketched here to reserve the module path. Not in v1.

local kms = require("resty.pay_kit.kms")

kms.gcp({ key_name = "...", pubkey = "..." })          -- not implemented yet
kms.aws({ key_id = "...", region = "...", pubkey = "..." })  -- not implemented yet
kms.vault({ addr = "...", path = "...", pubkey = "..." })    -- not implemented yet
```

Remote signers all take a `pubkey` field explicitly so configure
doesn't probe the enclave at boot. Misconfiguration surfaces at the
first `:sign()` call, not at startup — same as the other languages.

##### `signer.from_env(name)` contract

`signer.from_env("VAR_NAME")` returns:

- A signer table, parsed from the env var's contents (auto-detecting
  the encoding: Solana-CLI JSON array, base58, or hex — same formats
  the explicit `signer.json` / `signer.base58` / `signer.hex` factories
  support).
- `nil` (only, no error) if the env var is unset or empty.
- `nil, err` if the env var is set but malformed. Silent fallback on
  malformed input would mask bugs.

The plain-`nil` return on unset composes with the operator's
nil-as-no-opinion behavior — `operator = { signer = signer.from_env(…) }`
leaves the default `signer.demo()` in place when the env var is unset.

For boot-time `var`-block-equivalent use (where you want a hard error
on missing), use a small wrapper:

```lua
operator = {
  signer = assert(signer.from_env("PAY_KIT_OPERATOR_KEY"),
                  "PAY_KIT_OPERATOR_KEY required"),
}
```

#### Cascading

The operator is one source of truth for things that used to be
scattered:

| Was                                   | Now                                          |
|---------------------------------------|----------------------------------------------|
| `opts.pay_to`                         | `opts.operator.recipient`                    |
| `opts.x402.facilitator_secret_key`    | `opts.operator.signer` (when running x402)   |
| (manual SOL fee management)           | `opts.operator.fee_payer = true`             |

Env-var conventions follow the same rename. Old names trigger
`ngx.log(ngx.WARN, ...)` at configure-time pointing at the new name;
removed after one minor release.

| Old                                | New                                       |
|------------------------------------|-------------------------------------------|
| `PAY_KIT_PAY_TO`                   | `PAY_KIT_OPERATOR_RECIPIENT`              |
| `PAY_KIT_X402_FACILITATOR_KEY`     | `PAY_KIT_OPERATOR_KEY`                    |
| `PAY_KIT_X402_FACILITATOR` *       | `PAY_KIT_X402_FACILITATOR_URL`            |
| `PAY_KIT_MPP_SECRET`               | `PAY_KIT_MPP_CHALLENGE_BINDING_SECRET`    |

\* The old `PAY_KIT_X402_FACILITATOR` value in demo configs was actually
a Solana RPC URL, not a facilitator URL — that data should move to the
new `PAY_KIT_RPC_URL` and the facilitator var should be left unset
unless a real x402 facilitator is being pointed at.

The MPP rename tracks the spec's vocabulary —
`draft-httpauth-payment-00` §5.1.2.1.1 calls this the "server secret"
used for "stateless challenge binding" (HMAC-SHA256 over challenge
parameters).

Gate-level `pay_to` still overrides the operator's recipient (the
marketplace pattern — operator signs, seller is paid).

#### Rules

1. **`operator.recipient` defaults to `operator.signer:pubkey()`** when
   `nil`. The signer always has a pubkey, so this is a safe default.
2. **Gate `pay_to` overrides `operator.recipient`** per gate.
3. **`operator.signer` is the x402 facilitator key.** Setting
   `opts.x402.signer` is the escape hatch; not advertised in the
   getting-started docs.
4. **`operator.fee_payer = true` (default)** means the operator's
   signer pays Solana network fees on settlement. Set `false` when
   fees come from elsewhere.
5. **MPP secret stays separate.** MPP signs HMACs, not Ed25519, so
   `mpp.challenge_binding_secret` remains explicit — KMS-backed
   signers don't apply (the spec calls this the "server secret" /
   "shared secret" used for "stateless challenge binding"; see
   `draft-httpauth-payment-00` §5.1.2.1.1).
6. **Demo signer warns on every `configure`.** `configure` returns
   `(nil, err)` if the resolved signer is `signer.demo()` and
   `network = "solana_mainnet"`.

### Chain access — `rpc_url` and x402 modes

Two separate concerns that have been historically confused: **how
PayKit talks to Solana** (`opts.rpc_url`) and **whether PayKit handles
x402 verification/settlement itself** (`opts.x402.facilitator_url`).
Different fields, different layers.

#### `rpc_url` — Solana RPC URL

The HTTP endpoint of a Solana RPC node. Used by:

- MPP, always — verifying that a submitted settlement transaction
  landed on-chain with the right recipient and amount (via the
  cosocket-backed `resty.pay_kit.solana.rpc` client).
- x402 self-hosted mode — same plus tx construction and submission.
- x402 delegated mode — **not used**; the facilitator owns the RPC.

`rpc_url` is optional. When `nil`, configure picks a default from
`network`:

| `network`             | Default `rpc_url`                        |
|-----------------------|------------------------------------------|
| `"solana_mainnet"`    | `https://api.mainnet-beta.solana.com`    |
| `"solana_devnet"`     | `https://api.devnet.solana.com`          |
| `"solana_localnet"`   | `http://localhost:8899`                  |

Override with any custom RPC — Helius, QuickNode, private validator.
The public Solana RPCs are rate-limited and unsuitable for production
traffic; `configure` emits `ngx.log(ngx.WARN, ...)` when
`network = "solana_mainnet"` and `rpc_url` resolves to the public default.

The HTTP transport behind the RPC client is `lua-resty-http` by default
in OpenResty contexts, with `luasocket + luasec` as the fallback for
pure-Lua use. A custom transport function can be plugged in via
`opts.rpc_transport` — same pattern the existing `mpp.solana.rpc`
already uses.

#### `x402.facilitator_url` — optional delegation

Per the x402 spec, a *facilitator* is **a server that facilitates
verification and execution of payments for one or many networks**
(coinbase/x402 README). It exposes `/verify` and `/settle` HTTP
endpoints; the resource server (merchant) POSTs payment payloads and
the facilitator handles the RPC, gas/fees, and on-chain submission.
The merchant never sees the chain.

Two modes:

```lua
-- Delegated: PayKit POSTs to the facilitator. No RPC needed for x402.
x402 = { facilitator_url = "https://facilitator.example.com" }

-- Self-hosted: PayKit runs verification and settlement itself.
-- facilitator_url left nil; rpc_url + operator.signer do the work.
x402 = { facilitator_url = nil }   -- the default
```

Default is **self-hosted** (`nil`). Setting `facilitator_url` to any
URL opts into delegation. When delegated:

- `rpc_url` is unused by x402 (still used by MPP if enabled).
- `operator.signer` is still used to authenticate to the facilitator
  if the facilitator requires merchant identity (some don't; depends
  on the facilitator's auth scheme).
- `operator.fee_payer` is ignored by x402 — the facilitator handles
  fees per its own policy.

The existing nginx example's
`url = 'https://402.surfnet.dev:8899'` is a Solana validator RPC (port
8899), not a facilitator. After the migration, that value moves to
`rpc_url` and `x402.facilitator_url` stays nil in the demo.

### Price helper — denomination vs. settlement

```lua
local pay_kit = require("resty.pay_kit")

pay_kit.usd("0.10")                                   -- configured default stablecoins
pay_kit.usd("0.10", "USDC")                           -- narrow to one
pay_kit.usd("0.10", "USDC", "USDT")                   -- preference order
pay_kit.usd("0.10", unpack(opts.stablecoins))         -- splat from a table
```

Amounts use **integer micro-units internally** (no floats, no decimal
library overhead in LuaJIT). `usd("0.10")` parses into `100000` (10 ¢ ×
1e6 USDC base units). Float inputs are rejected at the call site —
pass a string, not `0.10`. Same anti-footgun discipline as the other
languages.

Future-proofed: `pay_kit.eur("0.20", "EURC")`, `pay_kit.gbp(...)`,
etc. slot in identically.

### Gates — registered at init phase

Two patterns, both supported:

**Registered gates** — the canonical form. Call `pay_kit.gate(name,
opts)` in `init_by_lua_block` after `configure`. The registry is
immutable after init; subsequent calls return an error.

```lua
pay_kit.gate("report", {
  amount      = pay_kit.usd("0.10"),
  description = "Premium report",
})

pay_kit.gate("api_call", {
  amount = pay_kit.usd("0.001", "USDC"),
  accept = { "x402" },
})

pay_kit.gate("paywall", {
  amount = pay_kit.usd("0.50"),
  accept = { "mpp", "x402" },   -- MPP preferred
})

-- Inclusive fee — taken out of the amount. MPP only (x402 auto-disabled).
pay_kit.gate("marketplace_sale", {
  amount     = pay_kit.usd("10.00"),
  pay_to     = SELLER_WALLET,
  fee_within = { [PLATFORM_WALLET] = pay_kit.usd("1.00") },
})

-- Surcharge — added on top of the amount.
pay_kit.gate("ticket", {
  amount     = pay_kit.usd("10.00"),
  pay_to     = SELLER_WALLET,
  fee_on_top = { [PLATFORM_WALLET] = pay_kit.usd("0.50") },
})
```

**Inline gates** — for one-off endpoints or dynamic pricing. Pass the
gate table directly to `require_payment` instead of a name:

```lua
-- access_by_lua_block:
require("resty.pay_kit").require_payment({
  amount      = pay_kit.usd("0.25"),
  description = "One-off",
})
```

**Dynamic gates** — for prices that depend on the request, register a
function. It's called at access time with the parsed request context:

```lua
pay_kit.gate("tiered", function(req)
  if req.query.tier == "premium" then
    return { amount = pay_kit.usd("5.00") }
  else
    return { amount = pay_kit.usd("0.10") }
  end
end)
```

### Amount and fees

A gate has one **amount** (the base it charges) and zero or more fees.
Two fields cover every real-world case; each is a `{ [address] = price }`
table, so one or many recipients use the same syntax:

- `fee_within = { ... }` — taken **out of** the amount. Customer pays
  the amount; the `pay_to` recipient nets less.
- `fee_on_top = { ... }` — added **on top of** the amount. Customer
  pays `amount + fee`; the `pay_to` recipient nets the full amount.

```lua
pay_kit.gate("sale", {
  amount     = pay_kit.usd("10.00"),
  pay_to     = SELLER,
  fee_within = { [PLATFORM] = pay_kit.usd("0.30") },
})
-- Customer pays $10.00 ▸ SELLER nets $9.70 ▸ PLATFORM $0.30

pay_kit.gate("ticket", {
  amount   = pay_kit.usd("10.00"),
  pay_to   = SELLER,
  fee_on_top = { [PLATFORM] = pay_kit.usd("0.30"), [GATEWAY] = pay_kit.usd("0.20") },
})
-- Customer pays $10.50 ▸ SELLER nets $10.00 ▸ PLATFORM $0.30 ▸ GATEWAY $0.20
```

Rules (validated at `pay_kit.gate(...)` registration time; the call
returns `(nil, err)` on violation, so a bad registration in
`init_by_lua_block` kills the master start):

1. **Fixed amounts only.** No basis points, no percentages, no rounding policy.
2. **`pay_to` is optional and defaults to `operator.recipient`.** Most
   gates omit it; marketplace gates set it to route to a seller.
3. **All amounts share one denomination.** Mix `pay_kit.usd(...)` and
   `pay_kit.eur(...)` and registration returns `nil, "mixed denoms"`.
4. **`sum(fee_within values) <= amount`.** The `pay_to` recipient
   can't end up with a negative payout.
5. **x402 is automatically disabled when fees are present.** Stock
   x402 facilitators settle to a single address. The resolver strips
   `"x402"` from `accept` silently. Explicitly setting
   `accept = {"x402"}` on a gate with fees returns
   `nil, "x402 incompatible with multi-recipient gates"`.
6. **Stablecoin preference is gate- or config-level**, not per-fee.
   Per-recipient override via the amount form
   (`pay_kit.usd("3.00", "USDC")`) is the escape hatch.

### The trio — `require_payment` / `paid_for` / `payment`

Mirrors Ruby's `require_payment!` / `paid?` / `payment` and the
prefix-verb forms in Python/Go/PHP, adapted to OpenResty's
`ngx.exit`-driven control flow. Same surface, three functions:

```lua
local pay_kit = require("resty.pay_kit")

-- access_by_lua_block — halts the request with 402 on failure:
pay_kit.require_payment("report")
-- If we get here, payment is valid. Stored on ngx.ctx.pay_kit_payment.

-- access_by_lua_block — opportunistic, free route observes the payment:
local payment = pay_kit.payment()        -- nil if unauthenticated
if payment and pay_kit.paid_for("bulk_report") then
  ngx.req.set_header("X-Premium", "1")
end
```

| Function                              | Returns       | On failure |
|---------------------------------------|---------------|------------|
| `pay_kit.require_payment(name)`       | `payment` table; sets `ngx.ctx.pay_kit_payment` | writes 402 + calls `ngx.exit`, never returns |
| `pay_kit.require_payment(inline_gate)`| same          | same |
| `pay_kit.payment()`                   | `payment \| nil` | never |
| `pay_kit.paid()`                      | `bool`        | never |
| `pay_kit.paid_for(name)`              | `bool`        | never |

`require_payment` deliberately does **not** follow the
`(result, err)` return convention — it follows
`lua-resty-openidc.authenticate(opts)`'s precedent, which calls
`ngx.exit` internally on failure. Reasoning: at the access-phase call
site, every caller wants the same response (write 402, halt). Hoisting
that into the lib avoids `if err then ngx.exit(402) end` boilerplate at
every gated location. For callers that genuinely want to handle the
402 themselves, the underlying `pay_kit.try_payment(name)` form returns
`(payment, err)` and never calls `ngx.exit`.

```lua
-- For the rare case where you want custom 402 handling:
local payment, err = pay_kit.try_payment("report")
if err then
  ngx.status = 402
  ngx.header["X-Custom"] = "yes"
  ngx.say(my_custom_402_body(err))
  return ngx.exit(402)
end
```

### Replay protection — `lua_shared_dict pay_kit_replay`

MPP needs to track consumed challenge nonces to prevent replay. The lib
auto-detects an nginx `lua_shared_dict` named `pay_kit_replay` and uses
it as the cross-worker replay store. The recommended nginx directive:

```nginx
http {
  lua_shared_dict pay_kit_replay 10m;   # ~70k entries at default eviction
}
```

If the directive is missing, the lib logs `ngx.log(ngx.WARN, ...)` once
at configure and falls back to a per-worker LRU. The per-worker
fallback works for single-worker dev setups but allows replay across
workers in production — the warning makes that explicit.

For pure-Lua use (no OpenResty), pass `opts.replay_store` to
`configure` — any table with `:get(key)`, `:set(key, value, ttl)`, and
`:delete(key)` methods works. The existing `mpp.store` module already
provides `memory()` and `shared_dict()` implementations.

### Kong plugin

The `kong-plugin-pay-kit` LuaRocks package wraps `resty.pay_kit` in
Kong's plugin envelope. Lives in the same monorepo under
`kong/plugins/pay-kit/`. Two files, by Kong convention:

```
kong/plugins/pay-kit/
├── handler.lua    -- phase methods (access, header_filter, log)
├── schema.lua     -- declarative plugin config schema
└── api.lua        -- (optional) admin API for inspecting cached state
```

**`handler.lua`** — thin wrapper around `resty.pay_kit`:

```lua
local pay_kit = require("resty.pay_kit")

local PayKitHandler = {
  PRIORITY = 1000,         -- after auth plugins (1003-1006), before rate-limiting
  VERSION  = "0.1.0",
}

function PayKitHandler:init_worker()
  -- Worker-level pay_kit.configure() is called from a dedicated
  -- KONG_NGINX_HTTP_TEMPLATE init block, not here (init_worker fires
  -- per worker; configure() is master-init). This phase is a no-op
  -- in the typical case.
end

function PayKitHandler:access(conf)
  -- conf is per-route plugin config from schema.lua. The gate either
  -- exists in the registry (registered at init) or is inline.
  local gate = conf.gate or {
    amount      = pay_kit.usd(conf.amount, unpack(conf.stablecoins or {})),
    pay_to      = conf.pay_to,
    accept      = conf.accept,
    description = conf.description,
  }
  pay_kit.require_payment(gate)
  -- pay_kit.require_payment writes the 402 + calls ngx.exit on failure;
  -- if we're still here, kong.ctx.shared.pay_kit_payment is set for
  -- downstream plugins (rate-limiting per-payer, logging, etc.) to use.
  kong.ctx.shared.pay_kit_payment = pay_kit.payment()
end

function PayKitHandler:log(_)
  -- (Optional) emit structured logs about scheme used, amount, recipient.
end

return PayKitHandler
```

**`schema.lua`** — Kong typedef config:

```lua
local typedefs = require("kong.db.schema.typedefs")

return {
  name = "pay-kit",
  fields = {
    { consumer  = typedefs.no_consumer },
    { protocols = typedefs.protocols_http },
    { config = {
      type = "record",
      fields = {
        -- Either reference a registered gate by name…
        { gate = { type = "string" } },
        -- …or specify an inline gate's fields:
        { amount      = { type = "string" } },
        { stablecoins = { type = "array", elements = { type = "string" } } },
        { accept      = { type = "array", elements = { type = "string", one_of = {"x402", "mpp"} } } },
        { pay_to      = { type = "string" } },
        { description = { type = "string" } },
      },
      entity_checks = {
        -- Either gate name or amount must be set:
        { at_least_one_of = { "gate", "amount" } },
      },
    }},
  },
}
```

**Operator / signer / RPC stay global**, not per-route — set them once
via env vars consumed by `pay_kit.configure` in a Kong
`KONG_NGINX_HTTP_TEMPLATE` init block. Per-route plugin config only
varies the gate; everything else is the operator's identity, which is
constant across routes.

Wiring the init block (typical Kong deployment):

```bash
# In your kong.conf or env vars:
KONG_NGINX_HTTP_INIT_BY_LUA_BLOCK="require('kong.plugins.pay-kit.init').setup()"
KONG_PLUGINS="bundled,pay-kit"
PAY_KIT_OPERATOR_RECIPIENT="..."
PAY_KIT_OPERATOR_KEY="..."
PAY_KIT_MPP_CHALLENGE_BINDING_SECRET="..."
```

The `kong.plugins.pay-kit.init` module reads the `PAY_KIT_*` env vars
and calls `pay_kit.configure(...)` once at master init. This keeps the
plugin config (per-route) cleanly separate from the operator config
(global).

### APISIX plugin

Bonus integration — the same `resty.pay_kit` lib backs an APISIX plugin
at `apisix/plugins/pay-kit.lua`:

```lua
local pay_kit = require("resty.pay_kit")
local plugin_name = "pay-kit"

local _M = {
  version  = 0.1,
  priority = 2000,
  name     = plugin_name,
  schema   = { ... },          -- APISIX JSON-schema form, parallels Kong's
}

function _M.access(conf, ctx)
  pay_kit.require_payment(conf.gate or {
    amount = pay_kit.usd(conf.amount),
    pay_to = conf.pay_to,
  })
end

return _M
```

The plugin shim is ~40 lines. The lib does the work.

### Pure-Lua usage (no OpenResty)

The lib still loads without `ngx`. When `ngx == nil`:

- `pay_kit.require_payment` returns `(payment, err)` instead of calling
  `ngx.exit` (no exit to call). Caller writes the 402 itself.
- Replay store defaults to in-memory LRU (no shared dict available).
- HTTP transport defaults to `luasocket + luasec` (no cosockets).
- `ngx.ctx.pay_kit_payment` substitute is a thread-local table (Lua
  coroutines or per-call return value, depending on host framework).

Concretely:

```lua
local pay_kit = require("resty.pay_kit")

pay_kit.configure({
  network          = "solana_devnet",
  rpc_transport    = function(url, body) return your_http(url, body) end,
  replay_store     = pay_kit.store.memory(),
  operator         = { signer = signer.file("/path/key.json") },
  mpp              = { challenge_binding_secret = "..." },
})

-- In your handler:
local payment, err = pay_kit.require_payment("report", { request = my_request })
if err then return write_402_response(my_response, err) end
```

Pure-Lua mode is supported and tested but isn't where the docs lead.
OpenResty + Kong is the headline.

### Errors

Errors are strings prefixed with `pay_kit:` for grep-ability. The shape
matches OpenResty conventions — `(nil, "pay_kit: payment required: <details>")`:

```lua
"pay_kit: payment required"
"pay_kit: invalid proof: <details>"
"pay_kit: challenge expired"
"pay_kit: scheme not supported"
"pay_kit: mixed denominations in gate"
"pay_kit: x402 incompatible with multi-recipient gates"
"pay_kit: demo signer cannot be used on solana_mainnet"
```

No exception classes — Lua's `error()` is for genuine bugs only, not
control flow. For tests, `pay_kit.errors` exports the literal strings
as constants so test code can `assert(err == pay_kit.errors.expired)`
without string-matching fragility.

### Layers

```
resty.pay_kit                       -- top-level: configure, gate, require_payment, payment, paid, paid_for, try_payment, errors
resty.pay_kit.signer                -- Signer factories: demo, bytes, json, base58, hex, file, from_env, generate
resty.pay_kit.kms                   -- Remote-enclave signer factories (FUTURE — reserved module path)
resty.pay_kit.solana.rpc            -- Cosocket-backed Solana JSON-RPC client (resty.http transport)
resty.pay_kit.store                 -- Replay store: memory(), shared_dict("name")
resty.pay_kit.schemes.x402          -- x402 adapter (self-hosted + delegated modes)
resty.pay_kit.schemes.mpp           -- MPP adapter
resty.pay_kit.util.base58           -- pure-Lua base58
resty.pay_kit.util.base64url        -- pure-Lua base64url
resty.pay_kit.util.json             -- canonical-JSON (RFC 8785) for MPP challenge binding
resty.pay_kit.util.crypto           -- thin wrapper over resty.openssl (Ed25519, HMAC-SHA256)
resty.pay_kit.internal.*            -- not part of the public API

kong/plugins/pay-kit/
├── handler.lua                     -- Kong plugin handler
├── schema.lua                      -- Kong plugin config schema
├── init.lua                        -- KONG_NGINX_HTTP_INIT_BY_LUA_BLOCK entrypoint
└── api.lua                         -- (optional) admin API extension

apisix/plugins/pay-kit.lua          -- APISIX plugin (single file)
```

No separate `lua-resty-pay-kit-kong` or `lua-resty-pay-kit-apisix`
LuaRocks packages — the gateway plugins live in their own packages
(`kong-plugin-pay-kit`, `apisix-plugin-pay-kit`) that depend on
`lua-resty-pay-kit`. Standard separation in the Lua ecosystem.

## Design rules — locked in

1. **A gate is the unit.** Amount alone undersells it — gates carry
   policy (accepted schemes), metadata (description), and optionally
   dynamic logic via a function value.
2. **Strings expand against config defaults; tables carry overrides.**
   Same hybrid on both axes — schemes (`{"x402"}`) and stablecoins
   (`{"USDC"}`). Strings cover ~90%.
3. **Order is semantic everywhere.** `accept` and `stablecoins` arrays
   are *preference order*, not sets. The 402 challenge advertises in
   this order.
4. **Immutable after `configure`.** Module-level state is set once at
   init phase and never mutated. Subsequent `configure` calls return
   an error. Gate registry is frozen after init.
5. **Denomination and settlement are separate.**
   `pay_kit.usd("0.10", "USDC", "USDT")` means "$0.10 USD, settle in
   USDC or USDT." Merchants think fiat.
6. **OpenResty first, plain Lua supported.** `ngx.exit` and
   `ngx.shared.DICT` are first-class; absent-ngx fallbacks exist but
   are not the headline. No framework is special among the OpenResty
   family — Kong, APISIX, plain OpenResty, lapis all consume the same
   `resty.pay_kit` module.
7. **`require_payment` halts via `ngx.exit`, `paid_for` returns bool,
   `payment` returns table-or-nil.** Mirrors the bang/predicate/accessor
   split from Ruby and the prefix verbs from Python. The
   `try_payment` form preserves the `(result, err)` escape hatch for
   callers that want custom 402 handling.
8. **Errors are strings with a `pay_kit:` prefix, exported as
   constants.** No exception classes; `error()` only for bugs. Apps
   compare against `pay_kit.errors.*` constants for stable handling.
9. **One source of truth per axis.** Stablecoin mints live in config
   with sensible mainnet/devnet defaults — operators shouldn't have to
   look up addresses.
10. **Amount + fees, never splits.** Multi-recipient gates are
    modelled as one `amount` plus `fee_within` / `fee_on_top` table
    fields. Fixed amounts only, MPP only. x402 auto-disabled on any
    gate with fees because stock x402 facilitators settle to a single
    address.
11. **Operator is the merchant identity, one source of truth.**
    Recipient, Ed25519 signer, and fee-payer flag live together on
    `opts.operator`. Gate-level `pay_to` and `opts.x402.signer` are
    escape hatches, not the primary surface. Zero config boots on the
    in-process demo signer with a visible warning; mainnet refuses to
    start under the demo key.
12. **Nil-as-no-opinion in option tables.** Setting a field to `nil`
    is the same as leaving it out — the configure resolver fills in
    defaults. `signer.from_env` returns plain `nil` for unset
    (no error), composing cleanly with the defaulting logic.
13. **Integer micro-units, never floats.** `usd("0.10")` parses to an
    integer in stablecoin base units. Floats are rejected at the call
    site to avoid the standard FP trap; same anti-footgun discipline
    as Ruby/Python/Go/PHP.
14. **Replay store via `ngx.shared.DICT` first.** Auto-detect
    `ngx.shared.pay_kit_replay`; warn + fall back to per-worker LRU
    when missing.
15. **Kong + APISIX shims are ≤ 100 lines each.** The core lib is the
    thing; gateway shims wrap it. New gateway integrations should be
    rejected from the core repo unless they're equally thin.

## Style — Lua micro-rules

- **Target LuaJIT 2.1** (OpenResty default). Lua 5.1 / 5.3 / 5.4
  compatibility maintained where free; if it conflicts, LuaJIT wins.
- **`local M = {}` module pattern.** `return M` at the bottom. No
  globals except via `ngx.*` reads.
- **`local ngx = ngx`** at module top to let the JIT compiler hoist
  the global lookup.
- **`snake_case` everywhere.** Module names, function names, table
  keys, config fields. No camelCase, no PascalCase.
- **`(result, err)` returns** for every function that can fail. `err`
  is a string, prefixed `pay_kit:`. Callers `if err then …`.
- **No exceptions for control flow.** `error()` is for bugs (asserts,
  invariants); `pcall` only at trust boundaries. Calls into user code
  are wrapped in `pcall` so a bad gate function doesn't crash the
  worker.
- **`luacheck`** in CI. Treat warnings as errors. The existing
  `.luacheckrc` stays.
- **`luacov` ≥ 90%** coverage gate, same as the existing `mpp`
  package.
- **No dependency on `bit32`** (Lua 5.2-only). Use `bit` (LuaJIT) or
  pure-Lua fallback (already in `mpp.util.bit`).
- **JSON via `cjson.safe`** — `cjson.safe.encode/decode` return
  `(nil, err)` instead of raising. Pair with `require_payment`'s
  error-string convention.
- **No `require` inside hot paths.** All `require` calls hoisted to
  the top of the module.

## Decisions

Calls made after surveying the Lua / OpenResty ecosystem. Each entry
locks in what's already shown in the surface above.

1. **Package name is `lua-resty-pay-kit` and module is `resty.pay_kit`.**
   Decided against keeping the existing `mpp` name. Reasoning: the
   `resty.*` namespace is the OpenResty community convention
   (`lua-resty-openidc`, `lua-resty-http`, `lua-resty-redis`,
   `lua-resty-jwt`); the unified x402+MPP surface needs a different
   name from the MPP-only predecessor; `pay-kit` matches the
   cross-language repo identity. `mpp` keeps working for one minor
   release via a shim module with a deprecation warning, then drops.

2. **`require_payment` halts via `ngx.exit` instead of returning
   `(result, err)`.** Decided to follow `lua-resty-openidc`'s
   precedent over OpenResty's typical `(result, err)` convention.
   Reasoning: every caller at the access-phase site wants the same
   behavior (write 402, halt); hoisting that into the lib removes
   boilerplate. `try_payment(name)` exists for the rare callers
   wanting custom 402 handling — they get `(payment, err)` and never
   see `ngx.exit`. Two functions, same underlying logic.

3. **`signer.from_env` returns plain `nil` on unset (no `(nil, nil)`).**
   Decided against the dual-nil form. Reasoning: plain `nil` composes
   directly with table-default resolution
   (`operator = { signer = signer.from_env("X") }`), no
   `local sgn, err = ...; if err then return nil, err end` ceremony.
   Set-but-malformed still returns `(nil, err)` because that's a real
   error, not absence. The pattern is "missing isn't an error, but
   broken is."

4. **Integer micro-units, no decimal library.** Decided against
   `decimal` or `bignum` Lua libs (`lua-bignum`, `lua-bc`). Reasoning:
   stablecoins have ≤ 9 decimals (USDC has 6, PYUSD has 6, USDT has
   6); micro-amounts fit in LuaJIT 2.1's 64-bit integers (`int64_t`
   via FFI cdata). No floats, no precision loss, no extra dep. Same
   call as the existing `mpp.util.uint` already makes.

5. **Single LuaRocks package for core; gateway plugins are separate
   packages.** `lua-resty-pay-kit` ships the lib;
   `kong-plugin-pay-kit` and `apisix-plugin-pay-kit` ship the
   gateway shims as thin dependents. Reasoning: Kong's `KONG_PLUGINS`
   discovery scans `kong/plugins/<name>/`, which requires that path
   to exist in the LuaRocks tree; same for APISIX. Keeping the
   gateway plugins in separate packages lets operators install just
   what they need, and lets the core lib evolve without dragging
   gateway compat concerns. Matches the
   `lua-resty-openidc` → `kong-plugin-openid-connect` precedent.

6. **`ngx.shared.pay_kit_replay` is auto-detected, not required.**
   Decided against making the directive a hard requirement.
   Reasoning: single-worker dev setups and tests don't need
   cross-worker state; the per-worker LRU fallback is fine for those.
   A boot-time `ngx.log(ngx.WARN, ...)` makes the choice visible. The
   nginx directive is documented as recommended for production.

7. **Crypto via `resty.openssl` (Ed25519 + HMAC-SHA256), not
   `luasodium`.** Decided to migrate from the existing
   `mpp`-package's luasodium dependency to `lua-resty-openssl`.
   Reasoning: `lua-resty-openssl` is the OpenResty community-default
   crypto binding (~1M downloads on LuaRocks), supports Ed25519
   (`EVP_PKEY_ED25519`) and HMAC-SHA256 (`EVP_MAC`) natively, ships
   with Kong and APISIX images, and integrates with OpenResty's
   cosocket model. `luasodium` works but is an extra system
   dependency (libsodium) that Kong/APISIX operators don't have by
   default. Pure-Lua fallback paths (for non-OpenResty use) stay
   pluggable via `opts.crypto`.

8. **Kong + APISIX are the only first-class gateway integrations
   shipped in-repo.** Decided against Tyk, KrakenD, custom-OpenResty
   shims. Reasoning: Kong + APISIX cover the OpenResty gateway market.
   The core lib is the contract; if a fifth gateway materializes,
   we add a one-file shim. Keeps the maintenance surface small.

## What we are not building (yet)

- Subscription / recurring billing.
- Custodial wallet management.
- Non-USD-pegged settlement (BTC/SOL-denominated pricing).
- Chargeback / refund flows.
- A CLI — `lua-resty-pay-kit` stays library-only; CLI is a separate
  package if needed.
- Tyk / KrakenD / Envoy gateway shims (see Decision #8).
- Pure-Lua HTTP server framework integration beyond OpenResty (lapis,
  sailor, pegasus). They run on OpenResty anyway, so the lib already
  works.

If a feature request lands in one of these buckets, push back hard
before expanding scope.

## Future work (post-v1)

Reserved here so the design has a place to grow without breaking
changes:

- **`resty.pay_kit.kms` module — remote enclave signers.** `gcp`,
  `aws`, `vault`. Cosocket-driven on `:sign(message)` (network I/O);
  share the signer duck-type contract so call sites don't change.
  Module path is locked in now; implementation lands after v1.
- **Additional fiat denoms.** `pay_kit.eur(...)`, `pay_kit.gbp(...)`
  slot into the existing `pay_kit.usd(...)` shape.
- **Additional remote-signer backends** (Turnkey, Privy, Fireblocks,
  HSMs) under `resty.pay_kit.kms.*`.
- **Per-consumer rate limiting integration.** Kong's built-in
  rate-limiting plugin can key on `kong.ctx.shared.pay_kit_payment.payer`
  once that field is populated by the access phase. Document the
  pattern; no plugin code needed.
- **Admin API endpoints** (`kong/plugins/pay-kit/api.lua`) for
  inspecting cached state — challenge issuance count, replay-store
  hit rate, last error. Useful for operators debugging in production.

EfeDurmaz16:
# PayKit umbrella port to Lua (OpenResty / Kong / APISIX)

Closes solana-foundation/pay-kit#140. Supersedes the earlier
exploration that became #134.

## What this delivers

A full PayKit umbrella for the LuaJIT 2.1 / OpenResty runtime, plus
Kong and APISIX plugin shims. The umbrella sits at `resty.pay_kit`
and re-exports the public surface from `DESIGN.md` Layers. The Ruby
gem (PR #138) is the closest reference; this PR matches feature
parity for the v1 self-hosted scope, with the verifier divergences
called out below.

## Why self-hosted x402 in Lua

Edge platforms (Kong, APISIX, raw OpenResty) live in front of every
upstream service inside many gateway deployments. Asking those
operators to add a sidecar facilitator just to serve a 402 is a
non-starter, and a delegated facilitator URL means another network
hop per request. So the Lua port ships the full self-hosted flow:
challenge issuance, structural verification, facilitator cosign,
broadcast, replay reservation, all inside the worker.

## Scope

Implements:

- `resty.pay_kit` umbrella: `configure`, `gate`, `usd`,
  `require_payment`, `try_payment`, `payment`, `paid`, `paid_for`
- Signer factory family (`signer.demo|bytes|json|base58|hex|file|from_env|generate`)
  and the reserved `kms` namespace
- Replay stores: `ngx.shared` (preferred), in-memory (fallback)
- MPP charge adapter wrapping `mpp.server` with proper feePayer,
  splits override, parsed-credential plumbing, and error mapping
- x402 (exact scheme, Solana) self-hosted adapter with the 11-rule
  structural verifier ported from Ruby + Rust (~280 LOC,
  divergences listed below)
- Kong plugin (`kong/plugins/pay-kit`, PRIORITY=1010, between
  basic-auth and OIDC)
- APISIX plugin (`apisix/plugins/pay-kit`, priority 2520)
- Three rockspecs: `lua-resty-pay-kit`, `kong-plugin-pay-kit`,
  `apisix-plugin-pay-kit`

Deferred (not in this PR):

- Delegated x402 mode (`config.x402.facilitator_url`): the
  dispatcher raises `errors.NOT_IMPLEMENTED` if the flag is set.
  The `lua/README.md` `forwards to the configured facilitator` line
  is aspirational and should be tightened post-merge; the actual
  code path is self-hosted only.
- KMS backends: `pay_kit.kms.{gcp,aws,vault}` raise `not implemented
  yet`. The namespace is reserved so the public surface stays stable
  for follow-up PRs.
- Token-2022 program-id strict distinction in rule 11 (see verifier
  divergences below).
- sol-native scheme.

## Verifier divergences from Rust + Ruby

The 11-rule x402 SVM-exact verifier covers the same shape but does
not match the references byte-for-byte. Two known divergences:

1. **Buyer-funded destination ATA creation is permitted.** The Lua
   verifier accepts an `AssociatedTokenAccount` create-idempotent
   instruction in the ix[3..6] allowlist even when the
   destination ATA does not yet exist. Rust's reference verifier
   refuses this; the Lua adapter follows the Ruby gem's looser
   behavior because the gateway use case typically wants new buyers
   to be able to fund their own ATA on first contact.

2. **Rule 11 (token program bind) accepts TOKEN_2022_PROGRAM in
   addition to `extra.tokenProgram`.** A strict implementation
   would reject any transferChecked whose program id does not
   match `extra.tokenProgram`. The Lua verifier permits Token-2022
   even when `extra.tokenProgram` is the legacy SPL token program
   id. This is the same gap the Ruby gem has and is tracked as a
   follow-up for both ports.

Both divergences are documented in
`lua/resty/pay_kit/schemes/x402_verify.lua`. Reviewers focused on
the security boundary should start there.

## Risk callouts

- **Operational latency.** Self-hosted x402 cosigns and broadcasts
  inside the worker process. RPC failures bubble up as a 402 with
  `invalid_proof: broadcast failed: ...`; operators should monitor
  their RPC endpoint health independently.
- **Replay store fallback.** Without
  `lua_shared_dict pay_kit_replay <size>` in nginx.conf, the
  dispatcher falls back to a per-worker in-memory LRU and logs a
  warning at boot. Multi-worker production deployments without
  the shared dict will leak the replay window across workers.
- **Demo signer guard.** `pay_kit.configure` refuses to start with
  the demo singleton when `network = solana_mainnet`. Devnet and
  localnet do not enforce this; document your KMS plan before
  promoting to staging.
- **Verifier divergences.** See above — the lua verifier is not a
  byte-for-byte port of the Rust reference. Pen-test scope should
  cover the ATA-create allowlist branch and the token-program
  matching in rule 11.
- **Blast radius.** This PR touches CI, harness, gateway plugins,
  protocol verification, and docs in one large diff. The umbrella
  was designed to land as one unit per DESIGN.md; splitting it
  would have created intermediate states that don't round-trip
  against the cross-language matrix.

## File structure (DESIGN.md Layers)

```
lua/
├── resty/pay_kit/
│   ├── init.lua, errors.lua, kms.lua, signer.lua, store.lua
│   ├── signer/{demo.lua, local.lua}
│   ├── solana/rpc.lua
│   ├── util/{base58,base64url,json,crypto,ed25519,tx_cosign}.lua
│   ├── schemes/{mpp,x402,x402_verify}.lua
│   └── internal/{config,dispatcher,fee,gate,operator,price,registry}.lua
├── kong/plugins/pay-kit/{handler,schema,bootstrap}.lua
└── apisix/plugins/pay-kit/{handler,bootstrap}.lua
```

## Test plan

- [ ] CI: `lua` workflow green (luacheck + suite + coverage + dual-protocol interop smoke)
- [ ] Local: `cd lua && luajit tests/run.lua` shows all specs passing
- [ ] Local: `cd harness && pnpm exec vitest run test/e2e.test.ts --testNamePattern "lua server"` (with surfpool + interop env) shows all lua server pairs green
- [ ] Reviewer eyeballs the verifier divergences above and decides whether to land as-is or block on byte-for-byte parity

## CI surface

- `.github/workflows/lua.yml` runs luacheck + suite + coverage gate
  + Surfpool report + dual-protocol interop smoke (TS-to-lua MPP +
  rust-x402-to-lua x402)

## Notes for reviewers

- The 4 filter-skipped scenarios in the dual-protocol smoke are
  `rust-x402-vs-ts-x402` and `rust-x402-vs-rust-x402` pairs that
  the `lua server` testNamePattern excludes. They are not gated by
  this PR.
- Coverage threshold for the lua tree is the existing 90% from the
  pre-existing `mpp/` gate. New trees (`resty/pay_kit`, `kong/`,
  `apisix/`) are included; if CI flags coverage, the gap is in
  the cosign + broadcast paths that require a live RPC to
  exercise. Happy to add stub-RPC tests in a follow-up if the gate
  blocks.


lua/mpp/init.lua
```diff
@@ -1,3 +1,22 @@
...
+    ngx_ref.log(ngx_ref.WARN, msg)
+  else
+    io.stderr:write('[pay_kit] WARN: ' .. msg .. '\n')
+  end
+end
+_pay_kit_deprecation_warn()
+
 local challenge = require('mpp.protocol.core.challenge')
 local headers = require('mpp.protocol.core.headers')
 local intents = require('mpp.protocol.intents.charge')
```
lgalabru: I'm struggling a bit understanding the file structure in Lua. could we have a clear split per protocol (x402, mpp) and within protocols, per schemes (exact, unto, etc) / intent (charge, etc) ?

harness/lua-server/server.legacy.lua
```diff
@@ -0,0 +1,370 @@
+#!/usr/bin/env luajit
```
lgalabru: No legacy / backward compatibility complexity please :) 
This is a brand new lib.

lua/pay_kit/init.lua
```diff
@@ -0,0 +1,73 @@
+--[[
```
lgalabru, line 1: regarding the high level lib structure, could we maybe have:
- lua/pay-kit
- lua/plugins/resty/pay-kit.lua -> re-export pay_kit + openresty helpers
- lua/plugins/apisix/pay-kit.lua -> re-export pay_kit + apisix helpers
- lua/plugins/kong/pay-kit.lua -> re-export pay_kit + kong helpers

would it make any sense?

EfeDurmaz16: yes I think this will be better. doing the pass

lua/pay_kit/solana/tx_cosign.lua
```diff
@@ -0,0 +1,49 @@
+--[[
+Cosign a base64-encoded Solana transaction with a 64-byte secret.
+
+The legacy `mpp.methods.solana.signer:cosign_base64` does the same
+thing but routes the actual Ed25519 sign through luasodium. This
+helper goes through `resty.pay_kit.util.ed25519` so the backend
+choice (openssl preferred, luasodium fallback) is consistent across
+the umbrella.
+
+The cosign path: parse the envelope, find the account index that
```
lgalabru, line 10: belongs to lua/pay-kit/solana?

lua/pay_kit/solana/mints.lua
```diff
@@ -42,7 +42,11 @@ function M.default_rpc_url(network)
   if network == 'devnet' then
```
lgalabru, line 42: I don't understand this `lua/mpp` dir, is this something we can move / can get rid of?

PR #160, opened 2026-06-08
feat(go): mpp/session client and server + playground-api example

EfeDurmaz16:
# feat(go): mpp/session client and server

Comprehensive `mpp/session` support for the Go SDK, rebased onto main after #165 and aligned with the rewritten `skills/pay-sdk-implementation/references/intents/mpp-session.md`. The Rust crate (`rust/crates/mpp/src`) is the wire truth throughout; the TypeScript implementation merged in #165 is the second reference. This PR ships both halves of the intent: the client surface and the full server-side session method, plus a Go port of the playground API example.

## What this adds

Wire and program layer:

- Session wire types (`SessionRequest`, the tagged `SessionAction` union, `OpenPayload`, `VoucherData`/`SignedVoucher`, `CommitPayload`/`CommitReceipt`, `TopUpPayload`, `ClosePayload`, `MeteringDirective`, `MeteringUsage`) in `go/protocols/mpp/intents/session.go`, mirroring `rust/crates/mpp/src/protocol/intents/session.rs` field for field: salt as a u64 string accepting string or number on decode (with a raw-bytes re-parse to avoid float precision loss past 2^53), the `cumulative` decode alias with `cumulativeAmount` the only serialized name, `SessionID()` keyed channelId first with tokenAccount fallback, `topUp.newDeposit` as the new total, and `DefaultSessionExpiresAt = 4_102_444_800`.
- A codama-go generated payment-channels client under `go/protocols/programs/paymentchannels/` (recipe in `skills/pay-sdk-implementation/codegen/`, reproducible via `generate-payment-channels-client-go.ts`), covering every instruction (open, topUp, settle, settleAndFinalize, distribute, finalize, requestClose, withdrawPayer, emitEvent) plus account and type decoders, with the production program id overriding the IDL placeholder and the single-byte discriminators (open = 1, topUp = 3). A frozen-hex parity suite (`go/protocols/programs/paymentchannels_parity_test/`) pins the borsh byte layouts against the Rust spine.
- The 48-byte voucher preimage (`channel_id || cumulative u64 LE || expires_at i64 LE`) has a single packer that `VoucherData.MessageBytes` delegates to, and is pinned by cross-SDK conformance vectors (`harness/vectors/session-voucher.json`, including a near-u64-max cumulative case) implemented in `go/cmd/conformance`.

Client layer, keyed to the skill's component inventory:

- Challenge parsing and mode selection: `ParseSessionChallenge` plus challenge selection with mode gating in `client/challenge_selection.go`, encoding modes empty or omitted as push-only (TS `sessionRequestModes` semantics, Rust membership checks).
- Challenge-driven open layer in `client/payment_channels.go`, ported line by line from `rust/crates/mpp/src/client/payment_channels.rs`: open derivation (mint from the challenge currency with the localnet to mainnet fallback, deposit defaulting to the cap, grace defaulting to 900 seconds, token program resolved from the currency so Token-2022 currencies open correctly, splits, random u64 salt), transaction assembly with fee payer = challenge operator and payer partial-signing only its own slot, the challenge `recentBlockhash`, the `PENDING_SERVER_SIGNATURE` placeholder (64 ones), and standard base64 with padding for the wire transaction. A per-call program id can be threaded through for non-mainnet clusters.
- Session state with the prepare/record split: `ActiveSession` signs vouchers without advancing the watermark and records them only after server acceptance, with channel-match checks, strict monotonicity, u64 overflow guards, and the Rust nonce rule.
- Metered consumer: `SessionConsumer` validates the directive session, prepares the voucher for `watermark + amount`, and commits with the directive's `deliveryId`. On a `replayed` receipt it reconciles the watermark to `min(settled, prepared)` instead of recording blindly, matching the hardened semantics proposed for the Rust client in #162; the clamp treats the server as untrusted.
- Streaming consumption in `client/http_stream.go`, porting `rust/crates/mpp/src/client/http_stream.rs`: incremental SSE decoding, metered event classification (`mpp.metering`/`metering`, `mpp.usage`/`usage`, `done`, `[DONE]`), enforcement that a usage event's `deliveryId` matches the live directive and that usage overrides only the amount, and a net/http-backed commit transport.

Server layer, mirroring the TS server that landed in #165 (`typescript/packages/mpp/src/server/Session.ts` and `server/session/*`) and `rust/crates/mpp/src/server/session.rs`:

- `go/protocols/mpp/server/session_store.go`: per-channel state store with an atomic `UpdateChannel` read-modify-write mutator (per-channel mutex; a failing mutator leaves state unchanged), pluggable like the existing replay stores.
- `session_voucher.go`: the ordered voucher verifier as a pure function with the exact reference check sequence (parse, finalized, close pending, idempotent replay on same cumulative and same re-verified signature, strictly above watermark, within deposit, `minVoucherDelta`, Ed25519 against the stored authorizedSigner, expiry), preflighted outside the store lock and re-checked inside.
- `session.go`: the off-chain `SessionServer` core (challenge request building with cap clamping, open/voucher/commit/topUp/close handlers, reservation accounting `cumulative + pendingTotal + amount <= deposit`, sequence assignment with the default `<sessionId>:<sequence>` delivery id, replay returning the cached receipt).
- `session_onchain.go`: open-transaction verification (decode, fee-payer signature binding, deposit and recipient checks) and settle/finalize composition (settleAndFinalize with the ed25519 precompile instruction immediately preceding it, distribute bundled in the same transaction), gated behind an injectable verifier seam so the no-RPC trust model matches Rust `rpc_url = None`. Server-broadcast open submission completes the fee-payer signature and broadcasts, recording the settled signature on channel state.
- `session_method.go` and `session_routes.go`: the HTTP-facing session method (HMAC-bound 402 challenges, credential verification dispatching across the five actions) and the reserve/commit metering side channel (`POST /__402/session/deliveries`, `POST /__402/session/commit`), flagged as the TypeScript-server extension it is.
- `session_lifecycle.go`: the idle-close watchdog (single-shot timer per channel, reset on voucher/commit/topUp, close-and-settle on fire), mirroring the TS-only lifecycle extension.
- `session_stream.go`: a server-side metered SSE stream writer emitting the directive/usage/done event shapes the client consumer parses.

Example:

- `go/examples/playground-api`: a Go port of `typescript/examples/playground-api` with endpoint parity for the playground web app. Charges (stock quote, marketplace splits, fortune payment link, faucet), sessions (`/sessions/stream` pay-per-chunk SSE and `/sessions/compute` pay-per-call with real opens, side-channel metering, and idle-close on-chain settlement), the x402 exact demo routes with the embedded facilitator, and the config catalog endpoint. Setting `PAYKIT_PLAYGROUND_API_URL` points the playground's `pnpm dev` at this server instead of launching the TypeScript one.

Encoding boundaries follow the repo rule: canonical JSON (RFC 8785) to base64url without padding for the credential envelope and request field, standard base64 with padding for transactions, base58 for signatures, pubkeys, and blockhashes.

## Notes for review

- The replayed-receipt reconcile in `SessionConsumer` intentionally tracks the #162 Rust semantics rather than the current TS client, which still records the prepared voucher unconditionally. If #162 lands first this is exact parity; if not, this is the same divergence flagged there.
- The generated client under `go/protocols/programs/paymentchannels/` is codegen output and is carved out of the coverage gate alongside the examples; the curated layer in `go/paycore/paymentchannels` is fully covered.
- The README scheme matrix now marks `mpp/session` shipped on both client and server, with the scope notes inline.

## Testing

- `go test ./...`: all packages pass, including the session intent serde suite, client session/consumer/stream suites, the server method, store, voucher, on-chain, concurrency, and lifecycle suites, and the playground example smoke tests.
- Surfpool-gated e2e: `session_e2e_test.go` and `playground_e2e_test.go` drive the full lifecycle (real open completed and broadcast by the server, metered SSE with per-chunk vouchers, side-channel reserve/commit, on-chain settle at idle close) against the hosted Solana Payment Sandbox; they skip explicitly when the sandbox is unreachable or under `-short`, never silently pass.
- Frozen-hex borsh parity tests pin the generated instruction encodings against the Rust spine; `harness/vectors/session-voucher.json` pins the voucher preimage cross-SDK.
- CI: the Go workflow gains a playground job that builds and boots `go/examples/playground-api` and runs the payment-link Playwright suite against it; the harness matrix job honors `PAY_KIT_HARNESS_PROTOCOL` in the Go client adapter.
- `gofmt` clean; lint configured to skip the generated client only.


go/examples/playground-api/charges.go
```diff
@@ -0,0 +1,410 @@
...
+		}))(logged(func(w http.ResponseWriter, r *http.Request) {
+			quote, err := yahooQuote(r.Context(), r.PathValue("symbol"))
+			if err != nil {
+				writeJSONError(w, http.StatusInternalServerError, "Failed to fetch quote")
+				return
+			}
+			writeJSON(w, http.StatusOK, quote)
+		})))
+
+	mux.Handle("GET /api/v1/stocks/search",
```
lgalabru, line 158: Let's implement the exact same endpoints we have in `typescript/examples/playground-api`

go/examples/playground-api/constants.go
```diff
@@ -0,0 +1,25 @@
...
+const (
+	// usdcMint is the mainnet USDC mint. Surfpool clones it from the
+	// datasource network, so the same mint works on the hosted localnet.
+	usdcMint = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v"
+
+	// usdcDecimals is the USDC token decimal count.
+	usdcDecimals = 6
+
+	// tokenProgram is the SPL Token program id.
+	tokenProgram = "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
```
lgalabru: This constants are very likely already available in paykit

go/examples/playground-api/main.go
```diff
@@ -0,0 +1,407 @@
...
+		},
+		MPP: paykit.MPPConfig{
+			Realm:                  "PayKit Playground",
+			ChallengeBindingSecret: []byte(a.secretKey),
+		},
+	})
+}
+
+// paykitNetwork maps the playground NETWORK tag onto the paykit enum.
+func paykitNetwork(tag string) (paykit.Network, error) {
```
lgalabru: There should already be a helper in paykit

go/examples/playground-api/main_test.go
```diff
@@ -0,0 +1,432 @@
+package main
```
lgalabru, line 1: Let's clean up this example as much as possible - if we find ourselves writing code that could be helpful for others, we consolidate in pay-kit

go/protocols/mpp/client/challenge_selection.go
```diff
@@ -0,0 +1,163 @@
...
+type SelectSessionChallengeOptions struct {
+	// Network is the Solana network the client wants to pay on, e.g.
+	// "mainnet", "mainnet-beta", "devnet", or "localnet". "mainnet" and
+	// "mainnet-beta" are treated as the same network.
+	Network string
+
+	// Currencies are the currency symbols or mint addresses the client wants
+	// to pay with. A challenge matches when its currency resolves to the same
+	// mint as any entry.
+	Currencies []string
```
lgalabru: should we be working with enums here?

go/protocols/mpp/client/http_stream.go
```diff
@@ -0,0 +1,481 @@
+// HTTP streaming helpers for metered sessions.
+//
+// LLM APIs commonly stream responses over Server-Sent Events (SSE) or chunked
+// HTTP. This file keeps the parser transport-neutral (SseDecoder works on raw
+// chunks from any reader), then layers a net/http-friendly stream and commit
+// transport on top for applications that want batteries included.
+//
+// Behavior mirrors rust/crates/mpp/src/client/http_stream.rs; the TypeScript
+// counterpart is typescript/packages/mpp/src/client/HttpStream.ts.
```
lgalabru: Can we remove these comments?

go/protocols/mpp/client/http_stream.go
```diff
@@ -0,0 +1,481 @@
...
+	return &MeteredSseStream{
+		consumer: consumer,
+		body:     body,
+		buf:      make([]byte, 4096),
+	}
+}
+
+// Next returns the next application message, or nil once the stream is done.
+//
+// Mirrors rust ReqwestMeteredSseStream::next.
```
lgalabru: Remove all these Mirrors comments.

go/protocols/mpp/client/session.go
```diff
@@ -0,0 +1,389 @@
...
+// goroutine or guard it with a mutex.
+//
+// Mirrors rust ActiveSession in rust/crates/mpp/src/client/session.rs.
+type ActiveSession struct {
+	channelID  solana.PublicKey
+	cumulative uint64
+	nonce      uint64
+	expiresAt  int64
+	signer     VoucherSigner
+}
```
lgalabru, line 67: Generally speaking, can we do a pass and get every single field of every single structure being commented?

lgalabru (approved):

PR #161, opened 2026-06-08
feat(python): MPP sessions (client + server + playground)

EfeDurmaz16:
## What this adds

The MPP **session** intent for the Python SDK, now both halves plus the example app, bringing Python to parity with the Go port (#160). Rebased onto main (which carries #160), so the diff is the Python session surface only.

- **Wire types** `protocols/mpp/intents/session.py` — `SessionRequest`, the `SessionAction` tagged union (open / voucher / commit / topUp / close), signed vouchers, metering types.
- **Client** `protocols/mpp/client/session.py` + `session_consumer.py` — `ActiveSession` (voucher signing, monotonic watermark, action builders) and `SessionConsumer` (metered ack/commit).
- **Server** `protocols/mpp/server/session_*.py` — the full server method ported from `go/protocols/mpp/server` against the Rust spine (`rust/crates/mpp/src/server/session.rs`) as wire truth: `session_store` (channel store with per-channel-locked read-modify-write and Go-faithful nil-slice serialization), `session_voucher` (offline voucher verifier), `session_lifecycle`, `session_onchain` (open/top-up verification with an optional RPC liveness seam, submit-open server-broadcast, and settle-at-close), `session_method` (`new_session` + open/voucher/commit/topUp/close handlers, re-drivable close, pull-mode guard, atomic settle-in-progress guard), `session_stream` (metered SSE), `session_routes` + `session` (HTTP routes with strict typed decode, public `SessionServer`).
- **Generated client** `protocols/programs/paymentchannels/` — codama-py output (Go uses `@codama/renderers-go`); regenerate with `just payment-channels-generate-py`.
- **Playground** `examples/playground_api/` — FastAPI port of `go/examples/playground-api`: charges (stocks/marketplace/weather with fee splits), the metered session stream/compute/receipt, x402, faucet, docs, health/config catalog.

## Reviewing it

Suggested read order (the generated client and the playground example can be skimmed):

1. `protocols/mpp/intents/session.py` (wire types)
2. `protocols/mpp/client/session.py`, `session_consumer.py`
3. `protocols/mpp/server/session_store.py`, `session_voucher.py`, `session_method.py` (the fund-safety surface)
4. `protocols/mpp/server/session_routes.py` (HTTP + strict decode)
5. `protocols/mpp/server/session_onchain.py` (server-broadcast open + settle-at-close + polling confirmation)
6. `examples/playground_api/` (runnable example)

## On-chain broadcast and settlement

The Python SDK now ships the server-side on-chain paths the Go/TS servers have:

- **Server-broadcast open** (`openTxSubmitter=server`): the server completes the fee-payer signature on a client-built open transaction, broadcasts it, and waits for confirmation before persisting channel state. Mirrors the TS `submitOpenTx` helper. A replayed open does NOT re-broadcast: a store existence pre-check short-circuits to the idempotent `process_open` path before the broadcast, since re-sending an already-landed open would break idempotent-replay semantics.
- **Settle at close** (`settle_and_finalize_channel`): a close with a signer and RPC configured builds the `settle_and_finalize` instruction (preceded by the Ed25519 precompile when a voucher was recorded) plus a `distribute` instruction, signs, broadcasts, and confirms, then records the settlement signature and finalizes. The close is re-drivable when no settlement signature was recorded so a transient settlement failure cannot strand the channel.
- **Polling confirmation** (`confirm_transaction_signature`): a freshly broadcast signature commonly returns `None` from `getSignatureStatuses` for hundreds of ms to seconds. The helper now polls (mirroring the TS `waitForSignatureConfirmation`) until the signature reaches at least `confirmed`, fails on-chain, or times out (default 30s). This replaces the old single-shot check, which raised spuriously on just-broadcast transactions. The already-landed open/top-up verify path (`verify_open_tx`) keeps the single-shot liveness check, since the signature there was supplied by the client and is expected to be known.
- **Atomic settle-in-progress guard** (`settling`): the settle path claims a `settling` flag under the per-channel store lock before broadcasting, so a concurrent close retry or idle-watchdog fire cannot both pass the finalize check and broadcast duplicate settle transactions. The guard is released by the finalize mutator (which sets `finalized`), or by a failed broadcast path so a retry can claim again.

The RPC client remains optional: with no RPC the open/top-up signatures and deposit amounts are trusted as provided (offline core), and close is a pure state-flip with `settledSignature=null`. Mirrors how the Go port treats the RPC client as optional; the playground uses the client submitter with `rpc=None` accordingly.

## Verification

`ruff check`, `ruff format`, `pyright` (0 errors), `pytest --cov=pay_kit --cov-fail-under=90` (cov 93.6%). The server modules mirror Go's test behaviors test-first. Manual check: the playground responds with real `402 Payment` challenges on the session stream, charge, and x402 routes; the app boots with all routes registered. The server-broadcast open path reaches the RPC (fails on blockhash-not-found without a live validator, not the old `verify_open_tx` hard-reject). Pull-mode opens with no transaction skip the server-broadcast path and trust the channel id/deposit, mirroring the TS `else` open branch. The 48-byte voucher preimage and single-byte discriminators (open=1, topUp=3) match the Rust/Go frozen vectors.

---

### Reviewer note (sakana-reviewer, fugu-ultra)

Independent verification against branch tip `81dde1d3`. All three fixes (polling confirm, replay-safe open broadcast, atomic settle guard) hold against the current tree.

- **B1 sound:** `confirm_transaction_signature` (`session_onchain.py:354-417`) now polls `getSignatureStatuses` on a 30s/1s deadline, exits on `err` (`transaction-failed`), treats `confirmationStatus` of `None`/`confirmed`/`finalized` as success, and cleanly distinguishes `transaction-not-found` (no status ever seen) from `transaction-not-confirmed` (saw a status, timed out). Wired into the settle broadcast (`:493`), so just-broadcast settles no longer reject spuriously.
- **S1 preserves idempotency:** the pre-broadcast `get_channel` check (`session_method.py:551-580`) only short-circuits the re-broadcast; `process_open` (`session.py:341-353`) remains the atomic check-and-insert source of truth, returning existing state unchanged with the voucher watermark intact. Sequential replay is broadcast-safe; truly concurrent first-opens stay bounded by the atomic mutator.
- **S2 race-safe:** `_settle_channel` (`session_method.py:768-833`) claims `settling` under the per-channel store lock, losers return `None`/the finalized signature without broadcasting, the guard is released on broadcast/confirm failure, and `finalize` clears it. Confirm-before-finalize + on-chain idempotency contain the restart window (`settling` is intentionally transient/unserialized).
- **No fund-safety regressions:** no path double-credits, leaks, or bypasses the deposit; deposit is never re-credited, settle is single-broadcast, and finalize keeps the winner's signature. **Approve.**


python/examples/playground_api/app.py
```diff
@@ -0,0 +1,400 @@
...
+
+def build_endpoint_list() -> list[dict[str, Any]]:
+    """Build the ``/api/v1/config`` endpoint catalog.
+
+    The subscription entry is omitted because the Python SDK has no subscription
+    server method yet (see README.md); the stocks-search / stocks-history /
+    weather / fortune / x402 routes stay live server-side but are not advertised
+    in the nav.
+    """
+    return [
```
lgalabru: Not a huge fan of this approach. can we have endpoints declared in-line instead of this json?

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,400 @@
...
+    """
+
+    @app.get("/api/v1/health")
+    async def health() -> JSONResponse:
+        body: dict[str, Any] = {
+            "ok": True,
+            "feePayer": state.fee_payer_pubkey,
+            "recipient": state.recipient,
+            "network": state.network,
+            "rpcUrl": state.rpc_url,
```
lgalabru: I don't think the RPC url should be exposed.

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,400 @@
...
+            result = await rpc_call(state.rpc_url, "getBalance", [state.fee_payer_pubkey])
+            value = result.get("value") if isinstance(result, dict) else result
+            if value is not None:
+                body["feePayerBalance"] = float(value) / 1e9
+        except Exception:
+            # Best-effort balance: omitted when the RPC is unreachable.
+            pass
+        return JSONResponse(body)
+
+    @app.get("/api/v1/config")
```
lgalabru: Is this endpoint actually needed by playground?

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,400 @@
...
+        yahoo = getattr(app.state, "playground_yahoo", None)
+        if yahoo is not None:
+            await yahoo.aclose()
+
+    return app
+
+
+# Module-level app for ``uvicorn examples.playground_api.app:app``. Boots from
+# the environment; the entrypoint in ``main.py`` does the same with funding.
+app = create_app()
```
lgalabru: I'm surprised by the amount of code required to setup a simple python server exposing 5 endpoints.
Any reusable helpers that can / should be pushed to PayKit?

python/examples/playground_api/README.md
```diff
@@ -0,0 +1,118 @@
...
+| GET | `/api/v1/marketplace/buy/:productId?referrer=` | charge with splits |
+| GET | `/api/v1/fortune` | charge 0.01 USDC, HTML payment link |
+| GET | `/api/v1/premium/feed` | 501 stub (see below) |
+| GET | `/sessions/stream` | session, cap 1.00 USDC, 0.0001 USDC/chunk |
+| POST | `/sessions/stream` | session voucher commits |
+| POST | `/sessions/compute` | session, cap 0.50 USDC, 0.005 USDC/call |
+| POST | `/__402/session/deliveries` | session side channel |
+| POST | `/__402/session/commit` | session side channel |
+| GET | `/sessions/receipt/:channelId` | free settle-status poll |
+| GET | `/facilitator/supported` | free |
```
lgalabru: Why so many endpoints?

python/examples/playground_api/sessions.py
```diff
@@ -0,0 +1,312 @@
...
+            {
+                "channelId": state_channel.channel_id,
+                "cumulative": str(state_channel.cumulative),
+                "deposit": str(state_channel.deposit),
+                "finalized": state_channel.finalized,
+                "settledSignature": state_channel.settled_signature,
+            }
+        )
+
+    return shutdown
```
lgalabru: Same feedback - too much code

python/examples/playground_api/x402.py
```diff
@@ -0,0 +1,147 @@
...
+        amount=usd("0.001"),
+        description="A random fun fact",
+        accept=(Protocol.X402,),
+        default_pay_to=x402_config.effective_recipient(),
+    )
+    require_fact = Depends(RequirePayment(fact_gate, config=x402_config))
+
+    @app.get("/x402/fact")
+    async def x402_fact(payment: Payment = require_fact) -> JSONResponse:
+        return JSONResponse({"fact": random.choice(FACTS), "source": "x402"})
```
lgalabru: Same feedback - Feels like a lot of code here belongs to PayKit

python/examples/playground_api/yahoo.py
```diff
@@ -0,0 +1,426 @@
...
+
+def _parse_float_prefix(text: str) -> float | None:
+    """Parse a float like JS parseFloat: surrounding whitespace is ignored."""
+    try:
+        result = float(text.strip())
+    except ValueError:
+        return None
+    if math.isnan(result):
+        return None
+    return result
```
lgalabru: Is there a python lib we can use instead of all this yahoo code?

python/src/pay_kit/protocols/mpp/intents/session.py
```diff
@@ -0,0 +1,928 @@
+"""Session intent request and voucher types.
+
+The session intent opens a payment channel between a client and server,
+allowing incremental payments via off-chain signed vouchers backed by the
+on-chain payment-channels program. The wire format mirrors the Rust spine in
```
lgalabru: "The wire format mirrors"
Can we get rid of these statements? If we're citing a source of truth, it's the spec

python/src/pay_kit/protocols/mpp/intents/session.py
```diff
@@ -0,0 +1,928 @@
...
+            new_deposit=data.get("newDeposit", ""),
+            signature=data.get("signature", ""),
+        )
+
+
+@dataclass
+class ClosePayload:
+    """Payload for the ``close`` action.
+
+    Mirrors rust ``ClosePayload``; ``voucher`` is omitted when ``None``.
```
lgalabru: Remove all these "Mirrors etc"

python/src/pay_kit/protocols/mpp/intents/session.py
```diff
@@ -0,0 +1,928 @@
...
+    """Result returned after a delivery commit is accepted.
+
+    Mirrors rust ``CommitReceipt``.
+    """
+
+    delivery_id: str
+    session_id: str
+    amount: str
+    cumulative: str
+    status: CommitStatus
```
lgalabru, line 958: Can we make sure we'll get high quality docs for all these classes, fields, etc.

justfile
```diff
@@ -65,8 +65,14 @@ payment-channels-generate-rs: codegen-install
     cd {{codegen_dir}} && pnpm run payment-channels:rust
     cd rust && cargo fmt -p payment-channels-client
 
-# Full refresh: pull IDL + regenerate Rust client.
-payment-channels-sync: payment-channels-pull-idl payment-channels-generate-rs
+# Render the Python client from the vendored IDL. Wipes
+# `python/src/pay_kit/protocols/programs/paymentchannels/` and rewrites
+# it in place — see {{codegen_dir}}/generate-payment-channels-client-py.ts.
+payment-channels-generate-py: codegen-install
+    cd {{codegen_dir}} && pnpm run payment-channels:python
```
lgalabru, line 72: I think we also have some codama commands in the main justfile, can we consolidate?

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,296 @@
...
+        fee_payer=fee_payer,
+        repo_root=find_repo_root(),
+        port=int(env_or("PORT", "3000")),
+    )
+
+
+# --- endpoint catalog (drives the playground web app's sidebar) -------------
+
+
+def build_endpoint_list() -> list[dict[str, Any]]:
```
lgalabru: Could we get rid of this function?

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,296 @@
...
+            value = result.get("value") if isinstance(result, dict) else result
+            if value is not None:
+                body["feePayerBalance"] = float(value) / 1e9
+        except Exception:
+            # Best-effort balance: omitted when the RPC is unreachable.
+            pass
+        return JSONResponse(body)
+
+    @app.get("/api/v1/config")
+    async def config() -> JSONResponse:
```
lgalabru: Is this function used?

python/examples/playground_api/charges.py
```diff
@@ -0,0 +1,196 @@
...
+        seller="7xKXtg2CW87d97TXJSDpbD5jBkheTqA83TZRuJosgAsU",
+        description="Premium Solana-branded hoodie",
+    ),
+    "validator-mug": _Product(
+        name="Validator Mug",
+        price=usd("1.00"),
+        seller="7xKXtg2CW87d97TXJSDpbD5jBkheTqA83TZRuJosgAsU",
+        description="Ceramic mug for node operators",
+    ),
+    "nft-sticker-pack": _Product(
```
lgalabru: What are these products?

python/README.md
```diff
@@ -263,7 +263,15 @@ Use MPP when:
 |---------------|--------|
 | `charge/pull` | ✅      |
 | `charge/push` | ✅      |
-| `session`     | —      |
+| `session`     | Client-only |
```
lgalabru: Arem't we supporting server?

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,293 @@
...
+        ``devnet`` / ``mainnet`` tags map onto the ``solana_*`` enum members.
+        """
+        normalized = self.network.strip().lower()
+        if normalized in ("localnet", "solana_localnet"):
+            return Network.SOLANA_LOCALNET
+        if normalized in ("devnet", "solana_devnet"):
+            return Network.SOLANA_DEVNET
+        if normalized in ("mainnet", "mainnet-beta", "solana_mainnet"):
+            return Network.SOLANA_MAINNET
+        raise ValueError(f"unknown NETWORK tag: {self.network!r}")
```
lgalabru: don't we have a helper in paykit / paycore for this?

python/examples/playground_api/app.py
lgalabru: UP

python/examples/playground_api/app.py
```diff
@@ -0,0 +1,293 @@
...
+    fee_payer = LocalSigner.from_base58(raw_fee_payer) if raw_fee_payer else LocalSigner.from_keypair(Keypair())
+    return AppState(
+        network=env_or("NETWORK", "localnet"),
+        # Default to the hosted Solana Payment Sandbox so the playground works
+        # zero-config: it has the payment-channels program preloaded and
+        # supports the surfnet cheatcodes the faucet uses. Override RPC_URL to
+        # point at a local surfpool when you need offline iteration.
+        rpc_url=env_or("RPC_URL", "https://402.surfnet.dev:8899"),
+        recipient=env_or("RECIPIENT", fee_payer.pubkey()),
+        secret_key=os.getenv("MPP_SECRET_KEY") or secrets.token_hex(32),
```
lgalabru: mpp_challenge_binding_secret

python/examples/playground_api/discovery.py
```diff
@@ -0,0 +1,167 @@
...
+from decimal import Decimal
+from typing import Any
+
+from pay_kit._paycore.protocol import Protocol
+from pay_kit.config import Config
+from pay_kit.gate import Gate
+from pay_kit.price import Price
+
+#: Solana base-unit precision for the supported stablecoins (USDC/USDT/PYUSD).
+USDC_DECIMALS = 6
```
lgalabru: there should be a paykit helper for stable coin mints config, no?

python/examples/playground_api/discovery.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def base_units(price: Price, decimals: int = USDC_DECIMALS) -> str:
+    """The integer base-unit string for a price (e.g. ``usd("0.01")`` -> ``"10000"``)."""
+    return str(int(price.amount.scaleb(decimals).to_integral_value()))
```
lgalabru: belongs to paykit / pay core

python/examples/playground_api/sessions.py
```diff
@@ -0,0 +1,188 @@
+# examples/playground_api/sessions.py
```
lgalabru, line 1: Could we move some of this code to paykit, to make sessions management easier with python? there's a lot of code that we have in this playground python that we don't in typescript.

python/src/pay_kit/protocols/mpp/server/session_routes.py
```diff
@@ -0,0 +1,215 @@
...
+            # string field, so a non-string JSON value is rejected up front.
+            delivery_id = _string_field(body, "deliveryId")
+        except (ValueError, json.JSONDecodeError):
+            return _error(400, "invalid request body")
+        if not delivery_id:
+            return _error(400, "deliveryId required")
+        voucher_raw = body.get("voucher")
+        if voucher_raw is None:
+            return _error(400, "voucher required")
+        voucher = SignedVoucher.from_dict(voucher_raw)
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 400ef1ad: a non-dict voucher now returns 400.

python/src/pay_kit/protocols/mpp/server/session_routes.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            receipt = await core.process_commit(CommitPayload(delivery_id=delivery_id, voucher=voucher))
+        except ValueError as exc:
+            return _error(400, str(exc))
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 400ef1ad: from_dict wrapped to return 400.

python/src/pay_kit/protocols/mpp/server/session_store.py
```diff
@@ -0,0 +1,364 @@
...
+            next_state = mutator(current_snapshot)
+
+            async with self._mu:
+                self._data[channel_id] = next_state.clone()
+            return next_state
+
+    async def delete_channel(self, channel_id: str) -> None:
+        async with self._mu:
+            self._data.pop(channel_id, None)
+
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 400ef1ad: delete_channel now takes the per-channel lock.

python/src/pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -0,0 +1,731 @@
...
+        failure is swallowed (logged) so it cannot crash the timer, and the
+        channel stays re-drivable with ``settledSignature`` unset.
+        """
+        if self._signer is None or self._rpc is None:
+            return None
+        try:
+            await self._handle_close(ClosePayload(channel_id=channel_id, voucher=None))
+        except Exception:
+            logging.getLogger(__name__).warning("idle-close settle failed for channel %s", channel_id, exc_info=True)
+        return None
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 400ef1ad: close state-flip always runs; settle stays gated.

python/src/pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -0,0 +1,731 @@
...
+
+def _parse_session_u64(value: str, name: str) -> int:
+    """Parse a non-negative decimal string into a u64, naming the field on
+    error."""
+    if not (value.isascii() and value.isdigit()):
+        raise ValueError(f"{name} is not an unsigned integer string: {value}")
+    parsed = int(value, 10)
+    if parsed > _U64_MAX:
+        raise ValueError(f"{name} is not an unsigned integer string: {value}")
+    return parsed
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 400ef1ad: added isinstance(str) guard before parse.

python/src/pay_kit/protocols/mpp/client/__init__.py
```diff
@@ -1,9 +1,33 @@
...
 
+from pay_kit.protocols.mpp.client.session import (
+    ActiveSession,
+    parse_session_challenge,
+    serialize_session_credential,
+)
+from pay_kit.protocols.mpp.client.session_consumer import (
+    MeteredDelivery,
+    SessionConsumer,
+)
```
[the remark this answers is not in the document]
EfeDurmaz16: Correction: it was not re-exported at HEAD. Fixed in ce2a90c9: CommitTransport is now imported from session_consumer in the client package.

python/src/pay_kit/protocols/mpp/client/__init__.py
```diff
@@ -1,9 +1,33 @@
...
 from pay_kit.protocols.mpp.client.transport import PaymentTransport
 
 __all__ = [
     "PaymentTransport",
+    "ActiveSession",
+    "SessionConsumer",
+    "MeteredDelivery",
+    "serialize_session_credential",
+    "parse_session_challenge",
 ]
```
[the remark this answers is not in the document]
EfeDurmaz16: Correction: it was not in __all__ at HEAD. Fixed in ce2a90c9: CommitTransport added to the client package __all__.

python/src/pay_kit/protocols/mpp/intents/session.py
```diff
@@ -0,0 +1,876 @@
...
+        return d
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> VoucherData:
+        if "cumulativeAmount" in data:
+            cumulative = data["cumulativeAmount"]
+        elif "cumulative" in data:
+            cumulative = data["cumulative"]
+        else:
+            cumulative = ""
```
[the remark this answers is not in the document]
EfeDurmaz16: No int(raw,10) call: base units parse via str-coercing _parse_base_units, so a JSON number is accepted and a bad value raises ValueError, never TypeError. from_dict already coerces cumulative to str.

python/src/pay_kit/protocols/mpp/intents/session.py
```diff
@@ -0,0 +1,876 @@
...
+        return cls(
+            delivery_id=data.get("deliveryId", ""),
+            session_id=data.get("sessionId", ""),
+            amount=data.get("amount", ""),
+            currency=data.get("currency", ""),
+            sequence=int(data.get("sequence", 0)),
+            expires_at=int(data.get("expiresAt", 0)),
+            commit_url=data.get("commitUrl"),
+            proof=data.get("proof"),
+        )
```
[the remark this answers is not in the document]
EfeDurmaz16: amount_base_units delegates to _parse_base_units (str-coerced), so there is no TypeError path; a malformed amount raises the expected ValueError.

python/src/pay_kit/protocols/mpp/intents/session.py
```diff
@@ -0,0 +1,881 @@
...
+            return int(self.amount, 10)
+        except ValueError as exc:
+            raise ValueError(f"invalid metering usage amount: {self.amount}") from exc
+
+    def to_dict(self) -> dict[str, Any]:
+        return {"deliveryId": self.delivery_id, "amount": self.amount}
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> MeteringUsage:
+        return cls(delivery_id=data.get("deliveryId", ""), amount=data.get("amount", ""))
```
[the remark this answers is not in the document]
EfeDurmaz16: MeteringUsage/CommitReceipt also parse via _parse_base_units (str-coerced); they raise ValueError, not TypeError.

python/src/pay_kit/protocols/mpp/_paymentchannels.py
```diff
@@ -0,0 +1,266 @@
...
+        AccountMeta(associated_token_program, False, False),
+        AccountMeta(event_authority, False, False),
+        AccountMeta(PROGRAM_ID, False, False),
+    ]
+
+    data = bytearray()
+    data.append(_OPEN_DISCRIMINATOR)
+    data += struct.pack("<Q", params.salt)
+    data += struct.pack("<Q", params.deposit)
+    data += struct.pack("<I", params.grace_period)
```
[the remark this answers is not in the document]
EfeDurmaz16: Addressed in e846b907: the manual struct.pack is gone, and build_open_instruction now range-guards salt/deposit (u64), grace_period (u32), and bps (u16) up front so an out-of-range value raises a clear ValueError instead of a low-level encoder error.

python/examples/playground_api/sessions.py
EfeDurmaz16: Done in e5660a5f: the session gate ships as pay_kit RequireSession + Session.handle.

python/examples/playground_api/sessions.py
EfeDurmaz16: Trimmed in e5660a5f: gate moved to pay_kit RequireSession (189->160 lines).

python/examples/playground_api/app.py
EfeDurmaz16: Done in e5660a5f: session gate + stablecoin_decimals now live in pay_kit.

python/examples/playground_api/discovery.py
EfeDurmaz16: Fixed in dc6e5664: discovery uses stablecoin_decimals.

python/examples/playground_api/discovery.py
EfeDurmaz16: Kept example-local: TS ships no discovery/OpenAPI builder either, so moving it would break parity. Flagged for a cross-SDK helper follow-up; corrected the docstring that wrongly claimed a TS openapi.ts.

python/examples/playground_api/app.py
EfeDurmaz16: Done: routes are declared inline now.

python/examples/playground_api/app.py
EfeDurmaz16: Fixed: the RPC url is no longer exposed.

python/examples/playground_api/app.py
EfeDurmaz16: Removed: that endpoint was dropped in the rewrite.

python/examples/playground_api/README.md
EfeDurmaz16: Each endpoint exercises a distinct gate (charge/x402/session).

python/examples/playground_api/x402.py
EfeDurmaz16: Removed: x402.py is gone; it uses the shared gate now.

python/examples/playground_api/yahoo.py
EfeDurmaz16: Removed: yahoo.py deleted; the quote route is a stub.

python/examples/playground_api/app.py
EfeDurmaz16: Removed: that function is gone in the rewrite.

python/examples/playground_api/app.py
EfeDurmaz16: Removed: the unused function was deleted.

python/examples/playground_api/charges.py
EfeDurmaz16: Removed: charges.py is gone; gates are named inline.

python/README.md
EfeDurmaz16: Yes: the session now ships server-side (new_session + RequireSession).

python/examples/playground_api/app.py
EfeDurmaz16: Yes: it now uses pay_kit RequirePayment + install().

python/examples/playground_api/app.py
EfeDurmaz16: It's pay_kit config().mpp.challenge_binding_secret.

python/src/pay_kit/protocols/mpp/intents/session.py
EfeDurmaz16: Done: the docstring now cites the MPP spec.

python/src/pay_kit/protocols/mpp/intents/session.py
EfeDurmaz16: Done: all 'Mirrors rust' citations were removed.

python/src/pay_kit/protocols/mpp/intents/session.py
EfeDurmaz16: Done: full per-class and per-field docstrings are in place.

justfile
EfeDurmaz16: Already consolidated in the root justfile.

python/src/pay_kit/protocols/mpp/server/session_routes.py
```diff
@@ -0,0 +1,216 @@
...
+            return _error(400, "voucher required")
+        # A non-dict voucher makes SignedVoucher.from_dict raise AttributeError
+        # (not ValueError), and a non-numeric expiresAt/nonce raises ValueError;
+        # both must surface as 400, matching Go's strict json.Decode.
+        if not isinstance(voucher_raw, dict):
+            return _error(400, "invalid request body")
+        try:
+            voucher = SignedVoucher.from_dict(voucher_raw)
+        except (ValueError, TypeError):
+            return _error(400, "invalid request body")
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 10af4467: added AttributeError to the catch, so a string/array voucher.data returns 400.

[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 10af4467: AttributeError now caught; returns 400 like Go's strict decode.

python/src/pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -0,0 +1,797 @@
...
+        def finalize(current: ChannelState | None) -> ChannelState:
+            if current is None:
+                raise ValueError(f"channel {channel_id} disappeared during settle")
+            nxt = current.clone()
+            nxt.finalized = True
+            nxt.settled_signature = signature
+            return nxt
+
+        await self._core.store().update_channel(channel_id, finalize)
+        return signature
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 10af4467: finalize mutator returns early if already finalized, so a concurrent re-drive cannot overwrite the signature.

python/src/pay_kit/protocols/mpp/server/session_onchain.py
```diff
@@ -0,0 +1,463 @@
...
+        mint=Pubkey.from_string(mint_address),
+        recipients=[Distribution(recipient=Pubkey.from_string(s.recipient), bps=s.bps) for s in config.splits],
+        token_program=Pubkey.from_string(default_token_program_for_currency(config.currency, config.network)),
+        program_id=program_id,
+    )
+
+    blockhash = Hash.from_string((await rpc.get_latest_blockhash()).value.blockhash)
+    tx = Transaction.new_signed_with_payer([*settle, distribute], merchant_pubkey, [merchant], blockhash)
+    sent = await rpc.send_raw_transaction(bytes(tx))
+    return str(sent.value)
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in 77fb689a: settle_and_finalize_channel now confirms the tx (confirm_transaction_signature) before returning, mirroring the open path, so a dropped/failed settle raises and the channel is not finalized with an unconfirmed signature.

python/src/pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -0,0 +1,804 @@
...
+        elif mode == "push" and not has_channel_id:
+            raise PaymentError("open payload missing transaction or channelId", code="invalid-payload")
+        elif mode == "push" and self._rpc is not None:
+            await confirm_transaction_signature(self._rpc, payload.signature, "open")
+
+        try:
+            state = await self._core.process_open(payload)
+        except ValueError as exc:
+            raise PaymentError(str(exc), code="invalid-payload") from exc
+        self._touch(state.channel_id)
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in dc6e5664: payer or tx required.

python/src/pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -0,0 +1,828 @@
...
+            except PaymentError:
+                raise
+            except Exception as exc:
+                raise PaymentError(f"server-broadcast open failed: {exc}", code="invalid-payload") from exc
+            try:
+                state = await self._core.process_open(payload)
+            except ValueError as exc:
+                raise PaymentError(str(exc), code="invalid-payload") from exc
+            self._touch(state.channel_id)
+            return payload.signature
```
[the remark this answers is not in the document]
EfeDurmaz16: Gated server-broadcast on attached transaction, pull opens now skip it.

lgalabru (approved):

PR #162, opened 2026-06-08
fix(rust): do not advance session watermark on a replayed commit

EfeDurmaz16:
# fix(rust): reconcile session watermark on a replayed commit

Suggested PR title (replaces the stale "do not advance session watermark on a replayed commit"): the final semantics do advance the watermark on replay, by reconciling it to the server-settled cumulative, clamped and never regressing.

## Problem

`SessionConsumer::commit_directive` recorded the freshly prepared voucher unconditionally after every commit, including commits the server answered with `status: replayed`. Per the replay rule in `skills/pay-sdk-implementation/references/intents/mpp-session.md` (lines 221, 294, 355-357), a commit is `replayed` only when the retried delivery carries the same cumulative AND the same signature as the cached committed delivery, with the signature re-verified, and the server returns the cached receipt without advancing its watermark. Both reference servers implement exactly that:

- Rust: `rust/crates/mpp/src/server/session.rs:495` (replay condition), `:555` (replay check inside the atomic store closure), `:782` (`CommitStatus::Replayed` with the cached receipt)
- TypeScript: `typescript/packages/mpp/src/server/Session.ts:893-901` (idempotent replay window inside `updateChannel`: same `cumulative` and same `voucherSignature`, signature re-verified via `assertVoucherSignature`, cached outcome returned and `return current` leaves channel state untouched)

The client, however, treated a replayed receipt like a fresh acceptance and recorded the voucher it had just prepared. Against a non-conforming or malicious server that reports `replayed` with an arbitrary cumulative, this over-advanced the local watermark, so the next signed voucher could over-authorize up to the on-chain deposit. Skipping the record entirely is also wrong: in the lost-response case (the original commit succeeded but the receipt never arrived) the local watermark is behind the server, and the next delivery would sign a non-monotonic cumulative the server must reject.

## Fix

- `ActiveSession::reconcile_settled(settled)`: advances the local watermark to a server-settled cumulative when it is ahead, never regresses, and bumps the request nonce on advance so the next prepared voucher does not reuse the already-settled nonce (mirrors the `record_voucher` nonce accounting).
- `SessionConsumer::commit_directive`: on `CommitStatus::Replayed`, parse the receipt cumulative and reconcile to `min(settled, prepared)` instead of recording the prepared voucher; fresh commits record the voucher as before. The clamp treats the server as untrusted: an honest lost-response replay settles at or below the voucher just signed (single-threaded session), so a server-reported cumulative above it cannot push the watermark past what the client actually authorized.
- `ActiveSession::record_voucher`: rejects vouchers for a different channel. This is parity catch-up with the TypeScript client, which already throws on a channel mismatch in `typescript/packages/mpp/src/client/Session.ts:466`.

Against a conforming server the change is unobservable: replay requires the identical cumulative and signature, Ed25519 signing is deterministic, so settled equals prepared and `reconcile_settled(min(settled, prepared))` advances exactly as `record_voucher` did. The fix is defensive hardening on the client, not a wire change.

## Note for review: TypeScript client has the pre-fix behavior

`typescript/packages/mpp/src/client/SessionConsumer.ts:76` (`commitDirective`) still calls `this.#session.recordVoucher(voucher)` unconditionally, with no branch on `receipt.status === 'replayed'`. After this PR the Rust and TypeScript reference clients diverge on the replayed branch. The divergence is latent against conforming servers (per above), but ports mirror both references, so a matching TypeScript follow-up is needed. Happy to open it once the semantics here are agreed.

## Tests

Unit tests cover: lost-response reconcile to the settled cumulative, no regression from a stale replayed receipt, clamp against an inflated server-reported cumulative, foreign-channel voucher rejection, and nonce freshness after a replay reconcile.

- `cargo fmt --check`: clean
- `cargo clippy -p solana-mpp --all-targets`: no warnings in the touched files (the remaining warnings pre-exist on main in files this PR does not modify)
- `cargo test -p solana-mpp`: 641 lib + 9 integration tests pass, 0 failed

Coverage is unit-level (transport fakes). The harness has no session scenario yet; when the session harness cell lands, a duplicate-commit scenario should exercise this replayed-client path cross-SDK. Tracked as a follow-up.

## Read order
1. `rust/crates/mpp/src/client/session.rs`, `reconcile_settled` (strict greater-than, nonce bump) and the channel guard in `record_voucher`
2. `rust/crates/mpp/src/client/session_consumer.rs`, the `Replayed` branch with the `settled.min(prepared)` clamp
3. session.rs tests: advance-never-regress, no-nonce-reuse, foreign-channel reject
4. session_consumer.rs tests: `ReplayTransport` + the three replay scenarios

## Generated vs handwritten
The load-bearing change is ~25-30 lines (`reconcile_settled` body, the `Replayed` branch, the channel guard); the other ~150 added lines are test scaffolding and rationale comments. Small diff, small blast radius.

## Reviewer note
Grade this as hardening, not a live-fund bug: both canonical servers emit `Replayed` only for a byte-identical voucher, and Ed25519 re-signing is deterministic, so on every reachable honest path the old code already landed on the same watermark. The fix's value is against non-canonical or lying servers (the clamp blocks over-authorization, strict greater-than blocks regression) and interop with looser replay-idempotency servers. Status: the bug still exists on main and nothing superseded it, but the branch conflicts with the `crates/mpp` to `crates/kit/src/mpp` restructure and needs a rebase that re-targets the moved files before merge. TS has the same gap in `SessionConsumer.commitDirective`; this is the template for that follow-up.

## Fable 5 self-review
The logic is correct at every boundary I could attack (equality no-op, reject at less-or-equal, clamp against inflated settled) and the adversarial tests are genuinely good. What a stranger would fairly flag: the PR sat unrebased through a crate restructure and now conflicts; the body originally described the first commit's semantics rather than the final reconcile-with-clamp; and the replay branch has the settled voucher's nonce in hand but bumps by a bare `+1` instead of reconciling it. All 641 solana-mpp lib tests pass on this head.


lgalabru:
Merge conflicts

PR #171, opened 2026-06-15
feat(rust): unified dual-protocol axum payment gate (paid_get)

lgalabru:
## What

Brings the Rust SDK to ergonomic parity with the other pay-kit SDKs: a one-liner axum gate that accepts **both** MPP charge and x402 on a single route, in the abstracted `paid_get` / `Payment` style (not intent-specific).

```rust
let pay = PayKit::new(PayKitConfig {
    recipient: "CXhr…MPY".to_string(),
    ..Default::default()
})?;

let app = Router::new().route("/report", paid_get(report, "0.10", &pay));

async fn report(payment: Payment) -> String { payment.reference }
```

## How it works

- **`PayKit::new(PayKitConfig)`** (in `solana-pay-kit`) derives both an `Mpp` charge handler and an `X402` handler from one shared config. The optional `fee_payer_signer` drives MPP fee-sponsorship and supplies x402's fee-payer address from the same key.
- **`paid_get` / `paid_post`** gate a handler. An unpaid request gets a **402 carrying both challenges** — `WWW-Authenticate` (MPP) and `PAYMENT-REQUIRED` (x402). The headers are disjoint, so the paid retry is detected unambiguously (`Authorization: Payment` → MPP, `PAYMENT-SIGNATURE`/`X-PAYMENT` → x402) and verified by the matching protocol.
- The resolved price is pinned during verification, so a credential minted for a cheaper route on the same server is rejected.
- **`Payment`** handler extractor exposes `amount` / `protocol` / `reference`; the response carries the protocol's settlement header.
- **`Price::dynamic`** prices a route per request.

Supporting change in `solana-mpp`: a reusable `Mpp::verify_payment_for_amount(credential, amount)` (the amount-pinned verify the gate builds on). The existing `MppCharge<C>` typed extractor is unchanged.

## Tests / docs

`cargo test -p solana-pay-kit --features axum` — 5 gate tests + a doctest, including "unpaid → 402 with both protocol challenges" and "cross-route replay → 402". Adds a runnable `examples/axum_quickstart.rs` and a crate README.

## Notes

Separate from the cross-language audit PR (#169). Independent of it — based directly on `main`.


EfeDurmaz16 (dismissed):
Light review of the gate logic and the amount-pinned verify — looks solid.

- **Fail-closed.** The handler runs only after a successful verify, and the `Payment` extension is inserted only on success, so the extractor is unreachable on an unpaid request; missing / unknown / failed payment all fall through to a 402. `challenge_response` builds the 402 best-effort, so a header-format error still denies access rather than opening it.
- **Cross-route replay pinned.** `verify_payment_for_amount` rebuilds the expected request from the resolved price (and x402 `process_payment` pins the same amount), so a credential minted for a cheaper route is rejected. `verify_payment_for_amount` reuses the existing `verify_credential_with_expected` — no new trust surface.
- **Protocol disambiguation** is unambiguous (`Authorization: Payment` → MPP, `PAYMENT-SIGNATURE`/`X-PAYMENT` → x402), MPP taking precedence when both are present.

Two non-blocking nits: `Err(_)` on verify swallows the reason (no log) — correct behavior, just harder to debug a rejected payment; and `query_param` returns the raw, undecoded query value, worth a doc note for the dynamic-price closure.

LGTM.

lgalabru:
Addressed the Greptile review in `27d49d1`:

- **P1 — percent-encoding bypass**: `PriceCtx::query_param` now percent-decodes (`%XX` + `+`), so `?tier=premi%75m` resolves to the premium price instead of falling through to the cheaper default. Added the `query_param_percent_decodes` regression test.
- **P2 — silent challenge-header omission**: each error arm in `challenge_response` now logs via `tracing::warn!` (with the amount and error), so a dropped MPP/x402 challenge header is diagnosable.
- **P2 — empty x402 reference**: `x402_reference` returns `Option`; a verified payment carrying no signature is now rejected with a fresh 402 (and a warn) rather than handing the handler/`Payment-Response` an empty reference.
- **P2 — no x402-path test**: added `x402_invalid_signature_returns_402` (a bad `PAYMENT-SIGNATURE` takes the x402 path and 402s, not falling through to MPP). The full x402 happy-path (valid credential → 200) is covered by the cross-language harness, which runs the rust/ts x402 clients against servers; a unit-level happy path would require constructing a signed on-chain envelope.

EfeDurmaz16 (approved):
Re-checked 27d49d1, all four hold:

- `percent_decode` covers `%XX` and `+` with safe bounds, so the `premi%75m` bypass is closed and the regression test pins it.
- `challenge_response` now warns on every dropped-header arm, so a dropped MPP/x402 challenge is visible to operators instead of silent.
- `x402_reference` returns `Option`, so an unsigned tx 402s instead of reaching the handler with an empty reference.
- `x402_invalid_signature_returns_402` locks the x402 rejection path.

Fail-closed gating and price pinning still hold. Good to go.

PR #172, opened 2026-06-15
chore(swift): drop per-user Xcode state, show demo screenshot in README

lgalabru:
## What

- **Remove tracked per-user Xcode state.** `xcuserdata/ludo.xcuserdatad` (personal scheme management) was committed under both `swift/.swiftpm/…` and the PayKitDemo `xcodeproj`. Removed both and added `xcuserdata/` to `swift/.gitignore` so it can't be re-committed.
- **Show the demo screenshot** in the `swift/README.md` Examples section, under the `PayKitDemo` (SwiftUI iOS app) entry.

Shared schemes (`xcshareddata/xcschemes/…`) are intentionally kept — those are meant to be tracked.


EfeDurmaz16 (approved):
Drops the committed `xcuserdata` plists and gitignores them, keeps the shared schemes, adds the demo screenshot to the README. Nothing tracked that shouldn't be. LGTM.

PR #173, opened 2026-06-17
Add example: cross-chain payments (pay-kit on Solana + Interline for Base/EVM)

Choppaaahh:
## What this adds

A small, zero-config FastAPI example showing **pay-kit composing with a neutral,
non-custodial cross-chain x402 router** ([Interline](https://github.com/Choppaaahh/interline-routes),
`pip install interline`). The same "premium report" capability is served two ways so the
difference is obvious:

| Route | Gated by | Accepts |
|---|---|---|
| `GET /solana/report` | **pay-kit** | Solana USDC |
| `GET /crosschain/report` | **Interline** | Base/EVM USDC *(add the Solana rail for both)* |
| `GET /health` | — | free |

pay-kit owns the Solana side exactly as in its own examples (`Gate.build(...)` +
the `RequirePayment` FastAPI dependency). Interline adds the EVM side with one
`Paywall(rails=[X402Rail(...)])` whose 402 challenge advertises a Base/EVM x402 offer —
so the **same service can be paid by an agent on Solana _or_ on Base**, and the deployer
still holds no funds (each rail settles buyer → wallet directly).

## Why it's useful for pay-kit users

A pay-kit-gated endpoint is Solana-native — great when buyers are on Solana, but an agent
paying from Base/another EVM chain can't settle against it. This example is a copy-paste
pattern for "keep pay-kit for Solana, put a neutral router in front for the chains pay-kit
doesn't cover," without giving up non-custody. Adding another chain is one list entry; the
route never changes.

## Boot-tested

Both routes run out of the box (pay-kit against its hosted Surfpool sandbox; Interline with
mock facilitators — no wallet/keys/chain needed):

```bash
pip install -e ".[fastapi]"   # pay-kit, from the python/ dir (not yet on PyPI)
pip install interline          # the cross-chain router
uvicorn server:app --port 8000

curl -i http://127.0.0.1:8000/solana/report        # 402 — Solana challenge
curl -i http://127.0.0.1:8000/crosschain/report    # 402 — Base/EVM challenge
```

`GET /health` → 200, both paid routes → 402 with the correct per-chain `accepts` offer.

## Notes

- Non-custodial on both sides — neither library ever holds funds.
- `EVM_PAY_TO` is read from env with a placeholder default + a startup warning, so the mock
  demo is safe and going live just means exporting your wallet address.
- Happy to adjust framing, path, or trim the README if you'd prefer it leaner — just say the word.

### DIFF
--- a/python/examples/interline-cross-chain/README.md
+++ b/python/examples/interline-cross-chain/README.md
@@ -1,0 +1,63 @@
+# pay-kit on Solana + Interline for Base/EVM
+
+> One service, agents on either chain pay it.
+
+pay-kit gives you Solana-native agent payments (x402 + MPP). That's exactly what
+you want when your buyers are on Solana. But an agent paying from **Base or
+another EVM chain** can't settle against a Solana gate.
+
+[**Interline**](https://github.com/Choppaaahh/interline-routes) is a neutral,
+non-custodial **cross-chain x402 router** (`pip install interline`). Put it in
+front of the same capability and your endpoint *also* accepts Base/EVM x402 — the
+chain pay-kit doesn't cover — and you still hold no funds (each rail settles
+buyer → your wallet directly).
+
+This example serves the **same** "premium report" two ways so the difference is
+obvious:
+
+| Route | Gated by | Accepts |
+|---|---|---|
+| `GET /solana/report` | pay-kit | Solana USDC |
+| `GET /crosschain/report` | Interline | Base/EVM USDC *(add the Solana rail for both)* |
+| `GET /health` | — | free |
+
+## Run
+
+```bash
+# pay-kit installs from this repo (the FastAPI extra) — run from the pay-kit python/ dir:
+pip install -e ".[fastapi]"
+# Interline — the cross-chain router (on PyPI):
+pip install interline
+# then, from this example folder:
+uvicorn server:app --port 8000
+```
+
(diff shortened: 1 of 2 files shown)


Choppaaahh:
Thanks for the thorough review — all fair. The two code points are addressed in the latest commit: `_evm_x402_requirements` now echoes the `resource` arg into the requirements (distinct 402 per route, not silently ignored), and `/crosschain/report` is now `async` with the blocking `gate()` call offloaded via `run_in_threadpool` so it can't tie up the event loop.

On the conflict-of-interest / supply-chain note — totally reasonable, and you're right that I authored `interline` (the PR body discloses this). I'm happy to make this comfortable to merge however you'd prefer: I can **genericize the example so it demonstrates the "put a neutral cross-rail router in front of pay-kit" pattern without hard-pinning `interline`** — it's one implementation of the pattern, not a requirement — or mark it clearly as a third-party option, or relocate it. The value here is the composition pattern; the specific router is swappable.

Separately, good catch on the top-level `router` namespace — I'm fixing that in `interline` itself (→ `interline.router`) regardless, since squatting a generic top-level name is a real collision risk.

lgalabru:
Thanks for opening this PR @Choppaaahh.
Adding multi-chain support in pay-kit - already a multi-language repo - does not seem like a great approach from an A&D point of view.
We stay opened to Solana related contributions!

PR #177, opened 2026-06-19
feat(rust/x402): add support for `batch-settlement` scheme 

lgalabru:
## What

Adds the x402 **`batch-settlement`** scheme to the Rust SDK — the high-throughput sibling of `upto` (#174). A client **deposits once** into an escrow channel, signs a **cumulative Ed25519 voucher per request** (verified off-chain and served immediately, no on-chain tx in the request path), and the operator **redeems the latest voucher per channel later, in batches**. Per-request cost and latency approach zero.

On SVM this is the MPP **`session`** model surfaced over the x402 wire, backed by the **same** payment-channels program + 48-byte voucher already used by `upto`.

SVM spec: [`specs/schemes/batch-settlement/scheme_batch_settlement_svm.md`](specs/schemes/batch-settlement/scheme_batch_settlement_svm.md).

## Commits (review in order)

1. **`refactor(rust/core): hoist wire-agnostic session/channel logic into core`** — `solana-x402` deliberately doesn't depend on `solana-mpp`, so the shared channel/voucher/session logic moves into `solana-pay-core` (mirrors the earlier `payment_channels` / `parse_units` extractions). `mpp` re-exports the moved paths and delegates to the new helpers; its full suite stays green, which is the guardrail that the extraction is behavior-preserving.
   - `core::store` ← channel store + state + `MemoryChannelStore`
   - `core::voucher::verify_voucher_signature` (Ed25519)
   - `core::session::accept_voucher` (monotonic + deposit cap + idempotent replay + atomic CAS advance)

2. **`feat(rust/x402): add the batch-settlement scheme on Solana`** — the scheme, self-contained in x402 (reuses core, not mpp):
   - `protocol::schemes::batch_settlement` — wire types (`deposit`/`voucher`/`refund` payload union, requirements, settlement response) + pure verify
   - `server::batch_settlement::X402BatchSettlement` — `challenge`, `verify_payment` (deposit broadcasts+binds the open, **reusing upto's SOL-drain guard**; voucher accepts via `core::session`; refund settle-and-finalizes), and the operator primitives `settle_batch` (packs ≤3 channels/tx, chunked) + `distribute`
   - `client::batch_settlement` — `build_deposit`, a `BatchChannel` cumulative tracker, encode/parse
   - kit gate: `paid_batch_get`/`paid_batch_post` + `batch_gate_middleware` (verify voucher → serve; **settlement deferred**), `pay.x402_batch()` for out-of-band `settle_batch`/`distribute`; `examples/axum_batch.rs`
   - README: `batch-settlement` → ✅

## Design notes

- **SVM divergence from EVM:** the SVM `settle` instruction draws the voucher's **exact** cumulative (not EVM's "claim up to a ceiling"). So vouchers commit the actual cumulative; EVM's `claim`/`settle` maps to our `settle` (watermark) / `distribute` (sweep). Consequently **v1 = fixed per-request price**; dynamic pricing below the advertised amount needs a commit round-trip — deferred.
- **Client is the `authorizedSigner`** (client-voucher / `session` model), unlike `upto` where the operator signs. The operator only ever submits on-chain `settle`/`distribute`/`finalize` — it can't exceed the on-chain `deposit` or redirect funds away from the fixed `payee`.
- **Settlement is operator-driven and out of the request path** (v1). The automatic background channel manager (cron cadence + forced-close watchdog), durable `ChannelStore` wiring, and metered/dynamic pricing are documented follow-ups.

## Testing

```
cargo test -p solana-pay-core -p solana-x402 -p solana-pay-kit --features solana-pay-kit/axum   # green
cargo fmt --check                                                                                # clean
cargo clippy ... --all-targets                                                                   # clean*
```

\* the one `large_enum_variant` warning is pre-existing in `mpp::session.rs`, pulled in transitively — not from this change.

No on-chain E2E in CI (needs Surfnet); the off-chain logic is unit-tested and the on-chain builders are the same ones `upto` exercises.


rust/crates/x402/src/server/batch_settlement.rs
```diff
@@ -0,0 +1,745 @@
...
+        &self,
+        channel_id: &str,
+        voucher: Option<BatchVoucher>,
+    ) -> Result<BatchOutcome, Error> {
+        // Accept an optional final voucher first so it settles in the close.
+        if let Some(v) = &voucher {
+            self.accept(channel_id, v).await?;
+        }
+        let sig = self.settle_and_finalize(channel_id).await?;
+        let distribute_sig = self.distribute(channel_id).await?;
```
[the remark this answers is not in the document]
lgalabru: Done in a5089d0 — guarded on the watermark. Rather than a second \`get_channel\`, I reuse the state returned by the \`update_channel\` that freezes the channel (see the P1 fix below), so \`distribute\` is skipped whenever \`frozen.cumulative == 0\`. \`settle_and_finalize\` already settles 0 in that case, so the close still finalizes cleanly with a single tx.

lgalabru:
Addressed all three findings in a5089d0 (no behavior change to the happy path).

**P1 — settlement-window race (`process_refund`):** Fixed. After accepting the optional final voucher, `process_refund` now atomically sets `close_requested_at` via `update_channel` *before* any on-chain work. `accept_voucher` already rejects vouchers once `close_requested_at.is_some()`, so a concurrent request can no longer advance the watermark past what `settle_and_finalize` reads — closing the window where a voucher would be accepted off-chain yet be unrecoverable on-chain. `settle_and_finalize` then reads the now-frozen watermark.

**P2 — no-op `distribute`:** Fixed. Guarded on the frozen `cumulative > 0`. I reuse the state returned by the freeze `update_channel` instead of issuing another `get_channel`, so when no voucher was ever accepted we skip the sweep entirely (and drop a redundant read in the process).

**P2 — `paid_out` always `"0"`:** Fixed/clarified. `snapshot` now takes `paid_out` explicitly: `0` for the `open` snapshots (genuinely un-swept), and the settled watermark at `finalized` *only when `distribute` actually landed* (it sweeps the full pool, so on-chain `paidOut == settled`). Documented on both the method and the `BatchChannelSnapshot::paid_out` wire field that it's the server's own accounting, not a fresh on-chain read.

No new unit test: this handler has no `mod tests` and `process_refund` drives live RPC, so there's no harness to extend without inventing RPC mocking. The freeze's correctness rests on `accept_voucher`'s `close_requested_at` rejection, which the core `session`/`store` suites already cover.

rust/crates/x402/src/server/batch_settlement.rs
```diff
@@ -0,0 +1,788 @@
...
+                network: caip2_network_for_cluster(&self.config.cluster).to_string(),
+                amount: String::new(),
+                charged_amount: None,
+                channel_state: Some(
+                    self.snapshot(channel_id, frozen.deposit, paid_out, "finalized")
+                        .await,
+                ),
+            },
+        })
+    }
```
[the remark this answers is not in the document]
lgalabru: Fixed in 6bade22. `process_refund` now requires the refund payload to carry a voucher signed by the channel's `authorized_signer`:

- voucher that advances the watermark → accepted via `accept` (which verifies the signature) and settled in the close;
- voucher at/below the watermark → signature verified for proof-of-ownership only (no advance).

A refund with no voucher is rejected outright, so a known `channel_id` alone can no longer force a close. Added a client `BatchChannel::refund_payload()` that signs the proof voucher at the current cumulative, plus two unit tests (`refund_without_voucher_is_rejected`, `refund_with_unauthorized_signer_is_rejected`) — both assert the channel is left open.

One honest caveat: the voucher itself travels in the header, so an on-path observer who captures an *unexpired* voucher could still replay it as a close. Fully replay-proof close needs a dedicated close-auth (nonce/timestamp signed for the close intent) — a documented follow-up, not in this PR.

rust/crates/x402/src/server/batch_settlement.rs
```diff
@@ -0,0 +1,788 @@
...
+                .sign_transaction(&mut tx)
+                .await
+                .map_err(|e| Error::Other(format!("settle signing failed: {e}")))?;
+            let sig = self
+                .rpc
+                .send_and_confirm_transaction(&tx)
+                .map_err(|e| Error::Rpc(format!("settle broadcast failed: {e}")))?;
+            signatures.push(sig.to_string());
+        }
+        Ok(signatures)
```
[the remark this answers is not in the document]
lgalabru: Fixed in 6bade22 — `settle_batch` now skips any channel whose `highest_voucher_expires_at <= now_unix()` before building its settle instructions, right alongside the existing "no voucher" / `cumulative == 0` skips. An expired voucher can never settle on-chain anyway, so dropping it is strictly correct and keeps it from failing — and atomically aborting — a transaction it shares with still-valid channels.

rust/crates/x402/src/server/batch_settlement.rs
```diff
@@ -0,0 +1,906 @@
...
+            &expected_mint,
+            &authorized_signer,
+            salt,
+            &program_id,
+        );
+        let mut tx = decode_transaction(&transaction)?;
+        validate_open_instruction(
+            &tx,
+            &program_id,
+            &self.operator,
```
EfeDurmaz16: This passes operator, but batch channels authorize payer.

rust/crates/x402/src/server/batch_settlement.rs
```diff
@@ -0,0 +1,906 @@
...
+        let deposit = self
+            .store
+            .get_channel(channel_id)
+            .await
+            .ok()
+            .flatten()
+            .map(|s| s.deposit)
+            .unwrap_or(0);
+        Ok(BatchOutcome {
+            serve: true,
```
EfeDurmaz16, line 534: This serves zero-delta replays and underpriced vouchers.

rust/crates/x402/src/server/batch_settlement.rs
```diff
@@ -0,0 +1,906 @@
...
+        if pc::from_address(&channel.payee) != expected_payee {
+            return Err(Error::RecipientMismatch {
+                expected: pc::pubkey_string(&expected_payee),
+                actual: pc::pubkey_string(&pc::from_address(&channel.payee)),
+            });
+        }
+        if pc::from_address(&channel.authorized_signer) != authorized_signer {
+            return Err(Error::Other("channel authorized_signer mismatch".into()));
+        }
+        if pc::from_address(&channel.payer) != payer {
```
EfeDurmaz16, line 426: Bind distribution, grace, and deposit before storing.

EfeDurmaz16 (changes requested):
Requesting changes because the new batch-settlement path has blocking correctness issues. The deposit flow validates the open transaction with the operator as the authorized signer, while this scheme and client use the payer, so valid channels fail before first serve. The voucher path also accepts any monotonic delta, including one base unit or a replay with zero charged amount, then still serves the protected route. Finally, the server stores and serves after checking only basic channel identity, without binding distribution splits, grace period, or deposit to the advertised requirements. These should be fixed before merge.

EfeDurmaz16 (dismissed):

rust/crates/x402/src/server/batch_settlement.rs
lgalabru: Good catch — this was a blocking bug. `validate_open_instruction` checks account position 3 against the value passed as `operator`, and batch channels put the **payer** there (client-signs-vouchers model), so every valid open failed with "authorized_signer mismatch". Fixed in ed0d5a7 by passing `&authorized_signer` (the config value already used for the PDA derivation and the on-chain bind) instead of `&self.operator`.

rust/crates/x402/src/server/batch_settlement.rs
lgalabru: Fixed in ed0d5a7. `verify_payment` now threads `per_request` into both voucher paths and rejects when the delta is below the advertised price — *before* advancing the watermark. That covers underpriced advances and idempotent replays of the latest voucher (delta 0), which previously served the route again for free. The first voucher in the deposit path is checked the same way (the `_per_request` argument was unused). Added unit tests `underpriced_voucher_is_rejected_without_advancing` and `replayed_voucher_is_rejected`.

Note: a replay now returns an error rather than an idempotent cached response — for a paid gate that is the safe default (no free serve); request-level idempotency caching would be a separate feature.

rust/crates/x402/src/server/batch_settlement.rs
lgalabru: Fixed in ed0d5a7. After confirming the channel, `process_deposit` now binds — in addition to mint/payee/signer/payer — the economically-relevant terms to what the server advertised, before `put_channel`:
- `channel.deposit >= per_request` (rejects an under-funded channel that cannot cover the served request),
- `channel.grace_period == config.grace_period_seconds`,
- `channel.distribution_hash == pc::distribution_hash(self.distributions())` (rejects splits that redirect proceeds away from the payee).

This mirrors how `upto` binds its deposit/distribution. A configurable explicit `minimum_deposit` (vs. the one-request floor) can come later if you want a higher bar.

EfeDurmaz16 (approved):
LGTM! Thanks

PR #182, opened 2026-06-23
refactor: improve payment-channel ops

lgalabru:


rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,521 @@
...
+
+            tokio::select! {
+                maybe = rx.recv() => match maybe {
+                    Some(unit) => {
+                        pending.push(unit);
+                        if deadline.is_none() {
+                            deadline = Some(Instant::now() + cfg.linger);
+                        }
+                        // Size trigger: seal full transactions eagerly.
+                        while pending.len() >= cfg.max_channels_per_tx {
```
EfeDurmaz16: Could we clamp `max_channels_per_tx` before this loop? Because the field is public, a value of `0` makes `pending.len() >= 0` always true, `drain(..0)` removes nothing, and the worker hot-loops while spawning empty flush tasks.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,521 @@
...
+        tx_sig = tracing::field::Empty,
+        tx_bytes = tracing::field::Empty,
+        latency_ms = tracing::field::Empty,
+    );
+    for u in &group {
+        if let Some(id) = u.parent.id() {
+            span.follows_from(id);
+        }
+    }
+    let _enter = span.enter();
```
EfeDurmaz16: Could we avoid holding this entered span across the awaited block below? This compiles (`Entered<'_>` is `Send`), but tracing's docs warn it can misattribute unrelated async work after a yield; `async { ... }.instrument(span).await` or scoped `in_scope` would keep traces correct.

rust/crates/mpp/src/server/session.rs
```diff
@@ -158,6 +185,42 @@ pub struct FinalizeParams {
...
+            }
+            _ => None,
+        };
+        payment_channels::build_settle_and_finalize_instructions(
+            operator,
+            &self.channel_id,
+            &authorized_signer,
+            signature.as_ref(),
+            self.settled,
+            self.voucher_expires_at.unwrap_or(0),
```
EfeDurmaz16: Can we make a missing expiry an error when a voucher signature is present? With `unwrap_or(0)`, a signed positive settlement can build a voucher message with epoch-0 expiry, which won't match the signed bytes and will fail on-chain verification.

EfeDurmaz16 (changes requested):
Requesting changes on a few correctness/operability issues before merge. The worker builds/tests fine, so the earlier `!Send` concern does not reproduce, but there are still edge cases worth fixing: clamp `max_channels_per_tx` so `0` cannot hot-loop the worker, avoid holding a tracing `enter()` guard across `.await`, and fail closed when a voucher signature lacks an expiry instead of defaulting to `0`. Also please run `cargo fmt`—`cargo fmt --check` currently reports diffs. After those are addressed, the shared packing extraction and event_authority fix look good.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,521 @@
...
+        let blockhash = broadcaster.latest_blockhash().await?;
+        let message = Message::new_with_blockhash(&flat, Some(&cfg.operator), &blockhash);
+        let mut tx = Transaction::new_unsigned(message);
+        cfg.operator_signer
+            .sign_transaction(&mut tx)
+            .await
+            .map_err(|e| format!("settle signing failed: {e}"))?;
+        broadcaster.send(&tx).await
+    }
+    .await;
```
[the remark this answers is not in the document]
lgalabru: Fixed in ce3c91e by switching to `async { … }.instrument(span).await`. One correction on the mechanism, though: this is **not** a compile error. `Span::enter()` returns `tracing::span::Entered<'_>`, which in tracing 0.1.44 is just `{ span: &Span }` and is `Send` — only `EnteredSpan` (from `.entered()`) carries `PhantomNotSend`. I confirmed `cargo build -p solana-pay-core --features worker` succeeds on the unmodified branch, and `tokio::spawn` requires `Send` regardless of runtime flavor. The real defect is the one it still had — holding the guard across `.await` misattributes spans after a yield — which `Instrument` fixes. So it's a P2 trace-correctness bug, not a P1 build break.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,521 @@
...
+
+/// Re-group units to be byte- and count-safe, then settle each group as one tx.
+fn spawn_flush(
+    units: Vec<SettlementUnit>,
+    cfg: Arc<SettlementConfig>,
+    broadcaster: Arc<dyn Broadcaster>,
+    sem: Arc<Semaphore>,
+) {
+    tokio::spawn(async move {
+        let _permit = sem.acquire_owned().await;
```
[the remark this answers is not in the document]
lgalabru: Fixed in ce3c91e: `acquire_owned().await.expect("settlement semaphore is never closed")`. The worker owns the semaphore and never calls `close()`, so the `Err` arm is unreachable today; the `expect` makes the invariant explicit so a future change that closes it fails loudly instead of silently bypassing `max_in_flight_tx`.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,521 @@
...
+                .value
+                .into_iter()
+                .next()
+                .flatten()
+                .map(|s| s.err.is_none())
+                .unwrap_or(false))
+        })
+        .await
+        .map_err(|e| e.to_string())?
+    }
```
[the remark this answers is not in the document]
lgalabru: Fixed in ce3c91e: `RpcBroadcaster` now holds one `Arc<RpcClient>` built in `new`, cloned into each `spawn_blocking`, so the connection pool is reused across `latest_blockhash`/`send`/`confirm`.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,521 @@
...
+                    Some(unit) => {
+                        pending.push(unit);
+                        if deadline.is_none() {
+                            deadline = Some(Instant::now() + cfg.linger);
+                        }
+                        // Size trigger: seal full transactions eagerly.
+                        while pending.len() >= cfg.max_channels_per_tx {
+                            let batch = pending.drain(..cfg.max_channels_per_tx).collect();
+                            spawn_flush(batch, cfg.clone(), broadcaster.clone(), sem.clone());
+                        }
```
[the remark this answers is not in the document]
lgalabru: Leaving this as-is, deliberately. Correctness is already guaranteed: `regroup` inside `spawn_flush` does the byte+count split and the linger timer bounds the wait. The count-only eager trigger is purely a latency optimization for the uncommon case where one channel's instructions are large enough that fewer than `max_channels_per_tx` fit in 1232 bytes — which doesn't arise for the calibrated settle/finalize path (3/tx). Duplicating `regroup`'s byte logic in the hot select loop to save <=350ms in a rare case isn't worth it; happy to revisit if an over-large-channel workload shows up.

rust/crates/core/src/settlement/worker.rs
lgalabru: Fixed in ce3c91e — the eager trigger now uses `let max_per_tx = cfg.max_channels_per_tx.max(1)` for both the loop condition and the `drain`, so a misconfigured `0` can't hot-loop draining nothing while spawning empty flush tasks. (`max_in_flight_tx` was already clamped via `.max(1)` at semaphore creation.)

rust/crates/core/src/settlement/worker.rs
lgalabru: Agreed, and you had the mechanism exactly right (it compiles — `Entered<'_>` is `Send`; only the trace attribution is wrong). Fixed in ce3c91e by wrapping the body in `async { … }.instrument(span).await` and dropping the manual `enter()`.

rust/crates/mpp/src/server/session.rs
lgalabru: Fixed in ce3c91e. When a voucher signature is present (settled > 0), a missing `voucher_expires_at` is now an error rather than defaulting to `0` — since the signed voucher message commits to its expiry, `unwrap_or(0)` would build bytes the signature can't match and fail on-chain verification. The no-voucher finalize path still uses `0`.

EfeDurmaz16 (dismissed):
LGTM !

EfeDurmaz16 (dismissed):

efebarandurmaz (commented):
## Thermo-Nuclear Code Quality Review

### Summary

This PR does a significant protocol refactor: `SettleArgs` is deleted, `SettleAndFinalizeArgs` reduced to `hasVoucher: u8` (voucher lives exclusively in the ed25519 precompile), a new batched-settlement tokio actor is introduced, shared greedy packing is extracted, and the program ID is updated to `CHNLxY…` across all 6 SDKs (Rust, Go, TS, Kotlin, Swift, Python). The TS client is migrated from hand-vendored to full codama codegen.

EfeDurmaz16's previous review comments (span.enter, max_channels clamp, voucher expiry guard, semaphore unwrap) are all addressed in commit ce3c91e. The settlement worker is now well-hardened.

**Overall: APPROVE with findings.** The protocol change is clean and consistently applied. The main gaps are in test coverage of failure modes (no negative tests for the settlement worker) and the absence of CI enforcement that generated code stays in sync with the IDL.

### Moderate Findings

| # | File | Issue | Suggested Remedy |
|---|---|---|---|
| 1 | `worker.rs:197` | `regroup` duplicates the greedy-packing algorithm from `packing.rs` with the same byte-probe logic written twice | Extract the boundary predicate (`exceeds_limit`) into `packing.rs` so both callers share it |
| 2 | `otel.rs:113` | OTLP error handler kills all three signals (trace, metrics, logs) when a single one fails | Initialize each provider independently so a metrics endpoint failure doesn't silence traces |
| 3 | `settlement_onchain.rs` | No negative tests: stale voucher, signature mismatch, duplicate finalize — all silently unexercised | Add at least one negative test (expired voucher → expected program error) |

### Low-Severity Notes

| # | File | Issue |
|---|---|---|
| 4 | `worker.rs:375` | `confirm()` maps on-chain failure and "not found" to `Ok(false)` — retry spins uselessly on permanent failures | Consider tri-state return or at minimum log the on-chain error |
| 5 | `settlement/packing.rs:46` | `pack()` loop reconstructs + serializes the full message per channel addition (O(n²) in bytes) | Document as accepted cost for realistic batch sizes (~3-10 channels) |
| 6 | `mpp/src/server/session.rs:59` | `SettlementAuthority` is well-designed but lacks `#[non_exhaustive]` — new variants would silently break downstream matches | Add `#[non_exhaustive]` to force future-proof match arms |
| 7 | `on-chain.ts:74` | Hand-vendored `PAYMENT_CHANNELS_PROGRAM_ID` duplicates the generated one in `programs/paymentChannels.ts` | Import from the generated module to prevent future drift |
| 8 | `python/.../_paymentchannels.py:85` | Duplicate hand-vendored program ID alongside the generated `program_id.py` | Either import the generated one or add a CI assertion both match |

### Code-Judo Wins

- **`packing.rs` extraction** — replacing ad-hoc `chunks(3)` in x402 with shared greedy packing is the right move. The `pack()` function correctly accounts for the 1232-byte limit, signature prefix, and compact-u16 overhead.
- **TS codegen migration** — replacing the minimal vendored subset with full codama-generated clients removes a maintenance burden. The generated client now covers all instructions (open, settle, settleAndFinalize, topUp, distribute, etc.) instead of only the handful that were hand-copied.
- **Settle instruction simplification** — reducing settle to a 1-byte discriminator with zero args is a clean protocol win. The voucher moving exclusively to the ed25519 precompile eliminates a whole class of deserialization errors.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,577 @@
...
+            .expect("settlement semaphore is never closed");
+        for group in regroup(units, &cfg.operator, cfg.max_channels_per_tx) {
+            settle_group(group, &cfg, broadcaster.clone(), trigger).await;
+        }
+    });
+}
+
+/// Greedy byte+count packing over units (mirrors [`super::packing::pack`] but
+/// keeps the per-unit reply channels).
+fn regroup(
```
EfeDurmaz16, line 214: `regroup` duplicates the greedy packing algorithm from `packing.rs::pack` — same byte-probe logic, same batch-cut logic. Extract a shared `exceeds_limit()` boundary predicate into `packing.rs` so changes to packing rules don't need mirroring in two places.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,577 @@
...
+            .map_err(|e| e.to_string())
+    }
+
+    async fn confirm(&self, signature: &str) -> Result<bool, String> {
+        use solana_signature::Signature;
+        let rpc = self.rpc.clone();
+        let sig: Signature = signature.parse().map_err(|_| "bad signature".to_string())?;
+        tokio::task::spawn_blocking(move || {
+            let statuses = rpc
+                .get_signature_statuses(&[sig])
```
EfeDurmaz16, line 399: `get_signature_statuses` returning `Some(TransactionStatus { err: Some(...) })` (permanent failure) is mapped to `Ok(false)`, same as "not found yet". The retry loop then spins 10×400ms before giving up — even for permanently failed txs. Consider a tri-state return or at minimum log the on-chain error for diagnostics.

rust/crates/core/src/settlement/worker.rs
```diff
@@ -0,0 +1,577 @@
...
+    broadcaster: Arc<dyn Broadcaster>,
+    sem: Arc<Semaphore>,
+    trigger: &'static str,
+) {
+    tokio::spawn(async move {
+        // The worker owns the semaphore and never closes it, so acquire cannot
+        // fail; make that explicit — an `Err` would otherwise silently proceed
+        // without a permit and bypass the `max_in_flight_tx` limit.
+        let _permit = sem
+            .acquire_owned()
```
EfeDurmaz16: One semaphore permit covers the entire `spawn_flush` call, which may produce multiple transaction groups via `regroup`. `max_in_flight_tx` is really `max_concurrent_flushes`. Worth renaming to avoid confusion, or moving the acquire into `settle_group` for per-TX backpressure.

rust/crates/core/src/settlement/packing.rs
```diff
@@ -0,0 +1,167 @@
...
+    let sig_prefix = if sigs < 128 { 1 } else { 2 };
+    msg.serialize().len() + sigs * 64 + sig_prefix
+}
+
+/// Greedily group channels into legacy-tx-sized batches. Each returned batch's
+/// flattened instructions serialize to `<= MAX_TX_BYTES` and hold at most
+/// `max_per_tx` channels. A single channel that alone exceeds the limit is
+/// returned as its own (over-size) batch — the caller surfaces that as an error
+/// at broadcast rather than silently dropping it.
+pub fn pack(
```
EfeDurmaz16, line 70: The inner loop reconstructs + serializes the full message for every channel addition — O(n²) in instruction bytes. Fine for realistic batch sizes (3-10 channels), but worth a doc comment acknowledging the cost so a future reader doesn't try to optimize it blindly.

rust/crates/core/src/otel.rs
```diff
@@ -0,0 +1,195 @@
...
+                        .with_filter(EnvFilter::new(opts.trace_filter)),
+                )
+                .with(logs_layer)
+                .try_init();
+            tracing::info!(endpoint = %base, service = %opts.service_name, "OTLP export enabled");
+            Guard {
+                providers: Some((tracer_provider, meter_provider, logger_provider)),
+            }
+        }
+        (Err(e), _, _) | (_, Err(e), _) | (_, _, Err(e)) => {
```
EfeDurmaz16: The catch-all `(Err(e), _, _) | (_, Err(e), _) | (_, _, Err(e))` kills all three signals when a single provider init fails. A metrics endpoint being down shouldn't silence traces or logs. Initialize each provider independently and log (don't swallow) partial failures.

rust/crates/mpp/src/server/session.rs
```diff
@@ -50,6 +50,33 @@ pub struct Split {
     pub bps: u16,
 }
 
+/// Who signs the channel's vouchers — the settlement authority model.
+///
+/// The simpler successor to the removed `multi_delegator` path. Both settle
+/// through the same batched worker; the difference is only who holds the
+/// channel's `authorized_signer`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum SettlementAuthority {
```
EfeDurmaz16, line 60: Good forward-looking design. Consider adding `#[non_exhaustive]` so future variants (e.g., `ThirdParty`) force downstream match arms to handle them explicitly instead of silently falling through.

EfeDurmaz16 (commented):
## Thermo-Nuclear Code Quality Review

### Summary

This PR does a significant protocol refactor: `SettleArgs` is deleted, `SettleAndFinalizeArgs` reduced to `hasVoucher: u8` (voucher lives exclusively in the ed25519 precompile), a new batched-settlement tokio actor is introduced, shared greedy packing is extracted, and the program ID is updated to `CHNLxY…` across all 6 SDKs (Rust, Go, TS, Kotlin, Swift, Python). The TS client is migrated from hand-vendored to full codama codegen.

EfeDurmaz16's previous review comments (span.enter, max_channels clamp, voucher expiry guard, semaphore unwrap) are all addressed in commit ce3c91e. The settlement worker is now well-hardened.

**Overall: APPROVE with findings.** The protocol change is clean and consistently applied. The main gaps are in test coverage of failure modes (no negative tests for the settlement worker) and the absence of CI enforcement that generated code stays in sync with the IDL.

### Moderate Findings

| # | File | Issue | Suggested Remedy |
|---|---|---|---|
| 1 | `worker.rs:197` | `regroup` duplicates the greedy-packing algorithm from `packing.rs` with the same byte-probe logic written twice | Extract the boundary predicate (`exceeds_limit`) into `packing.rs` so both callers share it |
| 2 | `otel.rs:113` | OTLP error handler kills all three signals (trace, metrics, logs) when a single one fails | Initialize each provider independently so a metrics endpoint failure doesn't silence traces |
| 3 | `settlement_onchain.rs` | No negative tests: stale voucher, signature mismatch, duplicate finalize — all silently unexercised | Add at least one negative test (expired voucher → expected program error) |

### Low-Severity Notes

| # | File | Issue |
|---|---|---|
| 4 | `worker.rs:375` | `confirm()` maps on-chain failure and "not found" to `Ok(false)` — retry spins uselessly on permanent failures | Consider tri-state return or at minimum log the on-chain error |
| 5 | `settlement/packing.rs:46` | `pack()` loop reconstructs + serializes the full message per channel addition (O(n²) in bytes) | Document as accepted cost for realistic batch sizes (~3-10 channels) |
| 6 | `mpp/src/server/session.rs:59` | `SettlementAuthority` is well-designed but lacks `#[non_exhaustive]` — new variants would silently break downstream matches | Add `#[non_exhaustive]` to force future-proof match arms |
| 7 | `on-chain.ts:74` | Hand-vendored `PAYMENT_CHANNELS_PROGRAM_ID` duplicates the generated one in `programs/paymentChannels.ts` | Import from the generated module to prevent future drift |
| 8 | `python/.../_paymentchannels.py:85` | Duplicate hand-vendored program ID alongside the generated `program_id.py` | Either import the generated one or add a CI assertion both match |

### Code-Judo Wins

- **`packing.rs` extraction** — replacing ad-hoc `chunks(3)` in x402 with shared greedy packing is the right move. The `pack()` function correctly accounts for the 1232-byte limit, signature prefix, and compact-u16 overhead.
- **TS codegen migration** — replacing the minimal vendored subset with full codama-generated clients removes a maintenance burden. The generated client now covers all instructions (open, settle, settleAndFinalize, topUp, distribute, etc.) instead of only the handful that were hand-copied.
- **Settle instruction simplification** — reducing settle to a 1-byte discriminator with zero args is a clean protocol win. The voucher moving exclusively to the ed25519 precompile eliminates a whole class of deserialization errors.

typescript/packages/mpp/src/server/session/on-chain.ts
```diff
@@ -72,25 +72,22 @@ export const ED25519_PROGRAM_ADDRESS =
  * matches the Rust mirror; callers can override per-call.
  */
 export const PAYMENT_CHANNELS_PROGRAM_ID =
```
EfeDurmaz16: This hand-vendored `PAYMENT_CHANNELS_PROGRAM_ID` duplicates the generated one in `programs/paymentChannels.ts`. If the IDL is regenerated, only the generated source updates. Import from the generated module instead.

rust/crates/mpp/tests/settlement_onchain.rs
```diff
@@ -0,0 +1,300 @@
+//! On-chain proof + bench/demo for the shared settlement worker, driven through
```
EfeDurmaz16, line 1: Strong happy-path coverage, but no negative tests: stale/expired voucher, signature mismatch, duplicate finalize, insufficient fee SOL. These would catch real program-level regressions. Add at least one expired-voucher → expected-program-error test.

rust/crates/core/src/settlement/worker.rs
lgalabru: Done in c769b44. Extracted the shared boundary rule into `packing.rs::would_overflow_tx(cur, cur_count, next, payer, max_per_tx)`; both `pack` and the worker's `regroup` now call it, so the count+byte packing logic lives in one place.

rust/crates/core/src/settlement/worker.rs
lgalabru: Done in c769b44. `Broadcaster::confirm` now returns a tri-state `ConfirmOutcome { Confirmed, Pending, Failed(String) }`. A permanent on-chain failure (`status.err` set) maps to `Failed(err)`; the retry loop logs it (with the error) and returns immediately instead of spinning the full 10×400ms window. `Err` stays reserved for transient RPC errors (still retried).

rust/crates/core/src/settlement/worker.rs
lgalabru: Done in c769b44 — moved the `acquire_owned()` into the per-group loop in `spawn_flush`, so one permit is held per settle transaction rather than per flush (a flush can `regroup` into several). `max_in_flight_tx` now genuinely bounds concurrent in-flight settle txs across all flushes.

rust/crates/core/src/settlement/packing.rs
lgalabru: Done in c769b44. The cost note now lives on `would_overflow_tx` (the shared predicate): it rebuilds + serializes the candidate message per call, so a greedy packer on it is O(n²) in instruction bytes — fine for a handful of channels, not for large fan-in.

rust/crates/core/src/otel.rs
lgalabru: Done in c769b44. Each OTLP provider is now initialized independently: a failed one is left `None` (its layer omitted) and the failure is logged, while the others still export. The console fmt layer is always installed. `Guard` holds the three providers separately and shuts down whichever came up. So a metrics endpoint being down no longer silences traces/logs.

rust/crates/mpp/src/server/session.rs
lgalabru: Done in c769b44 — added `#[non_exhaustive]` to `SettlementAuthority`.

typescript/packages/mpp/src/server/session/on-chain.ts
lgalabru: Done in c769b44. `on-chain.ts` now re-exports `PAYMENT_CHANNELS_PROGRAM_ID` from the generated client (`PAYMENT_CHANNELS_PROGRAM_ADDRESS`) instead of re-declaring the literal, so IDL regeneration is the single source of truth.

rust/crates/mpp/tests/settlement_onchain.rs
lgalabru: Added in c769b44 — `expired_voucher_is_rejected_on_chain` (asserts the program rejects with `VoucherExpired`/0xE9 and the channel stays Open), verified passing against the live surfnet.

While here we also found a related gap your instinct points at: `distribution_hash` was using blake3 while the deployed program commits with sha256. Fixed it, and added an on-chain **parity** test (`distribution_hash_matches_on_chain_commitment`) that opens a channel with splits and asserts pay-kit's hash equals the program's on-chain commitment — the kind of cross-impl check that would have caught both this and the treasury-owner drift. Thanks for pushing on negative/parity coverage.

EfeDurmaz16 (approved):
## Approval

All findings from the thermo-nuclear review have been substantively addressed:

**Structural fixes:**
- `would_overflow_tx` extracted to `packing.rs` — shared by `pack` and `regroup`, removing the DRY violation
- `ConfirmOutcome` tri-state enum replaces `Ok(false)` ambiguity
- OTLP providers initialized independently — single signal failure no longer kills all three
- Semaphore scoped per settle transaction (not per flush)
- O(n²) cost documented on the shared packing rule

**Correctness fixes:**
- `distribution_hash` changed from blake3 → sha2 for on-chain parity with `sol_sha256`
- Golden vector test + on-chain commitment parity E2E test added
- `expired_voucher_is_rejected_on_chain` negative test added

**Observability / docs:**
- Guard flushed/shutdown split per signal
- Large doc sections consolidated; README/support matrix cleaned

This is a well-structured refactor with careful parity testing. Ship it.

PR #184, opened 2026-06-25
feat(rust): cross-protocol, balance-aware payment selection

lgalabru:
## What

Adds `kit::select_payment` / `select_payment_parsed`: given the funded tokens a wallet is willing to spend (with available balances) and an ordering strategy, pick a charge option across **both** MPP charge challenges and x402 `accepts` that the wallet can actually fund.

### Why

A 402 can advertise one stablecoin over MPP and another over x402. Today the client commits to a protocol *before* considering balances, so a wallet holding only USDC (offered via x402) fails against a server whose MPP challenge is denominated in USDG — even though a fundable option was right there. Selection should be a wallet decision (which token can I spend?), not a protocol decision.

### Changes

- **core**: new canonical `mints` registry — single source of truth for stablecoin mint addresses.
- **mpp / x402**: re-export `mints` from core, removing the duplicated constants (there were three copies across the workspace).
- **x402**: add `parse_x402_accepts` returning *every* advertised accept (no client-side selection), so callers can rank offers themselves.
- **kit**: `AcceptableToken`, `OrderingStrategy` (`HighestBalance` / `CheapestPayable` / `FixedPriority` / `MppFirst`), `SelectedPayment`, and a `SelectError` that enumerates offered-vs-held for actionable diagnostics.

### Token language

One language crosses the boundary: the **mint address**. Advertised currencies (symbol or mint) are resolved to a mint via the shared registry before any comparison; `funded[].mint` must be canonical mints. Only `FixedPriority` accepts symbols, resolved through the same path.

## Test plan

- 7 new unit tests in `kit::select` covering the USDG-MPP/USDC-x402 case, each ordering strategy, the unfunded error, and amount-exceeds-balance.
- `cargo test -p solana-pay-kit --features client` ✅
- `cargo test -p solana-pay-core -p solana-mpp -p solana-x402 --features client` ✅ (mints re-export intact)
- fmt + clippy clean (kit).

### DIFF
--- a/rust/crates/kit/src/select.rs
+++ b/rust/crates/kit/src/select.rs
@@ -22,7 +22,7 @@ use crate::mpp::{
     parse_www_authenticate_all, resolve_stablecoin_mint, ChargeRequest, PaymentChallenge,
 };
 use crate::x402::client::exact::parse_x402_accepts;
-use crate::x402::exact::PaymentRequirements;
+use crate::x402::exact::{cluster_for_caip2_network, PaymentRequirements};
 
 /// A token the caller is willing to spend, with the balance currently available.
 ///
@@ -419,17 +419,23 @@ fn mpp_candidate(challenge: &PaymentChallenge, order: usize) -> Option<Candidate
 }
 
 /// Normalize one x402 accept to a [`Candidate`]. Returns `None` for
-/// SOL-denominated or unparseable requirements.
+/// SOL-denominated, unparseable, or non-Solana requirements.
 fn x402_candidate(requirement: &PaymentRequirements, order: usize) -> Option<Candidate> {
     if requirement.currency.eq_ignore_ascii_case("SOL") {
         return None;
     }
-    let network = normalize_network(
+    // Resolve the canonical CAIP-2 `network` (e.g. "solana:EtWT…" for devnet)
+    // — or an explicit `cluster` — to a cluster slug. A plain substring scan
+    // would miss CAIP-2 genesis-hash ids and mislabel them mainnet, so go
+    // through `cluster_for_caip2_network`, which also handles the short forms.
+    // `None` means the offer isn't on a recognized Solana network → skip it.
+    let cluster = cluster_for_caip2_network(&requirement.network).or_else(|| {
         requirement
             .cluster
             .as_deref()
-            .unwrap_or(&requirement.network),
-    );
+            .and_then(cluster_for_caip2_network)
+    })?;
+    let network = normalize_network(cluster);
     let mint = resolve_mint(&requirement.currency, &network)?;
     let amount = requirement.amount.parse::<u64>().ok()?;
     Some(Candidate {
@@ -647,5 +653,36 @@ mod tests {
         let selected =
             select_payment(&headers, Some(&body), &funded, &OrderingStrategy::MppFirst).unwrap();
         assert_eq!(selected.protocol(), "mpp");
+    }
+
+    // Regression: an x402 offer that carries a canonical CAIP-2 devnet network
+    // id and *no* human `cluster` field must be recognized as devnet (not
+    // silently mislabeled mainnet and dropped). Uses the real devnet genesis
+    // hash so the deserializer copies it into `cluster`.
+    #[test]
+    fn caip2_devnet_network_without_cluster_is_recognized() {
+        // solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1 == devnet (x402 SOLANA_DEVNET).
+        let body = serde_json::json!({
+            "x402Version": 1,
+            "accepts": [{
+                "network": "solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1",
+                "recipient": RECIPIENT,
+                "amount": "1000",
+                "currency": "USDC",
+                "decimals": 6,
+                "resource": "https://example.com/r",
+            }],
+        })
+        .to_string();
+        let funded = vec![AcceptableToken {
+            mint: mints::USDC_DEVNET.to_string(),
+            network: "devnet".to_string(),
+            available: 1_000_000,
+        }];
+
+        let selected = select_payment(&[], Some(&body), &funded, &OrderingStrategy::HighestBalance)
+            .expect("CAIP-2 devnet offer should be fundable with devnet USDC");
+        assert_eq!(selected.protocol(), "x402");
+        assert_eq!(selected.mint(), mints::USDC_DEVNET);
     }
 }
--- a/rust/crates/kit/src/select.rs
+++ b/rust/crates/kit/src/select.rs
@@ -1,0 +1,651 @@
+//! Cross-protocol payment selection.
+//!
+//! A 402 response can advertise charge options across *both* protocols — MPP
+//! charge challenges and x402 `accepts` entries — often in different
+//! stablecoins (e.g. USDG over MPP, USDC over x402). Picking which one to pay
+//! is a wallet decision, not a protocol decision: the client should settle in
+//! a token it actually holds, in its own priority order, regardless of which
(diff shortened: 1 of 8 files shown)


lgalabru:
Addressed all three findings in `ef62555`:

- **P1 — CAIP-2 network misclassification (`normalize_network`)**: `x402_candidate` now resolves the network through `cluster_for_caip2_network` (canonical genesis-hash ids *and* short forms) before the slug scan, so a `solana:EtWT…` devnet offer with no `cluster` field is correctly recognized as devnet instead of defaulting to mainnet. Non-Solana offers are now skipped. Added a regression test (`caip2_devnet_network_without_cluster_is_recognized`).
- **P2 — `serde_json` unconditional**: made it `optional = true`, pulled in by the `client` feature that gates the `select` module.
- **P2 — duplicated header scanning**: extracted a shared `payment_required_envelope(headers, body) -> Option<PaymentRequiredEnvelope>` helper; both `parse_x402_challenge_with_selection` and `parse_x402_accepts` use it (the legacy v1 single-requirement header is wrapped as a one-element envelope).

`cargo test -p solana-pay-kit -p solana-x402 --features client` green; fmt + clippy clean.

EfeDurmaz16 (commented):
## Summary

A well-structured PR overall. The mints deduplication is overdue and cleanly done; the `parse_x402_accepts` + `payment_required_envelope` extraction is a good refactoring that reveals structure rather than hiding it. The selection module is cohesive and follows a clear top-down organization.

### What I would like to see improved

**1. normalize_network should use structured types (MEDIUM)** — `contains("devnet")` / `contains("testnet")` is fragile. The codebase already has `cluster_for_caip2_network` in x402; either use that here too or extract a shared canonical network type into core.

**2. f64 for monetary comparison (MEDIUM)** — `human_amount` converts u64 base units to f64 for ranking. f64 cannot represent values like 10_000.000001 exactly; the comparison is non-deterministic for edge amounts. Compare in base units with decimal adjustment instead.

**3. resolve_mint passes through unknown symbols (LOW)** — If a currency is neither a recognized symbol nor a valid mint address, it passes through silently. A typo like "USC" would produce a confusing "no match" error.

**4. mpp_candidate silently defaults on parse failure (LOW)** — `serde_json::from_value(v).ok().unwrap_or_default()` swallows malformed `methodDetails` without logging. A misconfigured server would be invisible.

**5. rank duplicates the tiebreaker tail across all four arms (LOW)** — `.then(source.rank()).then(order)` appears in every arm. Extract a helper closure.

**6. select_from_candidates O(n*m) undocumented (LOW)** — Nested filter_map over funded×candidates deserves a doc comment.

### Positive

- Mint constant deduplication: long overdue, correctly done with re-exports.
- `payment_required_envelope` consolidation removes two narrow helpers cleanly.
- `parse_x402_accepts` is exactly the right API for cross-protocol selection.
- `SelectError` carries both offered and held — excellent for diagnostics.
- Test coverage is thorough and covers the cross-protocol fundability case.

### Verdict

No structural regressions, no file-size issues, no spaghetti growth. The `normalize_network` and `human_amount` concerns should be addressed before shipping, but the overall design is sound.

rust/crates/kit/src/select.rs
```diff
@@ -0,0 +1,688 @@
...
+    if currency.eq_ignore_ascii_case("SOL") {
+        return None;
+    }
+    resolve_stablecoin_mint(currency, Some(network))
+        .map(str::to_string)
+        .or_else(|| Some(currency.to_string()))
+}
+
+/// Collapse the various network spellings to a canonical slug.
+fn normalize_network(raw: &str) -> String {
```
EfeDurmaz16, line 465: Ad-hoc `contains()` on network identifiers is fragile. The codebase already has `cluster_for_caip2_network` — use that or extract a shared network type.

rust/crates/kit/src/select.rs
```diff
@@ -0,0 +1,688 @@
...
+                    .rank()
+                    .cmp(&b.1.source.rank())
+                    .then(a.0.cmp(&b.0))
+            });
+        }
+    }
+}
+
+/// Decimals-adjusted amount, for comparing cost across tokens.
+fn human_amount(c: &Candidate) -> f64 {
```
EfeDurmaz16, line 322: f64 for monetary comparison is a known anti-pattern. Even with 6-dec stablecoins, amounts like 10_000.000001 cannot be represented exactly. Compare in base units with decimal adjustment.

rust/crates/kit/src/select.rs
```diff
@@ -0,0 +1,688 @@
...
+/// Resolve a currency (symbol or mint) to a mint address. A recognized symbol
+/// maps to its mint on `network`; an already-mint currency passes through;
+/// `SOL` (or anything resolving to native) yields `None`.
+fn resolve_mint(currency: &str, network: &str) -> Option<String> {
+    if currency.eq_ignore_ascii_case("SOL") {
+        return None;
+    }
+    resolve_stablecoin_mint(currency, Some(network))
+        .map(str::to_string)
+        .or_else(|| Some(currency.to_string()))
```
EfeDurmaz16, line 461: Passing through unknown strings like "USC" (typo) as mint addresses produces confusing errors. Validate base58 shape before falling through.

rust/crates/kit/src/select.rs
```diff
@@ -0,0 +1,688 @@
...
+/// non-charge, SOL-denominated, or unparseable challenges.
+fn mpp_candidate(challenge: &PaymentChallenge, order: usize) -> Option<Candidate> {
+    if !is_solana_charge_challenge(challenge) {
+        return None;
+    }
+    let request = challenge.request.decode::<ChargeRequest>().ok()?;
+    if request.currency.eq_ignore_ascii_case("SOL") {
+        return None;
+    }
+    let details: MethodDetails = request
```
EfeDurmaz16, line 402: `unwrap_or_default()` swallows serde_json parse failures. A misconfigured server would silently produce wrong defaults — log or propagate.

rust/crates/kit/src/select.rs
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    rank(&mut fundable, order);
+
+    let (_, winner, _) = fundable.into_iter().next().expect("non-empty after check");
+    Ok(winner.to_selected())
+}
+
+/// Sort fundable candidates in-place so the best option is first.
+fn rank(fundable: &mut [(usize, &Candidate, &AcceptableToken)], order: &OrderingStrategy) {
```
EfeDurmaz16, line 281: `.then(source.rank()).then(order)` duplicated across all four arms. Factor the common tiebreaker tail so the primary comparison is visible at a glance.

rust/crates/kit/src/select.rs
```diff
@@ -0,0 +1,688 @@
...
+    for requirement in x402_accepts {
+        if let Some(c) = x402_candidate(requirement, candidates.len()) {
+            candidates.push(c);
+        }
+    }
+    select_from_candidates(candidates, funded, order)
+}
+
+/// Filter normalized candidates to the fundable ones, rank, and return the winner.
+fn select_from_candidates(
```
EfeDurmaz16, line 250: O(|funded| × |candidates|) — fine at realistic sizes but worth documenting so future readers know it is intentional.

EfeDurmaz16 (approved):
Re-classifying to APPROVE after second look. The two MEDIUM concerns (`normalize_network` ad-hoc string matching, `human_amount` f64) are worth fixing but not blockers — the inputs to normalize_network come from `cluster_for_caip2_network` and `MethodDetails.network` which are well-structured in practice, and f64 comparison is unlikely to bite for whole-unit stablecoin amounts.

Overall: clean dedup, correct refactoring, thorough tests. Ship it.

PR #186, opened 2026-06-26
release(pay-kit): 0.1.0-rc.1

lgalabru:
First @solana/pay-kit release candidate. Bump the version to 0.1.0-rc.1 and publish release candidates under the 'rc' npm dist-tag (the npm-publish workflow now maps -rc.* → rc, other prereleases → beta, finals → latest).


.github/workflows/npm-publish.yml
```diff
@@ -286,7 +291,12 @@ jobs:
       - name: Check if prerelease
         id: prerelease
         run: |
-          if [[ "${{ steps.version.outputs.version }}" == *"-"* ]]; then
+          VERSION="${{ steps.version.outputs.version }}"
```
EfeDurmaz16, line 306: [P1] Split package selection before publishing.

This pay-kit RC path still shares the same manual dispatch with the mpp publish job. `@solana/mpp@0.6.0` already exists, so the release will attempt a duplicate publish and fail.

EfeDurmaz16 (changes requested):
## Verdict

Requesting changes. Current head is `b4a4ebe6e5d380e11b672674dc33223fc9b7545f`; GitHub CI is green and Greptile is `5/5`, but the release workflow is not safe to run.

## Findings

- [P1] The pay-kit RC path still shares one manual dispatch with the `@solana/mpp` publish job. `@solana/mpp@0.6.0` already exists on npm, while `@solana/pay-kit@0.1.0-rc.1` is not published, so a pay-kit release dispatch can fail by trying to republish mpp first.
- [P2] `.github/workflows/npm-publish.yml:380` still pushes `pay-kit-v${VERSION}` even when `publish-to-npm=false`, although the summary labels that mode as a dry run. This line is outside the changed hunk, so I am recording it here rather than as an inline comment.

## Hard Gates

- Release safety: failing. The workflow needs package selection or job-level gating before publishing.
- npm version evidence: verified. `@solana/mpp@0.6.0` exists; `@solana/pay-kit@0.1.0-rc.1` does not.
- Dist-tag behavior: verified. `0.1.0-rc.1` maps to `rc`, other prereleases map to `beta`, finals map to `latest`.
- CI: green for the current head.
- Greptile: satisfied at `5/5`.
- Protocol, Surfpool, and coverage gates: not applicable to this release-only PR.
- Idiomatic workflow practice: failing until dry runs stop mutating refs and package publish jobs are selectable.

.github/workflows/npm-publish.yml
lgalabru: Good point thanks!

EfeDurmaz16 (approved):

PR #198, opened 2026-06-29
chore(pay-kit): code-quality review fixes

amilz:
Quality fixes for the new `@solana/pay-kit` package, from a TypeScript review.

## Changes
- **Settlement fail-open + `onSettleError`** — unify the express/hono/fetch wrappers on one settle-failure contract: serve the handler response without settlement headers (the upto ceiling is already escrowed on-chain) and surface the error. Adds an optional `onSettleError` hook to `createPayKit`, defaulting to `console.warn`, so consumers can route failures to their own logger/alerting.
- **mpp handler cache-key fix** — key on every field a built handler captures (amount, description, externalId) so two gates differing only in those fields no longer share a handler (the cached closure was charging the first gate's amount).
- **Dedup** — `resolveCoin` + `requireMint` helpers and a shared `x402-shared` module (`x402PaymentHeader`, `errorMessage`); removes the copy-pasted coin/mint resolution and x402 header readers across the four adapters.
- **Docs** — document the `requirePayment` usage-gate settle contract; correct the upto comment (no in-process reconciler exists for upto; the on-chain channel timeout is the only fallback).
- **Tests** — cover the client `fetch` (passthrough, session-intent rejection, no-match) and the express buffered settle incl. the fail-open path. Coverage 61% → 72%.

No public API changes (the only behavior change is the error-path fail-open).

## Verification
typecheck · eslint · prettier · 79 tests · build + dts — all green.

### DIFF
--- a/typescript/packages/pay-kit/src/__tests__/buffered-settle.test.ts
+++ b/typescript/packages/pay-kit/src/__tests__/buffered-settle.test.ts
@@ -116,7 +116,6 @@ describe('express usage gate settle-after-response', () => {
         const response = await fetch(`${base}/summarize`, { headers: { [CREDENTIAL_HEADER]: 'present' } });
 
         expect(response.status).toBe(200);
-        expect(response.status).not.toBe(402);
         expect(response.headers.get('x-payment-response')).toBeNull();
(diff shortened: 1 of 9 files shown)


lgalabru (approved):
LGTM, thanks @amilz !

PR #200, opened 2026-06-30
feat(swift): x402 upto (payment-channel) client

EfeDurmaz16:
## Summary

Adds the x402 `upto` (payment-channel) client to the Swift SDK, completing the
`x402/upto` client cell. `upto` authorizes a maximum while the server settles
the actual usage (`actual <= max`); on Solana this is realized with an on-chain
payment channel whose deposit is the ceiling.

Wire-compatible with the Rust spine (`rust/crates/kit/src/x402/.../upto`) and the
SVM `upto` spec ([x402-foundation/x402#2697](https://github.com/x402-foundation/x402/pull/2697)),
on the wire shape introduced by #199 (`assetTransferMethod`, `facilitatorAddress`,
`facilitatorFee`, `channelProgram`, `tokenProgram`).

## Client behaviour

The client opens a channel depositing `maxAmount`, with the operator
(`extra.facilitatorAddress`) as channel payee, authorized signer, fee payer, and
rent payer. It signs only its payer slot of the `open` transaction; the operator
co-signs (fee payer + rentPayer), broadcasts, and later settles the metered
amount with a single voucher, refunding the remainder. The client never signs a
voucher and never needs SOL. The channel open reuses the existing
`PaymentChannels.buildOpenTransaction`.

New surface (mirrors the `exact` client layout):
- `Protocols/X402/Upto/UptoTypes.swift` wire types. `facilitatorFee` is omitted
  when `0`; the payload carries no `profile`/`signature`; the offered requirement
  is echoed verbatim in `accepted`.
- `Protocols/X402/Client/Upto/UptoPayment.swift` `parseUptoChallenge` /
  `parseUptoAccepts` and `buildUptoPayload` / `encodeUptoHeader` /
  `buildUptoHeader`.
- `UptoTransport.swift` the `X402UptoInterceptor` 402 retry path and the
  `PayKit.HttpClient.x402Upto` factory.

## Tests and coverage

36 adversarial unit tests: `assetTransferMethod` mismatch, missing
`facilitatorAddress` / `recentBlockhash`, fee out of range, fee `0` JSON
omission, `payTo == operator` empty recipients, `payTo != operator` split with
`bps = 10000 - fee` (decoded from the signed open tx), nonce random and
independent of the channel salt, `channelId == findChannelPda`, envelope
base64/JSON round-trip, `deposit == maxAmount`, `validAfter` default, and the
full interceptor retry. `swift test` passes 200/200; coverage on the new upto
code is 95.6% line.

## Interop

`harness/swift-x402-upto-client` drives the upto client; registered as
`swift-x402-upto` (intents `[x402-upto]`, opt-in via `X402_HARNESS_CLIENTS`). The
`swift.yml` interop leg runs it against `rust-x402-upto` on the embedded surfnet
with the canonical payment-channels program (`x402-upto-basic`,
`x402-upto-zero-actual`), mirroring the go/python upto legs.

## Read order
1. `swift/.../X402/Upto/UptoTypes.swift`, the four wire types, field-for-field against rust `schemes/upto/types.rs`
2. `swift/.../Client/Upto/UptoPayment.swift`, challenge parse, payload build, header encode
3. `swift/.../Client/Upto/UptoTransport.swift`, the 402 interceptor (now with the bounded expiry math)
4. `swift/.../Client/HeaderPairs.swift`, extracted shared helper
5. `swift/.../Client/Exact/Transport.swift`, X402 namespace refactor, behavior-neutral
6. `swift/.../PayCore/Errors.swift` onward, mechanical MppError to PayKitError rename churn, skim
7. `swift/Tests/.../X402UptoTests.swift`, decodes the actual signed transaction bytes
8. harness swift-x402-upto-client + `.github/actions/build-payment-channels` + `swift.yml`, e2e wiring

## Generated vs handwritten
No generated code in the diff. Handwritten: ~1,850 lines (types 374, payment 217, tests 838+, transport, harness, CI), plus ~200 lines of mechanical rename churn across ~30 files that can be skimmed.

## Reviewer note
Wire parity with the rust spine holds byte-for-byte: borsh open-args layout, the 14-account order, channel PDA seeds, deposit == maxAmount, payer-signed / fee-payer-empty signature slots. The 0-bps distribution entry at fee=10000 is intentional and matches the spine. The place to be skeptical was arithmetic on decoded ints, and that is where the one real bug lived.

## Fable 5 self-review
The protocol core is tight and the tests decode real transaction bytes rather than trusting the builder. But the first cut did `now + maxTimeoutSeconds` unchecked on a server-controlled Int in the one path every payment traverses; Swift `+` traps, so a hostile 402 could crash the paying app. Exactly the class of bug this SDK's threat model exists for, and three review rounds missed it. Fixed in a3099e51 with a range guard and boundary tests (202 tests green). Bundling the ~30-file error rename into this PR inflated the review surface for zero wire benefit; next time that ships as a separate stack layer.


swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoTransport.swift
```diff
@@ -0,0 +1,76 @@
...
+    private static func allHeaders(from http: HTTPURLResponse) -> [(name: String, value: String)] {
+        var result: [(name: String, value: String)] = []
+        for (rawKey, _) in http.allHeaderFields {
+            guard let key = rawKey as? String,
+                  let value = http.value(forHTTPHeaderField: key) else { continue }
+            result.append((name: key, value: value))
+        }
+        return result
+    }
+}
```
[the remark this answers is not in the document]
EfeDurmaz16: Consolidated both interceptors onto a shared `HTTPURLResponse.headerPairs()` extension so the exact and upto paths stay in sync. Fixed in de35d71f.

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoPayment.swift
```diff
@@ -0,0 +1,220 @@
+import Foundation
+
+// MARK: - x402 upto challenge parsing and payment building (payment-channel)
+
+/// Canonical v2 server challenge header name (lower-cased compare).
+let X402UptoPaymentRequiredHeaderName = "PAYMENT-REQUIRED"
```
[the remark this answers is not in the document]
EfeDurmaz16: Dropped the duplicate constant; the upto parser now uses the existing `X402PaymentRequiredHeaderName`. Fixed in de35d71f.

EfeDurmaz16:
On the facilitatorFee==10000 / non-operator payTo case: this is intentional parity, not a gap. The SVM spec says encode payTo at `pay_to_bps = 10000 - facilitatorFee` whenever payTo differs from the operator, and the canonical Rust client (plus Go and Python) emit the same 0-bps payTo entry. Eliding it client-side would diverge from the spine on the wire while the happy-path harness would not catch it. I added a test pinning the behavior (8f8d92ab). Whether the program should reject a 0-bps recipient is a spine + on-chain-program question rather than a Swift-only change, so I have left it matching the reference.

EfeDurmaz16:
Confirmed against the server: eliding the 0-bps entry would break verification, not fix it. The rust upto server derives its expected distribution as `Distribution { bps: 10_000 - facilitatorFee }` (server/upto.rs distribution()), so at facilitatorFee==10000 the server itself expects a 0-bps payTo entry, and verify_open requires the channel to commit to exactly that distribution_hash. If the client dropped the entry, the hashes would mismatch and the open would be rejected. The Swift client matches the server (and the rust/go/python clients) on purpose; the new test pins it. This is a verified false positive.

.github/workflows/swift.yml
```diff
@@ -88,3 +88,73 @@ jobs:
...
+          cargo-cache-key: swift-x402-upto
+      # x402 upto opens a real on-chain payment channel, so the embedded
+      # surfnet needs the canonical payment-channels program deployed. Mirrors
+      # the harness-go / harness-python upto legs: build the localnet SBF
+      # artifact (the mainnet deployment uses BPF instructions Surfnet cannot
+      # execute) and hand it to the e2e harness via PAYMENT_CHANNELS_PROGRAM_SO.
+      - name: Checkout payment channels program
+        uses: actions/checkout@v5
+        with:
+          repository: Moonsong-Labs/solana-payment-channels
```
lgalabru: repo is `solana-foundation/payment-channels`

.github/workflows/swift.yml
```diff
@@ -88,3 +88,73 @@ jobs:
...
+            exit 1
+          fi
+          perl -0pi -e 's/const TREASURY_OWNER_SENTINEL: \[u8; 32\] = \[\n.*?\n\];/const TREASURY_OWNER_SENTINEL: [u8; 32] = [\n    0xb0, 0x41, 0xd9, 0xd3, 0x37, 0xb7, 0x21, 0xbe, 0x57, 0x89, 0x4e, 0xb6, 0x9c, 0x3b, 0x68, 0x09,\n    0xa5, 0x3a, 0x0e, 0x2b, 0x6a, 0x23, 0x99, 0xfc, 0x7d, 0x5b, 0x7e, 0xda, 0x8c, 0xac, 0x89, 0xaa,\n];/s' "$sentinel_file"
+          if ! grep -Fq "$sentinel_prefix" "$sentinel_file" || ! grep -Fq "$sentinel_suffix" "$sentinel_file"; then
+            echo "::error::payment-channels treasury sentinel patch did not apply"
+            exit 1
+          fi
+      - name: Build payment channels program
+        working-directory: _payment-channels/program/payment_channels
+        run: cargo build-sbf --tools-version v1.52 --arch v1
```
lgalabru: I don't understand what's going on here. can we just use surfpool / surfnet sdk?

harness/swift-x402-upto-client/Sources/SwiftX402UptoClient/main.swift
```diff
@@ -0,0 +1,121 @@
...
+///   X402_HARNESS_NETWORK           - optional CAIP-2 / slug (diagnostics only)
+///
+/// Flow: GET target -> parse upto challenge -> buildUptoHeader (expiresAt =
+/// now + maxTimeoutSeconds) -> GET with Payment-Signature -> print ONE result
+/// JSON line on stdout. Diagnostics go to stderr.
+
+struct HarnessError: Error { let message: String }
+
+let settlementHeaderName = "x-payment-settlement-signature"
+let paymentSignatureHeaderName = "Payment-Signature"
```
lgalabru: don't we have these constants in SolanaPayKit?

harness/swift-x402-upto-client/Package.swift
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    targets: [
+        .executableTarget(
+            name: "SwiftX402UptoClient",
+            dependencies: [
+                .product(name: "SolanaPayKit", package: "SolanaPayKit"),
+            ]
+        ),
+    ]
+)
```
lgalabru, line 21: can we keep the previous screenshot?

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoPayment.swift
```diff
@@ -0,0 +1,217 @@
...
+    signer: any SolanaSigner,
+    requirements: X402UptoRequirements,
+    expiresAt: Int,
+    nonce: String? = nil,
+    nonceGenerator: (() -> Data)? = nil,
+    salt: UInt64? = nil
+) async throws -> X402UptoPayload {
+    let extra = requirements.extra
+    guard extra.assetTransferMethod == X402UptoAssetTransferMethod else {
+        throw MppError.invalidTransaction(
```
lgalabru: MppError in x402 ?

lgalabru: PayKitError maybe?

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoTransport.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// A client that drives x402 `upto` (payment-channel) endpoints: on a 402,
+    /// parse the `upto` challenge, build a partially-signed channel `open`
+    /// through `signer`, and replay with a `Payment-Signature` header.
+    static func x402Upto(
+        signer: any SolanaSigner,
+        urlSession: URLSession = .shared,
+        settlementHeader: String = "x-payment-settlement-signature"
+    ) -> PayKit.HttpClient {
+        PayKit.HttpClient(
+            interceptor: X402UptoInterceptor(signer: signer),
```
lgalabru: Is there a way to package these interceptor per protocol? `X402.Upto.Interceptor` or `X402.UptoInterceptor`

swift/README.md
```diff
@@ -130,7 +130,7 @@ Rust, Go, PHP, Ruby, Lua, and Python packages.
 | Intent | Client |
 |---|:---:|
 | `x402/exact` | pass |
-| `x402/upto` | --- |
+| `x402/upto` | pass |
```
lgalabru: can you make these matrices consistent with others (using ✅)

.github/workflows/swift.yml
EfeDurmaz16: Done — the CI upto leg now checks out solana-foundation/payment-channels. Fixed in e41842c1.

harness/swift-x402-upto-client/Sources/SwiftX402UptoClient/main.swift
EfeDurmaz16: Done — the adapter reuses the SolanaPayKit headers now (the x402Upto default settlement header + X402PaymentHeader) instead of redefining them. Fixed in e41842c1.

swift/README.md
EfeDurmaz16: Done — both swift README matrices use the shared checkmark style now. Fixed in e41842c1.

.github/workflows/swift.yml
EfeDurmaz16: This mirrors the existing go/python upto CI legs: the embedded surfnet needs the payment-channels program as a localnet SBF artifact because the mainnet deployment uses BPF instructions surfnet cannot execute, so the leg builds it from the pinned source, patches the localnet treasury sentinel, and hands the .so to the surfnet SDK via PAYMENT_CHANNELS_PROGRAM_SO. Happy to extract the repeated block into one shared composite action across go/python/swift/kotlin as a follow-up if you prefer.

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoPayment.swift
EfeDurmaz16: MppError is currently the SDK's single, protocol-agnostic error type (it lives in PayCore despite the name), and the exact x402 client throws it too — so upto matches exact here. Renaming it to PayKitError is a clean SDK-wide change; happy to do it as a focused follow-up so it stays out of the upto diff. Your call.

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoTransport.swift
EfeDurmaz16: The exact client ships a flat X402Interceptor, so X402UptoInterceptor matches it. Namespacing to X402.UptoInterceptor reads better but touches the exact interceptor too; happy to do it SDK-wide as a follow-up. Your call.

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoPayment.swift
EfeDurmaz16: Done — renamed MppError to PayKitError SDK-wide; it is the single protocol-agnostic error type (lives in PayCore, thrown by mpp + x402 exact + upto). 596c8d3c.

EfeDurmaz16: Done — see above: renamed to PayKitError SDK-wide in 596c8d3c.

swift/Sources/SolanaPayKit/Protocols/X402/Client/Upto/UptoTransport.swift
EfeDurmaz16: Done — the x402 interceptors are namespaced now: X402.Interceptor (exact) and X402.UptoInterceptor (upto). 596c8d3c.

harness/swift-x402-upto-client/Package.swift
EfeDurmaz16: Done — restored the previous PayKitDemo screenshot (and plain caption). 58455bad.

.github/workflows/swift.yml
EfeDurmaz16: Tried the surfnet-sdk streaming path (drop PAYMENT_CHANNELS_PROGRAM_SO so the harness streams the program from the datasource): 2/4 upto scenarios fail with 402, because the streamed hosted deployment does not carry the localnet treasury owner the harness pins via the sentinel patch. So the localnet build is required. To remove the duplication instead, I extracted the checkout + sentinel-patch + SBF build into a shared composite action (.github/actions/build-payment-channels) that swift.yml and kotlin.yml now call in a single step. Verified locally: swift, kotlin, go, python, and rust upto interop all pass against the rust upto server with the action's artifact. go/python can adopt the same action in a follow-up. 2f5853b7.

harness/swift-x402-upto-client/Sources/SwiftX402UptoClient/main.swift
```diff
@@ -0,0 +1,119 @@
...
+    headers: [String: String],
+    body: Any,
+    settlement: String?,
+    error: String? = nil
+) {
+    var payload: [String: Any] = [
+        "type": "result",
+        "implementation": "swift",
+        "role": "client",
+        "ok": ok,
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed — an absent settlement now serializes as NSNull (JSON null) instead of a wrapped nil, so the result line is always valid JSON. 0b2157d0.

EfeDurmaz16:
@greptile on the suggested `typealias X402Interceptor = X402.Interceptor`: this SDK is pre-1.0 (0.6.x) with no stable published consumers to keep source-compatible, and the namespacing (`X402.Interceptor` / `X402.UptoInterceptor`) was an explicit maintainer request — the rename is the intended change, not an accidental break. The repo also deliberately avoids backward-compat shims. A compat typealias would keep the old flat name alongside the namespaced one, defeating the namespacing intent, so I am skipping it. Is there a concrete backward-compat requirement that would make it necessary regardless?

swift/Examples/PayKitDemo/PayKitDemo/ContentView.swift
```diff
@@ -419,6 +457,10 @@ struct Endpoint: Identifiable, Hashable {
     /// Discovery intent of the first offer (`charge` / `session` / …); the demo
     /// only settles `charge` over MPP and explains the rest.
     let intent: String
```
lgalabru: Can we use enums as much as possible?

swift/Examples/PayKitDemo/PayKitDemo/ContentView.swift
```diff
@@ -419,6 +457,10 @@ struct Endpoint: Identifiable, Hashable {
...
+    /// Scheme of the first offer (`exact` / `upto` / …) when present. A metered
+    /// `upto` route advertises the generic `charge` intent, so the demo routes by
+    /// this scheme to reach the usage flow.
+    let scheme: String?
```
lgalabru: Can we use enums as much as possible?

swift/Sources/SolanaPayKit/PayCore/RpcClient.swift
lgalabru, line 50: I'm confused. why do we need an RpcClient, client side?

swift/Examples/PayKitDemo/PayKitDemo/ContentView.swift
EfeDurmaz16: Done — the demo models intent as an EndpointIntent enum (charge/session/subscription/.other) and routes by matching on it. ed322293.

swift/Examples/PayKitDemo/PayKitDemo/ContentView.swift
EfeDurmaz16: Done — scheme is now an EndpointScheme enum (exact/upto/.other), so the upto route is matched as `.upto`. ed322293.

swift/Sources/SolanaPayKit/PayCore/RpcClient.swift
EfeDurmaz16: The RpcClient is only used by the MPP charge + x402 exact clients (getAccountOwner resolves a mint's token program when the server omits it, and blockhash fetch); the upto client needs neither, so it is not on the upto path. Pre-existing, not added by this PR.

lgalabru (approved):

PR #201, opened 2026-06-30
feat(kotlin): x402 upto (payment-channel) client

EfeDurmaz16:
## Summary

Adds the x402 `upto` (payment-channel) client to the Kotlin SDK, completing the
`x402/upto` client cell. `upto` authorizes a maximum while the server settles
the actual usage (`actual <= max`); on Solana this is realized with an on-chain
payment channel whose deposit is the ceiling.

Wire-compatible with the Rust spine (`rust/crates/kit/src/x402/.../upto`) and the
SVM `upto` spec ([x402-foundation/x402#2697](https://github.com/x402-foundation/x402/pull/2697)),
on the wire shape introduced by #199 (`assetTransferMethod`, `facilitatorAddress`,
`facilitatorFee`, `channelProgram`, `tokenProgram`).

## Client behaviour

The client opens a channel depositing `maxAmount`, with the operator
(`extra.facilitatorAddress`) as channel payee, authorized signer, fee payer, and
rent payer. It signs only its payer slot of the `open` transaction; the operator
co-signs (fee payer + rentPayer), broadcasts, and later settles the metered
amount with a single voucher, refunding the remainder. The client never signs a
voucher and never needs SOL. The channel open reuses the existing
`PaymentChannels.buildOpenTransaction`.

New surface (mirrors the `exact` client layout):
- `protocols/x402/upto/Types.kt` wire types. `facilitatorFee` is omitted when
  `0`; the payload carries no `profile`/`signature`; the offered requirement is
  echoed verbatim in `accepted` via `raw`.
- `protocols/x402/client/upto/Payment.kt` `parseUptoChallenge` /
  `parseUptoAccepts` and `buildUptoPayload` / `encodeUptoHeader` /
  `buildUptoHeader`.

## Tests and coverage

Adversarial unit tests: `assetTransferMethod` mismatch, missing
`facilitatorAddress` / `recentBlockhash`, fee out of range, fee `0` JSON
omission, `payTo == operator` empty recipients, `payTo != operator` split with
`bps = 10000 - fee`, nonce random and independent of the channel salt,
`channelId == findChannelPda`, envelope base64/JSON round-trip,
`deposit == maxAmount`, and `validAfter` default. `gradle check` passes; the
jacoco 90% line gate holds (98.8% on the new upto code).

## Interop

`harness/kotlin-x402-upto-client` drives the upto client; registered as
`kotlin-x402-upto` (intents `[x402-upto]`, opt-in via `X402_HARNESS_CLIENTS`).
The `kotlin.yml` interop leg runs it against `rust-x402-upto` on the embedded
surfnet with the canonical payment-channels program (`x402-upto-basic`,
`x402-upto-zero-actual`), mirroring the go/python upto legs.

## Read order
1. `kotlin/.../protocols/x402/upto/Types.kt`, wire shapes and serde-tolerance annotations
2. `kotlin/.../client/upto/Payment.kt`, challenge precedence and the open-transaction builder, the heart
3. `kotlin/.../client/upto/UptoPaymentTest.kt`, byte-level instruction assertions and adversarial parse cases
4. `harness/kotlin-x402-upto-client/.../Main.kt`, the e2e adapter and its env contract
5. `harness/src/implementations.ts` + gradle plumbing, opt-in registration
6. `.github/actions/build-payment-channels` + `kotlin.yml`, SBF build + upto harness job
7. `kotlin/examples/AndroidDemo/.../MainActivity.kt` + `OpenApi.kt`, demo scheme-first routing
8. READMEs + screenshot, docs

## Generated vs handwritten
Only the demo screenshot is an asset; ~1,390 added lines are handwritten (Payment.kt 274, Types.kt 139, tests 537, harness 166, CI/gradle ~160, demo ~85).

## Reviewer note
Every byte-level invariant checks out against the rust spine: borsh layout, 14-account codama order with writable/signer flags, PDA seeds, deposit == maxAmount, zero-filled unsigned slots, salt independent of nonce, the intentional 0-bps entry at fee=10000. Header-precedence semantics match the rust `or_else` chain case by case. The upto harness job runs green against the rust upto server.

## Fable 5 self-review
The SDK layer asserts parity at the instruction-byte level instead of hand-waving it, which is the right bar. The Android demo was written faster than the SDK: it logged the raw settlement envelope instead of running it through the `signatureFromReceiptHeader` helper three functions below, classic copy-paste drift from the exact path; fixed in 0cab25b1. Two open polish items I would still flag: the harness adapter hardcodes the settlement header where rust reads `X402_HARNESS_SETTLEMENT_HEADER`, and the fee-range check sits outside the split branch where rust has it inside, stricter but a parity wrinkle.


EfeDurmaz16:
Addressed the settlement-response finding: `UptoSettlementResponse.transaction` is now nullable (`String? = null`) to match the exact settlement type, so decoding a failure body that omits it no longer throws `SerializationException`. Added a failure-path test covering it. Fixed in 04c01404.

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
```diff
@@ -0,0 +1,139 @@
...
+    val errorReason: String? = null,
+    /** Payer wallet (base58) the operator settled against. */
+    val payer: String? = null,
+    /** Settlement transaction signature; absent on a failure response. */
+    val transaction: String? = null,
+    /** CAIP-2 network identifier. */
+    val network: String,
+    /** Actual base units charged (may be ``0``). */
+    val amount: String,
+)
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed: network and amount are now nullable (`String? = null`) to match the exact settlement type, so a generic failure body that omits them no longer throws MissingFieldException. Extended the failure-body test to a response carrying only success + errorReason. Fixed in 93093bdf.

kotlin/examples/AndroidDemo/app/src/main/java/com/solana/paykit/demo/MainActivity.kt
```diff
@@ -312,7 +314,17 @@ private fun DemoScreen() {
                             onSelectProtocol = { protocolChoice = protocolChoice + (endpoint.id to it) },
                             onClick = onClick@{
                                 val s = signer ?: return@onClick
-                                when (endpoint.intent) {
+                                // A metered x402 upto route advertises the generic
+                                // `charge` intent, so route by scheme first: authorize a
+                                // ceiling; the server meters + settles actual <= max.
+                                if (endpoint.scheme == "upto") {
```
lgalabru: does kotlin have enum? if not can we match against a constant?

EfeDurmaz16: Kotlin has enums, but matching the wire string against the exported UPTO_SCHEME constant is the cleanest fit here. Done in 0656ed57.

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/client/upto/Payment.kt
```diff
@@ -0,0 +1,274 @@
...
+ * client never signs a voucher and never needs SOL.
+ *
+ * The envelope is the canonical x402 v2 shape ``{ x402Version, accepted,
+ * payload }``: scheme and network live inside ``accepted``, not at the envelope
+ * level. It is encoded as compact JSON then standard base64 (with padding) and
+ * carried in the ``Payment-Signature`` header.
+ */
+
+/** x402 protocol version stamped in the envelope. INVARIANT: 2. */
+private const val X402_VERSION = 2
```
lgalabru: we may have that constant somewhere else already?

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
```diff
@@ -0,0 +1,139 @@
...
+
+/** An ``upto`` payment requirement (one ``accepts[]`` entry in a 402 challenge). */
+@Serializable
+data class UptoRequirements(
+    /** Scheme identifier; ``upto`` for this requirement. Always emitted so the
+     *  server reads it from the echoed ``accepted`` object. */
+    @EncodeDefault(EncodeDefault.Mode.ALWAYS)
+    val scheme: String = UPTO_SCHEME,
+    /** CAIP-2 network identifier. */
+    val network: String,
```
lgalabru: enum?

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
```diff
@@ -0,0 +1,139 @@
...
+    /** Whether settlement succeeded. */
+    val success: Boolean,
+    /** Reason for failure when ``success`` is false. */
+    val errorReason: String? = null,
+    /** Payer wallet (base58) the operator settled against. */
+    val payer: String? = null,
+    /** Settlement transaction signature; absent on a failure response. */
+    val transaction: String? = null,
+    /** CAIP-2 network identifier; absent on a generic failure response. */
+    val network: String? = null,
```
lgalabru: enum?

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/client/upto/Payment.kt
EfeDurmaz16: Good catch — hoisted it: both the exact and upto builders now share one internal X402_VERSION in the x402 package. 7fc79c9a.

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
EfeDurmaz16: network is a CAIP-2 identifier kept as String to match the rust spine (`pub network: String`), the exact client, and swift, and to stay robust to networks the client doesn't know; the typed CAIP-2 constants live in the Network object. Happy to introduce a wire-compatible sealed Network type (Mainnet/Devnet/Testnet/Other(caip2)) SDK-wide as a follow-up so the typed surface lands consistently rather than upto-only — want me to?

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
EfeDurmaz16: Same as the requirements.network field: kept as String for rust-spine + exact + swift parity and unknown-network robustness. The sealed Network type above would cover this field too. Let me know if you'd like that follow-up.

EfeDurmaz16:
Addressed the consumeUpto finding: the demo now falls back to 300s when maxTimeoutSeconds is non-positive (matching the harness adapter), so the authorization can no longer expire immediately. 1596dc3d.

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
EfeDurmaz16: Done — network is now a sealed SolanaNetwork in paycore (Mainnet/Devnet/Testnet + Other preserving unrecognized CAIP-2 ids verbatim) whose serializer emits the bare CAIP-2 string, so the wire is byte-identical and an unknown network still parses; verified against the rust upto server on-chain. a0e7d3ec.

kotlin/src/main/kotlin/com/solana/paykit/protocols/x402/upto/Types.kt
EfeDurmaz16: Done — same sealed SolanaNetwork type on the settlement field (nullable for failure responses). a0e7d3ec.

lgalabru (approved):

PR #203, opened 2026-06-30
fix(sdk): recover x402 upto follow-ups

lgalabru:
## Summary
- recover the remaining TypeScript x402 upto follow-up diff from `feat/blockhash-cache-main`
- align TypeScript and Python settle_and_finalize builders with the deployed payment-channel account order
- advance the vendored x402 submodule to the pushed SVM upto account-order fix

## Test Plan
- x402: `pnpm test -- test/unit/upto.test.ts`
- x402: `pnpm exec prettier -c .prettierrc --check src/payment-channels/onchain.ts test/unit/upto.test.ts`
- x402: `pnpm exec eslint src/payment-channels/onchain.ts test/unit/upto.test.ts --ext .ts`
- pay-kit: `PYTHONPATH=src /Users/ludo/Coding/pay-kit/python/.venv/bin/python -m pytest tests/test_paymentchannels.py -q`
- pay-kit: `vitest run packages/mpp/src/__tests__/session-on-chain.test.ts`
- `git diff --check`

Stacked on #202.


typescript/packages/pay-kit/src/adapters/mpp.ts
```diff
@@ -52,22 +51,17 @@ export function createMppAdapter(config: PayKitConfig): ProtocolAdapter {
...
-            gate.kind,
-            gate.payTo,
-            mint,
-            splits,
-            gate.subscription ?? null,
-            totalAmount(gate).toString(),
-            gate.description ?? null,
-            gate.externalId ?? null,
-        ]);
+        const key = JSON.stringify([gate.kind, gate.payTo, mint, splits, gate.subscription ?? null]);
```
EfeDurmaz16, line 64: **BLOCKER (money path):** narrowing the key to `[kind, payTo, mint, splits, subscription]` drops `amount`/`description`/`externalId`, but the built handler closes over `gate` and reads the price per-request via `optionsFor(gate)`/`totalAmount(gate)` (l114). `handlers` is a Map at adapter scope (l52), shared across all gates (createMppAdapter is per-protocol, paykit.ts:214). So two fixed same-`payTo` same-coin no-fee no-subscription gates (e.g. a $1 basic and a $5 premium USDC route to one wallet) collide on this key: the second reuses the first's closure and issues/settles the first's amount, while `acceptsEntry` advertises the correct one — a divergent, wrong on-chain charge amount. The comment you removed ("Key on every field a built handler captures…") flagged exactly this. Restore the three fields to the key, or compute options from the current gate at call time.

EfeDurmaz16 (commented):
## Review summary — adversarial line-by-line pass

**Verdict: not approve-ready — 1 blocker on the money path.** The settle-order alignment, the x402 recovery, the submodule advance, and the deletions are otherwise sound.

- **Settle account order is correct.** TS and Python `settle_and_finalize` now agree with each other (channel writable → merchant readonly-signer → instructions sysvar) and match the canonical Rust spine `build_settle_and_finalize_instructions`, which deliberately overrides the stale codama merchant-first order because the deployed program expects channel first (else `InvalidAccountOwner`).
- **Deletions leave no dangling imports.** `coin.ts` and `x402-shared.ts` are fully removed with zero remaining importers; the removed test files are self-contained.

**Blocker**
- `typescript/packages/pay-kit/src/adapters/mpp.ts:64` — the handler cache key was narrowed to `[kind, payTo, mint, splits, subscription]`, dropping `amount`/`description`/`externalId`, but the built handler closes over its `gate` and reads price per-request via `optionsFor(gate)`/`totalAmount(gate)` (l114). `handlers` is a Map at adapter scope (l52) shared across all gates (createMppAdapter is per-protocol, paykit.ts:214). Two fixed same-`payTo` same-coin no-fee no-subscription gates (e.g. a $1 and a $5 USDC route to one wallet) collide on this key, so the second reuses the first's closure and issues/settles the first's amount while `acceptsEntry` advertises the correct one — a wrong on-chain charge amount. The removed comment ("Key on every field a built handler captures…") documented exactly this. Restore `amount`/`description`/`externalId` to the key, or compute options from the current gate at call time.

**Follow-ups (non-blocking)**
- `typescript/packages/pay-kit/src/paykit.ts` (~822) — the post-handler settle-failure path was refactored (the `onSettleError`/replay machinery removed) and its only test, `buffered-settle.test.ts` (-139), was deleted, so both the metered settle-success path and the new settle-failure semantics now ship with no coverage. Add a usage-gate express test for settle success (200 + settlement headers) and failure.
- `typescript/packages/pay-kit/src/client/index.ts` — `client.test.ts` (-65) was removed but `createPayKitClient` (non-402 passthrough, session-intent guard, unmatched-protocol passthrough) still ships live, so labelling it dead code is inaccurate; restore the regression coverage.

PR #209, opened 2026-07-02
feat(python): add route metadata paywall config

lgalabru:
## Summary
- Adds `PayConfig` as an app-level Python SDK config object for pay-gated route surfaces.
- Adds FastAPI route-metadata paywall helpers so apps can gate routes by decorators/tags/default policy without duplicating endpoint paths.
- Adds `PayConfig.from_sources(...)` and `install_paywall(..., env_prefix=...)` so host apps can pass config mappings and env prefixes through the SDK.
- Bumps the Python package minor version to `0.3.0`.
- Documents the FastAPI paywall pattern and adds focused Python tests.

## Test Plan
- `uv run pyright src/solana_pay_kit/config.py src/solana_pay_kit/fastapi.py tests/test_pk_config.py tests/test_pk_fastapi_paywall.py`
- `uv run ruff check src/solana_pay_kit/config.py src/solana_pay_kit/fastapi.py src/solana_pay_kit/__init__.py tests/test_pk_config.py tests/test_pk_fastapi_paywall.py`
- `uv run pytest tests/test_pk_config.py tests/test_pk_fastapi_paywall.py`
- `uv lock`


EfeDurmaz16 (commented):
Two focused findings from review.

python/src/solana_pay_kit/fastapi.py
```diff
@@ -364,6 +544,70 @@ async def _paykit_usage_settle(  # pyright: ignore[reportUnusedFunction]  # regi
...
+    routes = getattr(app, "routes", ())
+    for route in routes:
+        matcher = getattr(route, "matches", None)
+        if not callable(matcher):
+            continue
+        try:
+            match, _child_scope = cast("tuple[Match, dict[str, Any]]", matcher(request.scope))
+        except Exception:  # pragma: no cover - defensive for third-party routes
+            continue
+        if match is Match.FULL:
```
EfeDurmaz16, line 560: This returns only the top-level matched route. For `app.mount("/api", sub)`, Starlette returns the `Mount` with no endpoint, so `@pay_required()` on the child route is never seen and the route can pass through unpaid.

python/src/solana_pay_kit/fastapi.py
```diff
@@ -243,6 +318,111 @@ def _usage_challenge(
...
+
+    pay_config = _PayConfig.from_sources(config, env_prefix=env_prefix)
+    if not pay_config.enabled:
+        return
+
+    install_paywall_from_config(
+        app,
+        PaywallConfig(
+            gate_ref=pay_config.gate_ref(),
+            config=pay_config.configure(),
```
EfeDurmaz16: `pay_config.configure()` mutates the module-level singleton, so `install_paywall()` after an explicit `configure(mpp=..., x402=...)` drops those sub-configs globally. Consider building a `Config` without replacing `_config`, or making this helper the sole supported config entry point.

python/src/solana_pay_kit/fastapi.py
EfeDurmaz16: Still reproduces on c7478ee. TS keeps the gate on the route stack (and openapiFromExpress walks mounted Express stacks), while Rust attaches paid_get directly to the route, so this global FastAPI scanner should either descend into mounted subapps or document mounted apps as out of scope.

lgalabru:
Addressed the Greptile findings and the CI failure.

- Removed the redundant FastAPI paywall settlement-header write and now rely on the existing settlement-header middleware.
- Fixed `{prefix}NO_PREFLIGHT=false` so it explicitly re-enables preflight.
- Changed `install_paywall()` to build a middleware-local config without replacing the global singleton, while preserving existing MPP/X402 sub-configs.
- Added regression coverage for the env override and global-config preservation cases.
- Updated the Python README/docs/example/snippet for the route-metadata paywall flow.
- Fixed Ruby CI by removing the vulnerable/unpatched `webrick` dependency and updating `json` to `2.20.0`.

Local checks:

- `uv run --extra dev --extra fastapi --extra flask --extra django ruff check src tests examples/fastapi/app.py`
- `uv run --extra dev --extra fastapi --extra flask --extra django pyright`
- `uv run --extra dev --extra fastapi --extra flask --extra django pytest --cov=solana_pay_kit --cov-report=term-missing --cov-report=json:coverage.json --cov-fail-under=90 --ignore=tests/test_server_html.py`
- `uv run --extra dev --extra fastapi python -m py_compile examples/fastapi/app.py docs/snippets/paywall.server.py`
- `mise exec ruby@3.4.9 -- bundle exec bundle-audit check --update`
- `mise exec ruby@3.4.9 -- env COVERAGE=1 bundle exec ruby -Itest test/run.rb`

Docs sync:

- `PAY_KIT_DIR=/private/tmp/pay-kit-pr209-finish node apps/docs/scripts/gen-sdk-snippets.mjs`
- `pnpm --filter @pay/docs check-types`
- targeted Prettier check on changed pay-web-ui docs/snippet files

CI is green on the latest head commit.

python/src/solana_pay_kit/fastapi.py
```diff
@@ -364,6 +541,70 @@ async def _paykit_usage_settle(  # pyright: ignore[reportUnusedFunction]  # regi
...
+    routes = getattr(app, "routes", ())
+    for route in routes:
+        matcher = getattr(route, "matches", None)
+        if not callable(matcher):
+            continue
+        try:
+            match, _child_scope = cast("tuple[Match, dict[str, Any]]", matcher(request.scope))
+        except Exception:  # pragma: no cover - defensive for third-party routes
+            continue
+        if match is Match.FULL:
```
EfeDurmaz16, line 560: Mounted sub-app requests match the parent `Mount` here, so child endpoint metadata/tags are never seen. I reproduced `/api/paid` returning 200 with no `PayCore.process` call under `default_policy="public"`.

EfeDurmaz16 (changes requested):
I re-reviewed the latest head; the global-config and env/preflight fixes look good, and the focused paywall tests pass. One mounted FastAPI sub-app shape still bypasses route metadata because the parent app scan stops at the `Mount`, so child `@pay_required` endpoints are treated as public. Since mounted apps/routers are a common FastAPI composition pattern, I think this should be fixed before merge.

EfeDurmaz16 (approved):
Re-reviewed current head 9307ea1. The FastAPI paywall now recurses into mounted apps and the targeted Python paywall/config suite passes locally (52 tests), including the mounted child-route regression. I also compared the behavior against the TS/Rust reference spine: TS recovers gate metadata from the mounted Express route stack, Rust binds gates directly with paid_get/paid_post, and the Python behavior no longer contradicts that model. Approving; at review time GitHub showed 43 completed checks green with only the Lua job still in progress.

PR #217, opened 2026-07-07
feat: migrate all SDKs to epoch-addressed payment channels

lgalabru:
## Summary
- Vendors the epoch-addressed payment-channels program (upstream Moonsong-Labs/solana-payment-channels@0c07d575, PR #77): `settleAndSeal`/`seal` renames, new permissionless `reclaim`, `openSlot` as an `openArgs` field and channel PDA seed, 2-byte voucher magic `[0x56, 0x01]`, `sealed`/`distributed` statuses, `closedChannel` removed — regenerated clients plus the hand-written cascade in Go, Rust, TypeScript, Python, Kotlin, and Swift (one commit per language).
- Wire conventions shared across SDKs: the challenge (`SessionRequest` and x402 upto `extra`) carries a server-fetched `recentSlot` (u64-as-string, taken from the `getLatestBlockhash` response context — no extra `getSlot` round-trip); the session `OpenPayload` echoes it; the voucher JSON is unchanged (magic lives only in the signed bytes); durable channel state persists `open_slot` and renames `finalized` → `sealed`.
- Distribution hash: confirmed SHA-256 everywhere; fixed stale BLAKE3 doc comments (Kotlin/Swift/README) and dropped the unused `blake3` dependency from the Rust kit.
- Python `just` gates now run through `uv` so `just test`/`lint`/`typecheck`/`fmt` work without a manually activated venv.

## Test Plan
- Per-language unit suites are green (Python: 1313 passed, 1 skipped — the Surfpool-gated e2e; run it with `MPP_RUN_SURFNET_E2E=1` against the sandbox).
- Python `just lint` / `just typecheck` (pyright: 0 errors) / `just fmt` clean.
- Cross-language: run the conformance harness against a validator with the new program deployed (`harness/vectors/session-voucher.json` updated for the 50-byte magic-prefixed preimage).

## Breaking Changes
- Pre-1.0 API renames in every SDK: `finalize`→`seal` (builders, `mark_sealed`, store field `sealed`, voucher reject tag `channel-sealed`), `find_channel_pda`/open params now require `open_slot`, voucher preimage is 50 bytes with a magic prefix, and channel PDAs are per-incarnation (same params at a different slot derive a different address). Servers must reissue challenges (`recentSlot`) and clients must echo them; old open payloads without `recentSlot` cannot be re-derived.


typescript/packages/pay-kit/src/adapters/x402-upto.ts
```diff
@@ -256,20 +256,24 @@ export class X402Upto {
...
         const base = this.#requirements(maxPrice);
         try {
-            const { value } = await createSolanaRpc(this.#rpcUrl).getLatestBlockhash().send();
+            const { context, value } = await createSolanaRpc(this.#rpcUrl).getLatestBlockhash().send();
             return {
                 ...base,
                 extra: {
                     ...base.extra,
                     lastValidBlockHeight: value.lastValidBlockHeight.toString(),
                     recentBlockhash: value.blockhash,
```
EfeDurmaz16: This enriched requirement is only used for the `Payment-Required` header; the usage JSON body still goes through bare `accepts()`, so body-based `upto` clients miss `extra.recentSlot`. Can we expose/reuse this same requirement for both header and body offers?

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -553,8 +623,8 @@ async def _handle_open(self, payload: OpenPayload) -> str:
             session_id = payload.session_id()
             existing = await self._core.store().get_channel(session_id)
             if existing is not None:
-                if existing.finalized:
-                    raise PaymentError(f"channel {session_id} is already finalized", code="invalid-payload")
```
EfeDurmaz16, line 557: Can we thread the challenge `request.recentSlot` into `VerifyOpenTxExpected` before treating this decoded slot as authoritative? If the payload omits `recentSlot`, this accepts a transaction built against a different slot and then overwrites the payload; TS rejects that via `expected.openSlot`.

go/protocols/mpp/server/session_store.go
```diff
@@ -68,12 +68,19 @@ type ChannelState struct {
...
+
 	// Cumulative is the highest cumulative amount accepted by the server (the
 	// settled watermark).
 	Cumulative uint64 `json:"cumulative"`
 
-	// Finalized is true once the channel has been finalized on-chain.
-	Finalized bool `json:"finalized"`
+	// Sealed is true once the channel has been sealed on-chain (the program's
+	// seal instruction, formerly finalize).
+	Sealed bool `json:"sealed"`
```
EfeDurmaz16, line 85: Can we keep `finalized` as a legacy decode alias for `sealed` here, and mirror that in Python `ChannelState.from_dict`? Rust already has `serde(alias = "finalized")`; without the alias, persisted pre-rename stores reload closed channels as unsealed after upgrade.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
```diff
@@ -510,8 +522,8 @@ async def settle_and_finalize_channel(
             )
         expires_at = state.highest_voucher_expires_at
 
-    settle = build_settle_and_finalize_instructions(
-        merchant=merchant_pubkey,
+    settle = build_settle_and_seal_instructions(
+        payee=merchant_pubkey,
         channel=channel,
         authorized_signer=authorized_signer,
```
EfeDurmaz16, line 539: Just above this, can we match Go’s strict payer handling and fail when `state.operator` is missing? Falling back to `config.recipient` can derive the wrong refund ATA and only fail after the settlement transaction is built/sent.

EfeDurmaz16 (changes requested):
Thanks for taking on a wide, high-risk migration; the core cross-SDK shape looks coherent, and the focused Go/Python/TS tests I ran passed. I’m requesting changes for a few edge cases where the new slot/sealed semantics are not applied consistently across the reference spine: TS pay-kit’s usage 402 body can omit `recentSlot`, Python’s open verifier does not bind the transaction `openSlot` to the issued challenge, and Go/Python durable state should preserve legacy `finalized` decoding. I also left one Python settlement-path comment where it diverges from Go’s strict channel payer handling. Once these are fixed, I think this is close.

python/Justfile
```diff
@@ -1,31 +1,36 @@
...
 
 # Type-check with pyright.
 typecheck:
-    pyright
+    {{uv_run}} pyright
 
 # Run tests with the 90% line-coverage gate.
 test-cover:
-    pytest --cov=pay_kit --cov-report=term-missing --cov-fail-under=90 --ignore=tests/test_server_html.py
+    {{uv_run}} pytest --cov=pay_kit --cov-report=term-missing --cov-fail-under=90 --ignore=tests/test_server_html.py
```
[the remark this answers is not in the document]
lgalabru: Fixed in 60a62c2: `test-cover` now targets `--cov=solana_pay_kit`. The gate is real again — 91.87% against the 90% bar.

python/src/solana_pay_kit/_paycore/rpc.py
```diff
@@ -104,12 +119,27 @@ async def send_raw_transaction(self, raw_tx: bytes) -> Any:
...
-        return _RpcResponse(_BlockhashValue(blockhash))
+        slot = ((result or {}).get("context") or {}).get("slot") if isinstance(result, dict) else None
+        if isinstance(slot, bool) or not isinstance(slot, int) or slot < 0:
+            slot = None
+        return _RpcResponse(_BlockhashValue(blockhash), context=_RpcContext(slot))
+
+    async def get_slot(self, commitment: str = "confirmed") -> int:
+        """Fetch the current slot. Used at channel-open time: the program
+        requires ``openSlot <= clock.slot`` with a 1500-slot freshness window,
+        so clients stamp the open (and its PDA seed) with the current slot."""
```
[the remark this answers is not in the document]
lgalabru: Fixed in 60a62c2: the docstring now says servers stamp the challenge `recentSlot` (this call is the fallback when `getLatestBlockhash` lacks `context.slot`) and clients never fetch the slot themselves.

typescript/packages/pay-kit/src/adapters/x402-upto.ts
lgalabru: Fixed in b95f816: `accepts()` now returns the same server-enriched requirement (`extra.recentBlockhash` + `extra.recentSlot`), and the usage challenge builds the header and JSON body from one shared fetch (`challengeHeaders` takes the precomputed entries), so body-based upto clients get `recentSlot` without a second RPC round-trip.

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
lgalabru: Fixed in 60a62c2: `VerifyOpenTxExpected` gained `recent_slot`, threaded from the verified challenge through `_handle_open`/`_open_tx_expected`; `verify_open_tx` now rejects a transaction whose decoded `openSlot` differs from the challenge `recentSlot` even when the payload omits it — matching the TS `expected.openSlot` bind. Covered by `test_verify_open_tx_binds_challenge_recent_slot`.

go/protocols/mpp/server/session_store.go
lgalabru: Maintainer decision: no backward compatibility for pre-seal records at this point (pre-1.0, and pre-rename channels reference the old program's addressing anyway). Instead of adding aliases we went the other way in 17b194d/60a62c2: the Rust `serde(alias = "finalized")` is removed (a legacy record now fails loudly on its missing `sealed` field) and Python `ChannelState.from_dict` explicitly rejects records carrying `finalized`. Go's `ChannelState` is a plain struct — JSON decoding lives in embedder-owned durable stores, and `encoding/json` ignoring unknown fields is standard Go semantics, so no alias is added there either. Operators upgrading across the rename should drain/reset durable channel stores.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
lgalabru: Fixed in 60a62c2: `settle_and_seal_channel` now requires the recorded channel payer and raises `invalid-config` before building/broadcasting anything (no more `config.recipient` fallback deriving the wrong refund ATA), mirroring Go. Covered by `test_settle_requires_recorded_channel_payer` (asserts nothing is broadcast and the settle guard is released).

lgalabru:
@greptile-apps review please

go/protocols/mpp/server/session_store.go
```diff
@@ -68,12 +68,19 @@ type ChannelState struct {
...
```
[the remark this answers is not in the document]
lgalabru: Fixed in 4557641: `ChannelState.UnmarshalJSON` now rejects records carrying `finalized` loudly ("legacy pre-seal channel record ... migrate or reset the durable channel store") instead of silently reloading a closed channel as unsealed — no backward-compat alias, per maintainer decision, matching the Python `from_dict` guard and Rust's missing-required-field behavior. Covered by `TestChannelStateRejectsLegacyFinalizedRecord` (also asserts post-rename records still round-trip through the custom decoder).

go/paycore/paymentchannels/paymentchannels.go
```diff
@@ -32,6 +32,12 @@ const channelSeed = "channel"
 // eventAuthoritySeed is the event-authority PDA seed prefix.
 const eventAuthoritySeed = "event_authority"
 
+// voucherMagic is the constant 2-byte magic prefix leading every signed
+// voucher payload ([0x56, 0x01]). It is part of the signed bytes only, never
+// carried in wire JSON; the on-chain program rejects payloads without it
+// (voucherBadMagic).
```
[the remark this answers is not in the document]
lgalabru: Fixed in 4557641: the magic bytes are now typed constants (`voucherMagic0`/`voucherMagic1`) written directly into the preimage buffer — no mutable package-level var.

lgalabru:
@greptile-apps review please

typescript/packages/pay-kit/src/adapters/x402-upto.ts
```diff
@@ -256,20 +268,24 @@ export class X402Upto {
...
+                    recentSlot: context.slot.toString(),
                 },
             };
         } catch {
```
EfeDurmaz16: If the lifetime fetch fails, this fallback still advertises an `upto` offer without `recentSlot`/`recentBlockhash`. Because clients now require `recentSlot` to build `openSlot`, fail or omit `upto` rather than returning an unusable challenge.

EfeDurmaz16 (changes requested):
I re-reviewed the latest head and several prior issues are fixed: body/header `recentSlot` sharing on the happy path, legacy `finalized` rejection, and strict settlement payer handling. I still see the x402 `upto` open verifier accepting an open transaction without decoding `OpenArgs.openSlot` and comparing it to the challenged `extra.recentSlot`, which leaves the new slot-addressed channel invariant unenforced. There is also a smaller fallback path that can advertise a bare, unusable `upto` offer when recent lifetime fetch fails. Requesting changes for the openSlot binding; the rest of the architecture looks much closer.

EfeDurmaz16 (dismissed):
I re-reviewed the latest head again, including the previous request-changes items. The Python open verifier now binds the decoded transaction openSlot to the challenge recentSlot, payer/refund handling is strict, and the sealed/finalized migration now fails loudly where intended. The remaining x402 recent-state fallback is consistent with Python's non-fatal challenge-time behavior and is not a merge blocker for me; CI is green across the matrix and the focused Go/Python checks I ran pass.

typescript/packages/pay-kit/src/adapters/x402-upto.ts
lgalabru: Fixed in the `fix(ts)` commit: `#challengeRequirements` now throws when the recent-state fetch fails ("x402 upto challenge requires extra.recentBlockhash/recentSlot") instead of returning the bare requirements — the usage 402 propagates the failure, and the discovery path omits the upto offer rather than advertising one no client can act on. Covered by `fails the challenge instead of advertising a bare upto offer`.

lgalabru:
@EfeDurmaz16 both items from your re-review are addressed:

**upto openSlot binding** (0f09e61 / 17239e8 / 5509830 — Python, Go, Rust): the open verifiers now decode the instruction's own `openArgs` and enforce the slot-addressed channel invariant end to end — the channel account must equal the PDA re-derived from the args' `salt`/`openSlot`, the args' `deposit` must equal the authorized maximum, and `openSlot` must sit at-or-before the challenged `extra.recentSlot` inside the program's 1500-slot freshness window, all before the operator co-signs and broadcasts. Since the upto requirement is recomputed with a fresh slot at verify time, the bind is a window (`openSlot <= challenged` and `challenged - openSlot <= 1500`) rather than strict equality; `batch-settlement` (Rust) pins the exact config `recentSlot` through the same path. New coverage in each language (`test_client_open_tx_validator_binds_open_args`, `TestValidateUptoOpenInstructionBindsOpenArgs`, `binds_open_args_deposit_and_slot`). The TS server side lives in the vendored `@x402/svm` facilitator and is tracked with the fork.

**bare upto offer fallback** (a294e05, TS): the enriched requirement now throws when the recent-state fetch fails — the usage 402 propagates the failure and discovery omits the upto offer — instead of advertising an offer without `recentSlot` that no client can act on.

All language suites are green (Python 1320, Go 16 pkgs at 92.0% coverage, Rust 921 incl. the feature-gated server tests, TS 516).

EfeDurmaz16 (dismissed):
Re-reviewed a294e05. The two items from my previous review look addressed: upto open transactions now bind the decoded open args/deposit/slot to the expected channel window across Go/Python/Rust, and TS no longer advertises a bare unusable upto offer when recent-state fetch fails. The current red matrix looks like CI/setup fallout from Cargo failing to resolve solana-hash through LiteSVM (solana-slot-hashes v3.0.2 wants ^4.3.0 while litesvm selected ^4.2.0), so the harness jobs are not reaching the protocol assertions. Approving the code review, but CI still needs the resolver pin/rerun before merge.

EfeDurmaz16 (approved):
Re-reviewed 4727084 after the latest push. The Rust build fix removes the temporary litesvm git patch and moves integration tests to litesvm 0.13.1, which matches the prior Cargo resolver failure; the current rollup is 86 successful checks with only Lua still in progress and no completed failures. I also rechecked the earlier risk area: recentSlot/openSlot binding remains wired across TS/Rust/Python/Go, and TS still fails challenge creation instead of emitting a bare unusable upto offer. Verdict: approve, with CI left to finish before merge.

PR #222, opened 2026-07-09
fix(ruby): MPP replay-store fail-closed + x402 accepted-amount binding (+ #216 hardening)

EfeDurmaz16:
Ruby slice of the #216 pre-release security hardening, **re-delivered onto #219** (base = `split/pr216-ci-harness-mega`), plus the residual fixes the bedrock language audit found on top of it. No harness/CI files — those live in #219. First of the per-language / per-protocol PRs that flow into #219; when they all land, #219's SDK conformance/behavior gates go green.

## Security behavior
- **MPP replay-store fail-closed off-localnet** (`config.rb`, `runtime.rb`, `rack/payment_required.rb`). Localnet keeps the dev `MemoryStore` with a warning. **BREAKING:** non-localnet `Mpp.create` now requires a durable `replay_store` (raises `ConfigurationError`) — prevents same-signature replay across workers/restarts.
- **x402 v2 `accepted.amount` binding** (`schemes/exact/types.rb`): a credential echoing a drifting `amount`/`maxAmountRequired` is now rejected, aligned to the Rust verifier (`rust/crates/kit/src/x402/server/exact.rs:816`); omission is still tolerated for v2 client compatibility.
- Plus the rest of #216's Ruby hardening.

## Verification
`mise ruby@3.4.5 bundle exec rake test` → **463 runs, 1377 assertions, 0 failures, 0 errors**. New/updated regression tests cover the fail-closed store wiring and the accepted-amount tamper case.

## Follow-up
`refactor(ruby): idiomatic (Shopify Ruby + rubystyle.guide)` will follow as a separate PR (keeping the behavior fix and the style pass reviewable independently), also targeting #219.


EfeDurmaz16:
Fixed the 3/5 finding: the amount tolerance predicate was reversed (keyed on the server candidate `right`, which always carries the amount, so it was inert and a v2 credential omitting the amount got rejected). Now scheme/network/asset/payTo match strictly, and the amount is bound by VALUE across `amount`/`maxAmountRequired` only when the credential (`left`) echoes one — omission is tolerated, drift is rejected, mirroring the Rust verifier. Added `test_settlement_tolerates_accepted_omitting_amount` for the omission case (the drift case was already covered). Also fixed the Ruby CI red: CVE-2026-38969 (webrick, a transitive test dep with no patched release yet) is now a documented `ruby/.bundler-audit.yml` ignore with a removal trigger. standardrb clean, 464 tests green, bundle-audit clean.

@greptile-apps please review

EfeDurmaz16:
Fixed the 4/5 note: you're right, webrick is a declared runtime dependency in `solana-pay-kit.gemspec` (a Rack server backend alongside puma), not test/tooling-only. Corrected the `.bundler-audit.yml` rationale to say exactly that — the SDK never `require`s webrick directly, puma is the recommended production server, there's no patched webrick yet, and the removal trigger notes dropping webrick from runtime deps as the alternative. The ignore itself is unchanged (bundle-audit checks the full lockfile regardless of dep group).

@greptile-apps please review

EfeDurmaz16:
Both remaining findings fixed (7d... see commit):
- **webrick runtime-dep / CVE exposure:** moved `webrick` from `add_dependency` to `add_development_dependency` in the gemspec — downstream consumers no longer receive it (the SDK never requires webrick at runtime; puma is the production server). It stays in the dev lockfile so the documented `.bundler-audit.yml` ignore keeps CI green.
- **`invoke_signature_confirmer` **kwargs edge case:** the introspection only matched `:key`/`:keyreq`, so a confirmer accepting `**kwargs` (:keyrest) was called without `last_valid_block_height`. Now the splat is honored; added `test_signature_confirmer_with_kwargs_receives_last_valid_block_height`.

gemspec validate + standardrb clean, 465 tests green, bundle-audit clean.

@greptile-apps please review

EfeDurmaz16:
@greptile-apps please review

EfeDurmaz16:
Both P1s addressed:
- **exact.rb (release gate):** the replay-reservation release no longer depends on the client-echoed `accepted.extra.lastValidBlockHeight`. It now requires the transaction's OWN signed `recentBlockhash` to be independently proven expired on-chain (`isBlockhashValid == false`) via a trusted, injectable `blockhash_validator` (real RPC by default), fail-closed on any nil/ambiguous/RPC error. A low echoed height can no longer coerce a premature release.
- **payment_required.rb (nil passthrough):** no longer inserts an explicit `replay_store: nil` into the `Mpp.create` options, so a localnet dev without a store gets the `MemoryStore` fallback while non-localnet still fails closed.

Regression tests cover premature-release-blocked, genuine-release-on-expiry, validator-error fail-closed, and the localnet fallback. rake test 483/0, standardrb clean, branch coverage 90.07%. @greptile-apps please review

lgalabru (approved):

PR #214, opened 2026-07-02
fix(go): migrate to solana-go v2 and harden verification

EfeDurmaz16:
Go SDK idiomatic cleanup, `solana-go` v2 migration, and the Go security-hardening delta, delivered as one per-language PR onto #219.

### What this PR contains
- Idiomatic Go cleanup following Effective Go, Uber Go, and existing repository conventions.
- Full migration to `github.com/solana-foundation/solana-go/v2` across the SDK and Go harness modules; `gagliardetto/solana-go` is not reintroduced.
- The security changes previously reviewed in #224: real x402 exact-verifier conformance, managed fee-payer drain protection, RFC 8785 canonical JSON fixes, strict key parsing, and their regression tests.

### Stack equivalence
#224 is now subsumed: its final head and this PR's current head are the same commit, `1e8d8d2166657be224d341e1d2de1503f4077e02`. This preserves the separate review history while presenting maintainers with the requested single green Go delivery into #219.

### Diff note
The cross-language `upto` files visible against #219 come from already-landed #220 because #219 trails `main` by that commit; they are not authored by this Go PR.

### Verification
- `go test ./...`
- Go client and server harness tests
- downstream module consumer guard
- repository-wide `gofmt -l` clean
- GitHub Go lint and core lint/format checks

The routed full-language harness failures are integration-base gaps that clear as their corresponding language PRs land into #219; the Go-owned checks are required to remain green.

PR #224, opened 2026-07-09
fix(go): x402 source-ATA drain guard + canonical-JSON + conformance driver

EfeDurmaz16:
## Summary
- keep #224 Go-only after #226 was merged into this branch
- harden x402 exact managed-signer/source-ATA checks using the transaction's actual token program
- align Go JCS with RFC 8785 / ECMAScript float64 semantics and drive the real exact verifier from conformance
- require explicit shared replay protection off localnet and add downstream module-consumer coverage

## Scope
This PR is stacked on #214 and now contains 23 Go-owned files only: `go/**`, `.github/workflows/go.yml`, and the Go harness client/server. The non-Go portion of #226 was removed from the final diff and moved to the sibling per-language/harness PRs targeting #219.

## Verification
- `go test ./...`
- `golangci-lint run ./...`
- `go/scripts/check_module_consumer.sh`
- `gofmt -l` clean

## Stack
#214 must merge first; #224 can then be retargeted to #219. The shared harness PR lands after the language verifiers it exercises.


EfeDurmaz16:
Addressed both 4/5 notes: the fee-payer ATA derivation now fails closed (`return fmt.Errorf(...)`) instead of silently skipping the source-ATA check, matching the recipient-ATA derivation in the same function; and the dead `else if` in `rejectedX402Exact` (unreachable — every reject returns a `*x402.VerifyError`) plus its now-unused `strings` import are removed. Conformance still 0 failures; build/vet/test green. @greptile-apps please review

EfeDurmaz16:
@greptile-apps please review

PR #235, opened 2026-07-10
fix(ruby): require shared replay reservations

EfeDurmaz16:
Require an explicit durable-and-shared replay-store capability outside localnet. `FileStore` remains suitable for single-process examples but is rejected for multi-worker production because its mutex is not cross-process.

Tests:
- `mise exec ruby@3.3 -- bundle exec rake test` (484 runs, 1461 assertions)


EfeDurmaz16:
@greptile-apps please review.

EfeDurmaz16:
@greptile-apps please re-review exact head a20b530. This hardens the harness replay store with cross-process atomic replacement and tightens durable/shared store capability validation.

PR #245, opened 2026-07-15
refactor(rust): signing architecture design

lgalabru:
Superseding https://github.com/solana-foundation/pay-kit/pull/206. Versioned hardware-wallet transaction signing remains intentionally out of scope; https://github.com/solana-foundation/pay-kit/pull/210 stays open for that work.


rust/crates/kit/src/x402/server/exact.rs
```diff
@@ -645,28 +644,9 @@ impl X402 {
...
-                "fee payer is not a required transaction signer".into(),
-            ));
-        }
-        let signer_index = 0;
-        let signature = fee_payer
-            .sign_message(&tx.message.serialize())
-            .await
-            .map_err(|e| Error::Other(format!("fee payer signing failed: {e}")))?;
-        tx.signatures[signer_index] = Signature::from(<[u8; 64]>::from(signature));
+        crate::core::signing::cosign_versioned_fee_payer(fee_payer, &fee_payer_key, &mut tx)
```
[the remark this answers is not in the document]
lgalabru: Addressed in 805d446: the shared helper now documents why the configured sponsor must occupy fee-payer slot 0.

rust/crates/kit/src/core/signing.rs
```diff
@@ -0,0 +1,210 @@
...
+            "fee payer signer does not match the expected fee payer".into(),
+        ));
+    }
+    if tx.message.static_account_keys().first() != Some(expected_fee_payer) {
+        return Err(Error::Other(
+            "transaction fee payer does not match the expected fee payer".into(),
+        ));
+    }
+    sign_versioned_transaction_slot(signer, tx).await
+}
```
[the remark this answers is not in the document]
lgalabru: Addressed in 805d446: the helper now requires expected_fee_payer to be the independently configured sponsor key and explains the separate signer-identity check.

rust/crates/kit/src/x402/server/exact.rs
EfeDurmaz16: 🔴 blocking - Pin this against the configured sponsor. `expected_fee_payer` is derived from `fee_payer.pubkey()`, so the identity check is tautological: verification protects advertised sponsor A, but a mismatched supplied signer B can still sign a crafted transaction where B is fee payer and transfer authority. Parse the independently configured `self.config.fee_payer_key` or bound route requirement and reject mismatches before signing.

typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "packageManager": "pnpm@10.14.0",
+  "packageManager": "pnpm@11.13.0",
```
EfeDurmaz16, line 3: 🔴 blocking - pnpm 11.13.0 requires Node >=22.13, while this workspace still declares Node >=18 below. Node 18 and 20 now fail before install, and CI stays green because it only runs current Node 22; retain pnpm 10 or raise and test the declared Node floor.

EfeDurmaz16 (commented):
## Review - signer consolidation is sound, two small blockers remain

Verdict: Not approve

The Rust signing refactor preserves required-slot isolation across the verified paths, but sponsor identity is not actually pinned in `settle_exact`, and the pnpm bump breaks the declared Node support floor. Both fixes are small.

🔴 Blocking
- `rust/crates/kit/src/x402/server/exact.rs:647` - derive the expected sponsor from independently configured state; deriving it from the supplied signer makes the identity check tautological.
- `typescript/package.json:3` - pnpm 11.13 requires Node >=22.13 while the workspace still declares Node >=18.

🟡 Important
- The subscription and confidential regression tests use `MemorySigner`, so they would still pass if production reverted to `sign_message`. Reuse the transaction-only spy used by the charge test.
- The PR body says this supersedes #210, but the current head explicitly defers versioned hardware-wallet signing and still routes v0 transaction messages through `sign_message`. Keep #210 open or update the scope statement.

🎉 The shared helper improves slot isolation and malformed signature-vector handling.

rust/crates/kit/src/x402/server/exact.rs
lgalabru: Fixed in 61aa94f: settlement now parses and pins `self.config.fee_payer_key` before signing, and a regression test verifies a supplied signer cannot differ from the configured sponsor.

typescript/package.json
lgalabru: Fixed in 61aa94f: the repo now pins Node 22.13.0 in `.node-version`; every CI and publish workflow reads that same file. The declared Node floor is `>=22.13.0`, and all direct pnpm workspaces use pnpm 11.13.0.

lgalabru:
Addressed the remaining review feedback in 61aa94f: subscription and confidential signing tests now use transaction-only signers (so an off-chain `sign_message` regression fails), and the PR scope now keeps #210 open for versioned hardware-wallet signing. Node 22.13.0 is pinned in `.node-version` and used by every CI/publish workflow.

EfeDurmaz16 (approved):

PR #253, opened 2026-07-22
fix: mpp session lifecycle

lgalabru:
## Summary
- add a feature-gated Redis ChannelStore with atomic compare-and-swap updates
- make channel creation atomic and reject duplicate channel IDs without resetting accepted voucher state
- support TLS Redis connections
- add cursor-based, namespace-scoped channel enumeration for reconciliation workers
- deduplicate Redis SCAN results within each snapshot
- exercise the Redis store against a real Redis service in CI

## Test Plan
- `cargo test -p solana-pay-kit --lib`
- `PAY_KIT_TEST_REDIS_URL=redis://127.0.0.1:6379 cargo test -p solana-pay-kit --features redis-store --lib core::store::tests::redis_channel_store_roundtrip_and_atomic_watermark -- --exact`

### DIFF
--- a/rust/crates/kit/src/core/store.rs
+++ b/rust/crates/kit/src/core/store.rs
@@ -486,8 +486,14 @@ impl RedisChannelStore {
             .map_err(|e| StoreError::Internal(format!("Redis connect: {e}")))?;
         Ok(Self {
             connection,
-            key_prefix: key_prefix.into(),
+            key_prefix: Self::namespace_key_prefix(key_prefix.into()),
         })
+    }
+
+    fn namespace_key_prefix(key_prefix: String) -> String {
+        // Length-prefix the configured namespace so no encoded namespace can
+        // be a prefix of another, even when namespaces are nested.
+        format!("{}:{key_prefix}:", key_prefix.len())
     }
 
     fn key(&self, channel_id: &str) -> String {
@@ -808,6 +814,20 @@ mod tests {
     }
 
     #[cfg(feature = "redis-store")]
+    #[test]
+    fn redis_channel_store_prefix_has_namespace_boundary() {
+        let tenant = RedisChannelStore::namespace_key_prefix("tenant:1".to_string());
+        let adjacent = RedisChannelStore::namespace_key_prefix("tenant:10".to_string());
+        let nested = RedisChannelStore::namespace_key_prefix("tenant:1:child".to_string());
+
+        assert_eq!(tenant, "8:tenant:1:");
+        assert_eq!(adjacent, "9:tenant:10:");
+        assert_eq!(nested, "14:tenant:1:child:");
+        assert!(!adjacent.starts_with(&tenant));
+        assert!(!nested.starts_with(&tenant));
+    }
+
+    #[cfg(feature = "redis-store")]
     #[tokio::test]
     async fn redis_channel_store_roundtrip_and_atomic_watermark() {
         let Ok(redis_url) = std::env::var("PAY_KIT_TEST_REDIS_URL") else {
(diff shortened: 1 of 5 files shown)


rust/crates/kit/src/core/store.rs
```diff
@@ -379,6 +454,283 @@ impl ChannelStore for MemoryChannelStore {
...
+        // namespace is always interpreted literally.
+        let mut pattern = String::with_capacity(self.key_prefix.len() + 1);
+        for ch in self.key_prefix.chars() {
+            if matches!(ch, '*' | '?' | '[' | ']' | '\\') {
+                pattern.push('\\');
+            }
+            pattern.push(ch);
+        }
+        pattern.push('*');
+        pattern
```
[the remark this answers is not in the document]
lgalabru: Fixed in 13fa2a67. Redis namespaces are now length-prefixed before key construction, making the encoded namespace prefix-free even for adjacent or nested configured prefixes. Added a unit invariant test plus a Redis-backed overlapping-namespace regression assertion.

rust/crates/kit/src/core/store.rs
```diff
@@ -379,6 +454,289 @@ impl ChannelStore for MemoryChannelStore {
...
+        key_prefix: impl Into<String>,
+    ) -> Result<Self, StoreError> {
+        let client = redis::Client::open(redis_url)
+            .map_err(|e| StoreError::Internal(format!("Redis client: {e}")))?;
+        let connection = client
+            .get_connection_manager()
+            .await
+            .map_err(|e| StoreError::Internal(format!("Redis connect: {e}")))?;
+        Ok(Self {
+            connection,
```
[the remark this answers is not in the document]
lgalabru: Intentional; no compatibility layer will be added. RedisChannelStore is introduced by this PR and has not shipped from main, so there is no supported persisted keyspace to migrate. We will delete the pre-release staging keys once during rollout. Keeping the prefix-free format avoids carrying legacy ambiguity into the public API.

rust/crates/kit/src/core/store.rs
```diff
@@ -379,6 +454,293 @@ impl ChannelStore for MemoryChannelStore {
...
+    fn put_channel(
+        &self,
+        channel_id: &str,
+        state: ChannelState,
+    ) -> Pin<Box<dyn Future<Output = Result<(), StoreError>> + Send + '_>> {
+        let key = self.key(channel_id);
+        Box::pin(async move {
+            let encoded = Self::encode(&state)?;
+            let mut connection = self.connection.clone();
+            redis::cmd("SET")
```
EfeDurmaz16: This unconditional `SET` lets concurrent deposit requests interleave as `SET(0) -> accept(N) -> SET(0) -> accept(N)`, so both are treated as fresh and served. Please make channel creation atomic and reject existing state, or propagate replay to `serve = false`.

rust/crates/kit/src/core/store.rs
```diff
@@ -438,6 +800,119 @@ mod tests {
...
+        assert_eq!(adjacent, "9:tenant:10:");
+        assert_eq!(nested, "14:tenant:1:child:");
+        assert!(!adjacent.starts_with(&tenant));
+        assert!(!nested.starts_with(&tenant));
+    }
+
+    #[cfg(feature = "redis-store")]
+    #[tokio::test]
+    async fn redis_channel_store_roundtrip_and_atomic_watermark() {
+        let Ok(redis_url) = std::env::var("PAY_KIT_TEST_REDIS_URL") else {
```
EfeDurmaz16: This silently passes without executing Redis, while the PR's documented test filter does not select this test. Please run the exact Redis-backed test against a Redis service in CI and fail when its URL is missing.

EfeDurmaz16 (changes requested):
Requesting changes for one payment-safety race and one false-green CI gap. Redis-backed initialization must not overwrite live voucher state, and the real Redis path needs a blocking integration test. The remaining feature compilation, formatting, and focused tests are clean.

rust/crates/kit/src/core/store.rs
lgalabru: Fixed in 3d6953bd. `ChannelStore::put_channel` is now explicitly atomic create-only: both the memory store and Redis store reject an existing channel ID, and Redis uses the existing Lua CAS with an absent expectation. Concurrent deposit replays can no longer reset the watermark or both serve. Added regressions for duplicate preservation and two concurrent Redis creates having exactly one winner.

rust/crates/kit/src/core/store.rs
lgalabru: Fixed in 3d6953bd. The Rust CI job now provisions a health-checked Redis 7 service and runs the fully qualified Redis integration test with `--exact` and `PAY_KIT_TEST_REDIS_URL` set. The test now fails immediately when that URL is absent instead of silently returning. I also updated the PR test plan to the exact command and verified it locally against a real Redis container.

EfeDurmaz16 (approved):
LGTM, thanks !

PR #259, opened 2026-07-31
feat(mpp/session)!: final session wire contract + open context — Rust, TypeScript, Python (Go/Kotlin/Swift follow-up)

lgalabru:
## Summary
- replace the legacy settlement authority contract with `voucherSigner: client | operator` across Rust, TypeScript, and Python
- add reusable channel-bound payer proofs, the operator-mode `use` action, and operator-signed cumulative vouchers
- add negotiated idle-timeout fields, validation helpers, lifecycle configuration, and cross-SDK tests
- **Post-approval addition (per channel decision):** `process_commit` now refreshes `last_activity_at`/`lifecycle.close_after` — the idle-close root fix, folded into this PR instead of a stacked follow-up (details below)
- **Scope note:** the Go/Kotlin/Swift migration to the final e702dd8 wire contract is **de-scoped to a follow-up PR** (details below). Those SDKs received only the multiDelegate removal here; their session-wire conformance vectors are explicit, allowlisted skips.

## Fund-safety fix (from review)
`distributionSplits` are now verified against the **challenge** in TypeScript and Python, mirroring the existing Rust check (`payload_splits != config.splits` in `server/session.rs`). Previously both SDKs only checked payload↔transaction self-consistency, so a client could open with altered or empty splits: the open commits the on-chain `distributionHash` while `distribute` at settle is built from the server's config, so the substituted list steals platform-fee co-recipient shares and makes the bundled settleAndSeal+distribute revert atomically — stranding vouchers until the client force-closes and reclaims the full deposit.

- TypeScript: `VerifyOpenTxExpected` gains a **required** `splits` field enforced inside `verifyOpenTx`; both `handleOpen` paths plumb the configured splits through.
- Python: `_process_open_locked` rejects `payload.distribution_splits != config.splits`.
- Regression tests in both SDKs: drop-splits, redirect-share, inflate-share, and unsolicited-splits opens rejected; exact challenged splits accepted.

## Idle-close root fix (added post-approval, `a569ac52`)
`process_commit` advanced the voucher watermark but never refreshed the activity state, so a channel paying **exclusively through the metered-delivery flow** looked idle to the host's lifecycle worker and was idle-closed mid-use — the spine-side root cause of the failure class pay#416's lease heartbeat patches from the pay side.

- **Rust:** the committed path re-arms `last_activity_at` + `lifecycle.close_after` exactly like the voucher/use/top-up paths; idempotent commit replays leave both untouched (matching use-replay semantics). Regression test: `commits_refresh_activity_and_lifecycle`.
- **Python:** the commit mutator refreshes the durable `last_activity_at` read by the idle-close recheck and the post-restart reconcile (the route-level timer touch alone only delayed the fire). Regression test: `test_commit_refreshes_activity_watermark_but_replay_does_not`.
- **TypeScript:** already correct — `processCommit` refreshes `lastActivityAt` and the fire-time recheck is covered by `session-lifecycle.test.ts`. No change.
- **Go/Kotlin/Swift:** no session activity tracking yet; these semantics land with the wire-migration follow-up. The session spec reference now spells out which actions refresh `lastActivityAt`.

**Deliberately defense-in-depth with pay#416's lease heartbeat, not a replacement:** the heartbeat covers commit-less activity (a delegated lease held open with no payments flowing), which this spine fix structurally cannot see. Both activity signals belong; neither should be "simplified" away later.

Suites re-run at `a569ac52` (only Rust/Python code changed): Rust `cargo test` workspace — 832 passed, 0 failed (kit lib, +1 new test), all other targets green, `cargo fmt --check` + clippy clean (no new warnings); Python `just test` — 1213 passed, 1 skipped, 0 failed (+1 new test), ruff + pyright clean. TS/Go/Kotlin/Swift/conformance are untouched by this commit; their counts in the table below are from `2daa4505`.

## Breaking Changes
- `settlementAuthority` is removed; use `voucherSigner`
- `clientVoucher` becomes `client`; `delegated` becomes `operator`
- `closeDelayMs` is removed from the TypeScript server API; use `idleTimeoutSeconds`
- TS `VerifyOpenTxExpected` gains a required `splits` field (the challenged distribution splits)

## Open-transaction context (mpp-specs e702dd8)

- New-channel session challenges (no `channelId`) now carry `recentBlockhash` + `recentSlot` (decimal-string u64), both from **one** `getLatestBlockhash` (`result.value.blockhash` + `result.context.slot`). The server **fails the challenge instead of degrading** when the fetch fails; both fields stay absent on resume challenges.
- Clients use the challenged blockhash for the open transaction and default `openSlot = recentSlot` (earlier override allowed, later rejected; clear error when a new-channel challenge lacks either field).
- Open verification enforces `openSlot <= recentSlot`, `recentSlot - openSlot <= OPEN_SLOT_WINDOW` (1500), and that the compiled open message uses the challenged blockhash — all before broadcast. Same pattern in the x402 `upto` adapter via `extra.recentBlockhash` / `extra.recentSlot`.
- `SessionServer::with_blockhash_cache(...)` (Rust; mirrored hooks in TS/Python) replaces the per-challenge blocking `get_slot()` fetch.
- The Python wire parser now rejects requests missing required fields instead of defaulting them to empty strings (caught by the new conformance vectors).

### Golden session wire vectors

`harness/vectors/session-wire.json` (new `sessionWire` canonical-bytes mode): frozen new-channel challenge, resume challenge, and open payload, plus rejects for the pre-e702dd8 draft request, the draft `recentSlot` open-payload echo, and unknown action tags. **Honest scope:** these are enforced by the TypeScript and Python parsers; Go/Kotlin/Swift declare expected skips pinned to an explicit allowlist in the Go seeded conformance test (any new unexpected skip fails, and a leftover-allowlist check forces the allowlist's deletion once the migration lands). The vector descriptions carry the same note.

### Notes for the `pay` bump

1. Pass the existing gateway `BlockhashCache` down via `SessionServer::with_blockhash_cache(...)` and drop the field overwrite in `core/src/server/session.rs` — `SessionMpp` already holds the cache under the same builder name.
2. The best-effort "cache miss and RPC down → challenge without hints" degrade path must become **fail-the-challenge** for new channels (resume challenges unchanged). With the 10s refresh / 45s staleness cap this only fires when RPC has been down ~45s+, at which point the open would fail anyway.

## Follow-up PR: Go/Kotlin/Swift wire migration (required, tracked)
- Migrate Go `intents/session.go`, Kotlin `SessionTypes.kt`, Swift `SessionTypes.swift` off the draft wire contract (top-level `cap`/`operator`/`programId`, draft `recentSlot` open echo) to final e702dd8.
- **Must also remove Go's draft-shape acceptance:** `encoding/json` zero-values missing fields, so Go currently accepts both the draft shape and a mis-parsed final shape — an interop hazard, not cosmetic.
- Delete the session-wire expected-skip allowlist in `go/cmd/conformance/main_test.go` (its leftover-allowlist check fails loudly until deleted) and the Kotlin/Swift harness-runner skips.
- **Inherits the spec voucher wire shape (`7085443f`):** signed voucher's inner key is `voucher` (not `data`, in the voucher action and nested in close), the voucher action carries a REQUIRED top-level `channelId` the server rejects on mismatch, and an omitted `expiresAt` is never-expires encoded verbatim as 0. Also carries: commit-activity semantics, `schema_version`/`extra` round-trip, `processed_topup_signatures` exactly-once.

## Review follow-ups (tracked, not folded in)
- **Rust (medium):** operator-mode `use`/`close` lack outer-challenge expiry plumbing; `process_use` idempotency lacks a request fingerprint. (~~`process_commit` doesn't refresh `last_activity_at`/`lifecycle.close_after`~~ — **fixed in this PR** at `a569ac52`, see the idle-close root fix section.)
- **Python (medium):** extend the strict-parser fix to `charge.py` (empty-string defaults for amount/currency/recipient) and `_paycore/solana.py` (missing network defaults to mainnet).
- **All SDKs (low):** spec step 7's second clause (window vs the server's *current* slot) is unimplemented; TS x402-upto `#assertOpenSlotBoundToChallenge` fails open on RPC error and Rust upto passes `challenged_slot = None` on fetch failure — both should fail closed.
- **Harness (pre-existing):** the sessionWire silent-skip shape in `conformance.test.ts` — skips are uncounted, so dropping `session` from `python.json` would green-skip everywhere; the driver needs skip accounting.
- **Spec drift (4f3918c):** `Subscription.ts` still types/defaults `mainnet-beta`; Rust doc comments still advertise `testnet`.
- **TS test gaps (low, from re-review):** `session.routes()` deliveries/commit endpoints and the default-store sharing between `routes()` and `verify()` have zero direct tests (new client tests mock the gateway); method-level open-replay semantics (replay preserves watermark, different `authorizedSigner`/sealed rejects, no-rebroadcast) are covered at store level but not credential level; the `expiresAt` safe-integer guard in `server/session/voucher.ts` is untested.
- **Harness (low):** the 22-byte fallback secret in `harness/src/fixtures/typescript/shared.ts` will break when the harness workspace bumps its own mppx (pinned 0.5.x) past 0.8; and the python-server accept-all-verifier comment overstates what the session e2e leg proves (transaction bytes are never decoded) — wording fix.
- **TS (low):** the session challenge zod schema marks `recentBlockhash`/`recentSlot` plain-optional — it documents but does not refine the conditional rule (required when `channelId` absent, forbidden on resume).

From Efe's round-2 pass (2026-08-01, none move money today):
- **Python (medium):** watchdog re-arm gap — an idle fire during a client settle, or the losing side of two racing client closes, drops the timer via the `_AlreadySettling` `None`; channel stays close-pending but re-drivable. Fix shape: remove the timer only once a signature is recorded or the channel is sealed; re-touch with the retry delay when a settle resolves `None` while still close-pending with signer/RPC configured.
- **Rust (low):** `store.rs` `mark_sealed_atomic`/`mark_finalized_atomic` substring guards on `"sealed":true` can false-positive via a foreign field round-tripped through the flattened `extra` — needs a JSON-aware check.
- **Rust (low):** `fetch_and_match_open_channel` adoption should require `settlement == 0` so the broadcast-error adoption arm cannot adopt a channel with a nonzero settled watermark while `fresh_state` seeds zero.
- **All SDKs (medium, additive):** session receipts are missing spec-REQUIRED `intent`/`acceptedCumulative`/`spent`/`idleTimeoutSeconds` — clients cannot recover the accepted watermark from receipts; deserves its own vector freeze.
- **All SDKs (low):** spec open step 7 current-slot window, step 2 non-curve `authorizedSigner` reject before fees are paid, and the open wire-`bump` MUST-reject are unimplemented everywhere; program fails closed, exposure is sponsored fees on guaranteed-to-revert opens.

## Test results (full suites, exact counts, tree = 2daa4505 (content-identical to e9e2ac0b, re-signed))

| Suite | Command | Result |
|---|---|---|
| Rust | `cargo test -p solana-pay-kit --features server,client --lib` / default | 831 + 355 passed, 0 failed |
| TypeScript workspace | `pnpm test` + typecheck + lint | **510** passed, 0 failed (419 before coverage reinstatement) |
| TypeScript integration | `pnpm test:integration` vs `harness/start-surfnet-proxy.mjs` | 8 passed, 0 failed (incl. both USDC legs) |
| Python | `pytest` + ruff + pyright | 1212 passed, 0 failed, 1 env-gated skip |
| Go | `go test ./...` | 21 packages ok; conformance runs all vectors with exactly the nine allowlisted session-wire skips |
| Kotlin | `gradle cleanTest test` | 309 passed, 0 failed |
| Swift | `swift test` | 202 passed, 0 failed |
| Cross-SDK conformance | `MPP_CONFORMANCE_LANGUAGES=typescript,python,go,kotlin,swift` | 156 passed, 70 by-design skips, 0 failed |
| Session e2e | harness `session-basic` (python server + python session client) | 2 passed, 0 failed |

### Reinstated TypeScript coverage
`session-settlement` (20: topUp update/rejects, close monotonicity, idempotent replay, failed-settlement retry, first-ever operator-vs-client close-auth tests), `session-verify-open-tx` (29: ALT reject, payee/mint/rentPayer mismatches, missing payer signature, confirmation polling, submitOpenTx re-verify), `session-onchain-builders` (15: byte-golden settleAndSeal/topUp/distribute/reclaim + client open incl. openSlot PDA seed), `challenge-selection-session` (16), `session-usage-meter`, `session-client-fetch`. Cases for genuinely removed surface (multiDelegate, modes, nonce, draft aliases) were dropped, not ported.

---

### Round-3 fix batch (Efe re-review, 2026-08-01 — tree `d1f2beed`)

All four round-2 blockers fixed, each red-tested (fix reverted → exactly the new test fails):

- **`dc30ef65` (Py, fund-adjacent):** omitted voucher `expiresAt` now encodes verbatim as 0 in the signed 50-byte preimage (was `DEFAULT_SESSION_EXPIRES_AT`), watermark records 0, docstring corrected, remaining sentinel uses audited. Cross-SDK freeze: `session-voucher-preimage-no-expiry` runs in the TS/Python/Go/Kotlin/Swift runners.
- **`08c77fb4` (Py):** open + top-up verifiers resolve preflight-rejected retries against the chain (confirmed-account field match for open; landed-clean signature status → confirm → deposit re-check for top-up; landed-but-FAILED stays authoritative). Mock RPC rejects duplicate sends like mainnet.
- **`7085443f` (all 3 SDKs, breaking wire):** spec voucher shape per ludovic's call — inner key `data` → `voucher`, REQUIRED top-level `channelId` on the voucher action (emit + parse + server reject on mismatch, close's nested voucher included). Frozen: voucher-action + close-action vectors (close also freezes omitted-`expiresAt`), plus a reject vector for the superseded inner-`data` shape. Go skip allowlist +3 until its wire migration.
- **`d1f2beed` (TS):** `submitOpenTx` shared the naked send — same confirmed-account rescue as Rust/Python.

No new version bumps: folds into the unpublished 0.5.0/0.8.0/0.6.0 breaking bump. **History note:** the full branch was rewritten once more and every commit SSH-signed (`2fad1c61` → this range); trees verified content-identical before the force-push.

| Suite | Result at `d1f2beed` |
|---|---|
| Rust `cargo test` (workspace) | 12 targets, 848 passed, 0 failed; `--features redis-store --lib` 360 passed vs live Redis |
| Rust fmt/clippy | clean (one pre-existing `upto.rs` `too_many_arguments` warning) |
| TypeScript `pnpm test` + prettier/eslint/tsc | 518 passed, 0 failed |
| Python `pytest` + ruff + pyright | 1231 passed, 1 env-gated skip |
| Go `go test ./...` | 21 packages ok (incl. conformance with the nine allowlisted skips) |
| Cross-SDK conformance | typescript,python,go: 139 passed / 9 by-design skips; kotlin,swift: 29 passed / 70 by-design skips — includes all 4 new vectors |

---

### 2026-08-01 incident follow-up (`3524a11b`)

Old writers (a draining revision, or any service pinned to pre-binding PayKit) sweeping the shared Redis lossily re-encoded newer rows and stripped the session proof binding. One commit, canonical here and mirrored to TS/Python:

- **Old-writer-safe records**: persisted `ChannelState` gains `#[serde(flatten)] extra` (unknown fields round-trip verbatim through read-modify-write) and a `schema_version` stamped on every durable write; stores refuse records stamped newer than the running writer instead of decoding them lossily. TS ships the contract as documentation (`CHANNEL_STATE_SCHEMA_VERSION`) since its stores are integrator-owned.
- **Error split**: use/close against a row whose binding fields are absent now fails `"session channel predates proof binding"` instead of sharing the generic `"does not match the proof bound at open"` — same fail-closed outcome, self-identifying in logs. No frozen wire vectors reference these strings.

Verified at `3524a11b`: rust `cargo test` 12 targets / 847 passed (+ redis-store feature suite against live Redis), TS 512 passed + lint/format/typecheck, Python 1216 passed + ruff + 90% coverage gate. Cascades: pay#416 (pin + mirror), agent-gateway#14 (settlement-worker pin).


python/src/solana_pay_kit/protocols/mpp/server/session.py
```diff
@@ -335,10 +349,28 @@ async def process_open(self, payload: OpenPayload) -> ChannelState:
...
+            validate_idle_timeout_options(self._config.idle_timeout_options_seconds)
+        if self._config.voucher_signer == "client" and payload.authentication is not None:
+            raise ValueError("authentication is only valid when voucherSigner is operator")
+        if self._config.voucher_signer == "operator":
+            if payload.authentication is None:
+                raise ValueError("operator voucher signing requires authentication")
+            expected_payer = payload.owner or payload.payer
+            if expected_payer is not None and payload.authentication.payer != expected_payer:
+                raise ValueError("session authentication payer does not match the channel payer")
+            if not verify_session_authentication(payload.authentication, session_id):
```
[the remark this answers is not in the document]
lgalabru: Reworked in 72d83249 to match the final session spec. The authenticationExpires field was removed from the Rust, TypeScript, and Python wire, config, and state surfaces. The standard challenge expires timestamp is enforced when opening the channel; it does not invalidate the bound proof after open. Python and TypeScript now cover rejecting an expired open and permitting established-channel activity after the opening challenge expires.

typescript/packages/mpp/src/server/Session.ts
```diff
@@ -568,6 +641,8 @@ async function handleOpen(args: HandleOpenArgs): Promise<Receipt.Receipt> {
         deposit,
         highestVoucherExpiresAt: undefined,
         highestVoucherSignature: undefined,
+        idleTimeoutSeconds: effectiveIdleTimeoutSeconds,
```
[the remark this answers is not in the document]
lgalabru: Fixed in 339aa75a and retained in the current head. Lifecycle touches accept the persisted per-channel idleTimeoutSeconds, and open arms the watchdog with the negotiated value. A fake-timer regression test verifies that the selected timeout overrides the server default.

typescript/packages/mpp/src/server/Session.ts
```diff
@@ -619,6 +695,76 @@ interface HandleVoucherArgs {
...
+        ]);
+        const signature = signatures?.[args.operatorVoucherSigner.address];
+        if (!signature) throw new Error('operator voucher signer did not return a signature');
+        voucherSignature = getBase58Decoder().decode(new Uint8Array(signature));
+        return {
+            ...current,
+            cumulative,
+            highestVoucherExpiresAt: BigInt(DEFAULT_SESSION_EXPIRES_AT),
+            highestVoucherSignature: voucherSignature,
+            lastActivityAt: Date.now(),
```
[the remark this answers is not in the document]
lgalabru: Fixed in 339aa75a and retained in the current head. Accepted operator use actions refresh the lifecycle timer with the channel negotiated timeout. Accepted vouchers, commits, and top-ups also refresh activity; replay-only operations do not extend the timer.

typescript/packages/mpp/src/server/session/lifecycle.ts
```diff
@@ -60,8 +61,9 @@ export function createLifecycle(
             for (const handle of timers.values()) clearTimeout(handle);
             timers.clear();
         },
-        touch(channelId) {
-            if (closeDelayMs <= 0) return;
+        touch(channelId, idleTimeoutSeconds) {
+            const effectiveTimeoutMs = idleTimeoutSeconds === undefined ? idleTimeoutMs : idleTimeoutSeconds * 1_000;
```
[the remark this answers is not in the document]
lgalabru: Fixed in cd1a8296. The lifecycle now schedules against an absolute deadline and chunks delays at the Node timer maximum, rearming until the negotiated deadline is reached. A fake-timer regression covers a timeout above the runtime limit.

EfeDurmaz16:
@greptile-apps review please

EfeDurmaz16:
I'm on it !

rust/crates/kit/src/mpp/server/session.rs
```diff
@@ -498,101 +577,122 @@ impl<S: ChannelStore> SessionServer<S> {
...
+                    "invalid session authentication signature".to_string(),
+                ));
             }
+        } else if payload.authentication.is_some() {
+            return Err(Error::Other(
+                "authentication is only valid when voucherSigner is operator".to_string(),
+            ));
         }
 
+        verify_submit_and_fetch_open(
```
EfeDurmaz16, line 681: This broadcast runs before the store's atomic replay check, so a client retry after a lost response fails at preflight with AlreadyProcessed instead of returning the documented idempotent replay; the store's replay branch is unreachable against a real RPC, and the deposit sits locked on chain until a force-close. The pre-PR get_signature_status check was naturally idempotent, and TS and Python only broadcast on the fresh-create branch, so this is also a spine parity break. process_topup has the same shape. Short-circuiting to the replay path before broadcasting, or treating AlreadyProcessed plus a matching confirmed channel account as success, would restore the contract. The replay test only passes because the mock RPC accepts repeated sends.

typescript/packages/mpp/src/server/Session.ts
```diff
@@ -730,30 +844,30 @@ async function handleTopUp(args: HandleTopUpArgs): Promise<Receipt.Receipt> {
...
         if (!current) throw new Error(`Channel ${args.payload.channelId} not found`);
         if (current.sealed) throw new Error('Channel is already sealed');
         if (current.closeRequestedAt !== undefined) {
             throw new Error('Channel close is pending — no further top-ups accepted');
         }
-        if (newDeposit <= current.deposit) {
-            throw new Error(`newDeposit ${newDeposit} must exceed current deposit ${current.deposit}`);
-        }
-        return { ...current, deposit: newDeposit };
+        return { ...current, deposit: current.deposit + additionalAmount, lastActivityAt: Date.now() };
```
EfeDurmaz16: handleTopUp has no idempotency: this mutator unconditionally adds additionalAmount, nothing dedupes by transaction signature, and the Rust post-confirm on-chain deposit re-check (verify_submit_and_fetch_topup) was not ported. Two concurrent submissions of the same signed top-up both pass preflight while the tx is in flight, both confirm the same signature, and both increment, so the stored deposit rises twice while escrow rose once and vouchers get accepted beyond what settle can collect. Recording processed top-up signatures on ChannelState and re-checking the on-chain deposit before incrementing would close both holes.

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -759,73 +468,9 @@ async def _handle_close(self, payload: ClosePayload) -> str:
...
-                        raise ValueError(err)
-                    nxt.cumulative = cumulative
-                    nxt.highest_voucher_signature = voucher.signature
-                    nxt.highest_voucher_expires_at = voucher.data.expires_at
-            nxt.close_requested_at = now
-            return nxt
-
         try:
-            await self._core.store().update_channel(channel_id, mutator)
+            await self._core.process_close(payload)
```
EfeDurmaz16, line 469: The docstring still promises a re-drivable close, but the call now delegates to core process_close, whose mutator hard-rejects any close once close_requested_at is set; the old inline mutator's settled_signature-is-None re-drive branch was dropped in this PR. The idle watchdog does not recover it either: _close_on_idle only settles closes it flipped itself. So a transient settle failure leaves the channel permanently close-pending with accepted vouchers unsettled, and the payer reclaims the full deposit after the grace period. Restoring the re-drive semantics (in core or here) fixes it.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
```diff
@@ -553,7 +738,7 @@ async def settle_and_seal_channel(
     # another account (e.g. the recipient) would derive the wrong refund ATA
     # and only fail after the settle transaction is built and broadcast.
     # Mirrors Go's strict payer handling.
-    payer_address = state.operator
+    payer_address = state.payer
     if not payer_address:
         raise PaymentError(
             f"channel {state.channel_id} payer is unknown; cannot derive the refund account", code="invalid-config"
```
EfeDurmaz16, line 799: Just below this block, distribute is built with rent_payer=config.operator, but open now pins the channel rentPayer to fee_payer_key when fee_payer is set and to the payer otherwise (session.py:555). In any non-gasless config, or whenever feePayerKey differs from the operator, the on-chain rentPayer check rejects distribute and the bundled settleAndSeal reverts with it, so settle can never land. The stored state.rent_payer is the right source, matching the TS path. The harness e2e only passes because the python server runs gasless with the operator as fee payer.

python/src/solana_pay_kit/protocols/mpp/server/session.py
```diff
@@ -266,102 +348,216 @@ def store(self) -> ChannelStore:
...
-        if operator is None:
-            operator = payload.payer
+        if self._config.verify_open_tx is None:
+            raise ValueError("open transaction verification requires a configured RPC verifier")
+        try:
+            await self._config.verify_open_tx(payload, context)
+        except Exception as exc:
+            raise _wrap("open tx verification failed", exc) from exc
+
+        effective_idle_timeout = resolve_idle_timeout_seconds(
```
EfeDurmaz16, line 542: resolve_idle_timeout_seconds runs after verify_open_tx has already broadcast and confirmed the funding transaction, and it raises for a selection the server did not advertise, so an unsupported idleTimeoutSeconds locks the deposit on chain and then fails the open on every retry. Rust resolves this before the broadcast (session.rs:608) and so does TS (Session.ts:573); moving this above the verify_open_tx call restores fail-safe ordering.

typescript/packages/mpp/src/client/PaymentChannels.ts
```diff
@@ -203,20 +205,22 @@ export function createPaymentChannelSessionOpener(
...
-            throw new Error('payment-channel session opener requires a pull-mode session challenge');
-        }
-        if (challenge.request.pullVoucherStrategy !== 'clientVoucher') {
-            throw new Error('payment-channel session opener requires pullVoucherStrategy=clientVoucher');
-        }
-
-        const sessionSigner = parameters.sessionSigner ?? (await generateKeyPairSigner());
+        const voucherSigner = challenge.request.methodDetails.voucherSigner ?? 'client';
+        const sessionSigner =
+            voucherSigner === 'operator'
```
EfeDurmaz16, line 210: Non-blocking, tracking here: in operator mode the ActiveSession signer is the payer while authorizedSigner is the operator, and SessionFetch always signs commit vouchers with the session signer (SessionFetch.ts:395), so the server rejects every voucher this opener produces; the Rust client opener has the same gap (client/session.rs:611). No fund risk since everything is rejected, but the operator flow is not usable end to end from these clients yet. Fine as a follow-up alongside the Go/Kotlin/Swift migration if you prefer.

EfeDurmaz16 (changes requested):
Solid migration overall, and both the splits challenge binding and the commit watermark fix check out line by line. I hit five money-path issues though, one in Rust, one in TypeScript, three in Python; details inline. Also flagged inline as non-blocking: the operator-mode opener cannot yet produce vouchers the server accepts in either the TS or Rust client. Verified by reading the full changed files at a569ac52 in a worktree and tracing each finding to the code; CI is green.

rust/crates/kit/src/mpp/server/session.rs
EfeDurmaz16: One sharpening note on the fix direction: reordering alone (replay check before broadcast) does not cover the hardest case, a broadcast that lands followed by a store write failure. No state exists then, so the retry still dies at AlreadyProcessed before any replay check can help. That case only recovers if AlreadyProcessed plus a matching confirmed channel account is treated as success, so I would treat that as the primary fix and the reordering as the optimization.

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
EfeDurmaz16: Small wording correction to my scenario: the payer's deposit recovery after the stranded close runs through the on-chain request_close, seal and distribute path once the grace period passes, not through reclaim; reclaim only recovers the channel rent. The merchant-side outcome is unchanged, the accepted vouchers never settle.

rust/crates/kit/src/mpp/server/session.rs
lgalabru: Fixed in dd545d2d, taking your refinement as the primary fix: the broadcast result is no longer authoritative. On a broadcast rejection, open falls through to the existing `get_account` + field match against the verified params, and top-up checks `get_signature_status` for the transaction's own signature — a landed non-error status is treated as confirmed. That covers the plain retry and the hardest case you called out (broadcast landed, store write lost, so no state exists for the replay branch); the reorder alone indeed couldn't.

The test-side gap is closed too: the mock RPC now rejects a duplicate `sendTransaction` at preflight with `AlreadyProcessed` and reports `getSignatureStatuses` only for landed signatures. With that mock and the fix reverted, `confirmed_open_is_persisted_and_replays_without_reset`, the new `retried_open_survives_preflight_rejection_without_stored_state` (empty store + landed broadcast), and the extended top-up test are all red — your exact scenario — and green with the fix. You were right that the old replay test only passed because the mock was friendlier than mainnet.

typescript/packages/mpp/src/server/Session.ts
lgalabru: Fixed in 34f293f2, and your fix direction became the canonical one in all three SDKs rather than a TS patch: `ChannelState` records processed top-up signatures and the check-and-record lives inside the atomic mutator, so only the first submission credits. Rust got the same in-mutator dedupe in dd545d2d (its pre-CAS on-chain re-check ran on a stale read — two in-flight duplicates of the same signature could both pass it), Python in 8a94c02b. The schema addition rides the round-trip/schema_version contract from 3524a11b, so old writers preserve it.

Also ported the Rust post-confirm deposit re-check you noted was missing: `submitTopUpTx` now fetches the confirmed channel account and requires it to be open and reflect the recorded deposit plus the top-up before the cap is raised. New tests: a resubmitted top-up is answered from the recorded signature without re-broadcast, and two concurrent duplicates (`Promise.all`) credit exactly once — both red with the dedupe disabled.

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
lgalabru: Fixed in 5fb48361, both halves. The re-drive carve-out moved into core `process_close` where it belongs (the method docstring now just points at it): close-pending with no settlement signature re-drives a matching close idempotently — the retry must replay the recorded final voucher and the original close timestamp is preserved — and hard-rejects once a signature is recorded. The merge-base's drop-through-refactor tests are restored in that shape.

`_lifecycle.remove_channel` now runs after the settle attempt, not before. That alone would have been cosmetic though: `_close_on_idle` bailed on any close-pending channel and `_reconcile_lifecycle` skipped them at boot, so nothing would ever have retried. The watchdog now re-drives a stranded settle (close-pending, no signature), re-arms its timer after a failed settle, and reconciliation restores timers for stranded channels across restarts. End-to-end test drives a close whose settle raises, asserts the watchdog is not released and the channel stays re-drivable, then retries the same payload to a successful settle.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
lgalabru: Fixed in 8a94c02b — the one-liner you suggested: distribute is built with `state.rent_payer` (falling back to omitting rent recovery when an old record has none), matching TS. Added the non-gasless test you asked for: a channel whose recorded rent payer is neither the operator nor the payer settles successfully and the built transaction references the recorded rent payer — red with the old `config.operator` code.

python/src/solana_pay_kit/protocols/mpp/server/session.py
lgalabru: Fixed in 8a94c02b: `resolve_idle_timeout_seconds` now runs before `verify_open_tx`, matching the Rust/TS ordering. Test asserts an unsupported selection rejects with zero verifier invocations, i.e. before anything could broadcast — red against the old ordering.

typescript/packages/mpp/src/client/PaymentChannels.ts
lgalabru: Agreed and tracking, not fixed here: this folds into the client-side work item from the incident thread — returning the opening proof on the session handle plus the use-after-rotation conformance vector across the five SDKs — so the operator-mode opener gets fixed once with its conformance coverage rather than patched per SDK. Will reference this comment when that lands.

EfeDurmaz16:
@greptile-apps can I have a review ?

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
```diff
@@ -381,57 +447,193 @@ def account_at(indices: list[int], slot: int, label: str) -> Pubkey:
...
+            recent_blockhash=context.recent_blockhash,
+        )
+        verified = await verify_open_tx(expected, payload, None)
+        if config.fee_payer:
+            if fee_payer_signer is None:
+                raise PaymentError("fee-payer signing requires a configured signer", code="invalid-config")
+            signature = await cosign_and_broadcast_open(payload, fee_payer=fee_payer_signer, rpc=rpc_client)
+        else:
+            raw = base64.b64decode(payload.transaction, validate=True)
+            sent = await rpc_client.send_raw_transaction(raw)
```
EfeDurmaz16: Blocking: the open verifier treats a broadcast rejection as authoritative, so a landed open whose confirm timed out or whose persist was lost dies at preflight with AlreadyProcessed on every retry, deposit escrowed with no server record. Rust got the confirmed-account fallback in this range; TS submitOpenTx (on-chain.ts:916) shares the naked send. Fix: port the Rust fallback to Python and TS, same shape as dd545d2d.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
```diff
@@ -381,57 +447,193 @@ def account_at(indices: list[int], slot: int, label: str) -> Pubkey:
...
+            raise PaymentError("top-up instruction does not exactly match additionalAmount", code="invalid-payload")
+        required_signers = account_keys[: int(message.header.num_required_signatures)]
+        try:
+            payer_index = required_signers.index(state.payer)
+        except ValueError as exc:
+            raise PaymentError("top-up payer is not a transaction signer", code="invalid-payload") from exc
+        payer = Pubkey.from_string(state.payer)
+        if payer_index >= len(signatures) or not signatures[payer_index].verify(payer, _signed_message_bytes(message)):
+            raise PaymentError("top-up payer signature is invalid", code="invalid-payload")
+        sent = await rpc_client.send_raw_transaction(raw)
```
EfeDurmaz16: Blocking, same class as the open verifier above: no landed-signature rescue, and the processed_topup_signatures dedupe only covers persisted credits, so a landed top-up whose confirm or persist was lost is uncreditable for the channel's lifetime. Fix: port the Rust/TS signature-status rescue.

python/src/solana_pay_kit/protocols/mpp/intents/session.py
```diff
@@ -657,10 +517,14 @@ def message_bytes(self) -> bytes:
...
         try:
-            cumulative = _parse_base_units(self.cumulative)
+            cumulative = _parse_base_units(self.cumulative_amount)
         except ValueError as exc:
             raise ValueError("invalid voucher cumulative") from exc
-        return voucher_message_bytes(channel, cumulative, self.expires_at)
+        return voucher_message_bytes(
+            channel,
+            cumulative,
+            self.expires_at if self.expires_at is not None else DEFAULT_SESSION_EXPIRES_AT,
```
EfeDurmaz16: Correcting myself: not a pay-kit cross-SDK break, all three clients sign and send the same 4102444800 default. The residual is verify-side only: for a wire voucher omitting expiresAt, Rust and TS reconstruct 0 but this line substitutes 4102444800, so a spec-conforming external no-expiry voucher fails only on Python. Follow-up: encode the wire value verbatim, 0 for omitted.

typescript/packages/mpp/src/Methods.ts
```diff
@@ -207,111 +207,85 @@ export const session = Method.from({
...
+                    idleTimeoutSeconds: z.optional(z.number()),
+                    mint: z.string(),
+                    openSlot: z.string(),
+                    payee: z.string(),
+                    payer: z.string(),
+                    salt: z.string(),
+                    transaction: z.string(),
                 }),
                 z.object({
                     action: z.literal('voucher'),
```
EfeDurmaz16, line 234: Spec says the voucher action carries a top-level channelId and the signed voucher's inner key is voucher; all three SDKs use no top-level channelId and inner key data (close nests the same object). Mutually consistent so not a blocker, but it needs a decision: align the SDKs or pin the spec to the implemented shape, and freeze a voucher-action vector either way so it cannot drift again.

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
```diff
@@ -929,17 +573,76 @@ async def _close_on_idle(self, channel_id: str) -> None:
...
+            due = True
+            return nxt
+
         try:
-            await self._handle_close(ClosePayload(channel_id=channel_id, voucher=None))
+            await self._core.store().update_channel(channel_id, close_if_due)
+            if remaining is not None and self._lifecycle is not None:
+                self._lifecycle.touch(channel_id, remaining)
+            if due:
+                await self._settle_channel(channel_id)
```
EfeDurmaz16, line 611: Two interleavings drop the watchdog timer via the _AlreadySettling None: an idle fire during a client settle, and the losing side of two racing client closes (which also removes the timer and returns success before the winner's broadcast resolves). The channel stays close-pending but re-drivable via retry or restart, so follow-up. Fix shape: only remove the timer once a signature is recorded or the channel is sealed, and re-touch with the retry delay when a settle resolves to None while still close-pending with signer/RPC configured.

rust/crates/kit/src/core/store.rs
```diff
@@ -256,6 +304,27 @@ pub struct ChannelState {
...
+    /// before versioning. Durable stores refuse records newer than
+    /// [`CHANNEL_STATE_SCHEMA_VERSION`] instead of decoding them lossily.
+    #[serde(default)]
+    pub schema_version: u32,
+
+    /// Fields this crate version does not know, preserved verbatim so a
+    /// read-modify-write by an older writer can never strip a newer schema's
+    /// fields off a shared record (the 2026-08-01 proof-binding wipe).
+    #[serde(flatten)]
+    pub extra: serde_json::Map<String, serde_json::Value>,
```
EfeDurmaz16, line 318: The flattened extra map can round-trip a foreign field whose serialized bytes contain "sealed":true, false-positiving the mark_sealed_atomic/mark_finalized_atomic substring guards. Follow-up: JSON-aware check, or namespace extra under a single key.

rust/crates/kit/src/mpp/server/session.rs
```diff
@@ -1202,92 +1639,287 @@ fn unix_now_i64() -> i64 {
...
+    params: &payment_channels::OpenChannelParams,
+) -> Result<()> {
+    let channel_address = payment_channels::derive_channel_addresses(params).channel;
+    let account = rpc
+        .get_account(&channel_address)
+        .map_err(|error| Error::Rpc(format!("fetch confirmed channel failed: {error}")))?;
+    let channel =
+        payment_channels::generated::generated::accounts::Channel::from_bytes(&account.data)
+            .map_err(|error| Error::Other(format!("decode confirmed channel: {error}")))?;
+    if channel.status != 0
```
EfeDurmaz16, line 1772: fetch_and_match_open_channel does not require settlement == 0, so the broadcast-error adoption arm could adopt a channel with a nonzero on-chain settled watermark while fresh_state seeds zero. Hard to reach; adding the check makes adoption airtight.

typescript/packages/mpp/src/server/Session.ts
```diff
@@ -406,203 +496,166 @@ session.routes = function routes(parameters: session.Parameters): session.Routes
...
+            payerSigner: args.feePayer ? args.feePayerSigner : undefined,
+            rpc: args.rpc as SubmitOpenRpc,
+        });
+        signature = submitted.signature as unknown as string;
         return newState;
     });
-    args.lifecycle?.touch(channelId);
+    args.lifecycle?.touch(verified.channelId, persisted.idleTimeoutSeconds);
 
     return Receipt.from({
```
EfeDurmaz16, line 654: Spec marks intent, acceptedCumulative, spent, and idleTimeoutSeconds REQUIRED on session receipts; no SDK emits any of them, so clients cannot recover the accepted watermark from receipts. Additive follow-up across the three SDKs.

typescript/packages/mpp/src/server/Session.ts
```diff
@@ -406,203 +496,166 @@ session.routes = function routes(parameters: session.Parameters): session.Routes
...
+    // allowed, a later one never is), so a payload outside this window was
+    // not built against this challenge. Mirrors `process_open` in the Rust
+    // SessionServer.
+    const { recentBlockhash, recentSlot } = args.openContext;
+    if (openSlot > recentSlot) {
+        throw new Error(
+            `open openSlot ${openSlot.toString()} is ahead of the challenged recentSlot ${recentSlot.toString()}`,
+        );
+    }
+    if (recentSlot - openSlot > OPEN_SLOT_WINDOW) {
```
EfeDurmaz16, line 551: Spec open step 7 also wants openSlot checked against the server's current slot, step 2 wants a non-curve authorizedSigner rejected before fees are paid, and open MUST reject a wire bump field; none are implemented in any SDK. The program fails closed, so the exposure is sponsored fees for guaranteed-to-revert opens. Follow-up.

EfeDurmaz16 (changes requested):
Five previous blockers verified fixed at 2fad1c61, and the schema-version store hardening is a good addition. Two items block this round, both Python retry recovery on the money path and both converged on by every independent review pass: the open verifier (session_onchain.py:499) and the top-up verifier (:581) still treat a broadcast rejection as authoritative, the exact failure class dd545d2d fixed in Rust; TS open shares it. Everything else inline is tracked follow-up material, including the _AlreadySettling timer drop and the voucher-wire spec decisions.

EfeDurmaz16 (approved):
Both blockers are properly fixed at d1f2beed, traced in code: the Python open and top-up verifiers now resolve a preflight-rejected retry against the confirmed channel account and the landed signature status, same shape as the Rust fix, and TS open gained the same rescue. The voucher wire now matches the spec in all three SDKs with voucher and close vectors frozen, and omitted expiresAt encodes as 0. Remaining follow-ups are tracked in the inline threads. Verified against the full suite of green checks.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
lgalabru: Fixed in `08c77fb4` (Python) and `d1f2beed` (TS submitOpenTx — good catch that it shared the naked send).

Python: the broadcast error is captured on both the gasless and direct paths and `_verify_channel_account` still runs against the verified params — a full field match is success, anything else re-raises the original error. TS: the existing post-confirm field match now arbitrates the send/confirm failure the same way, with the transaction's own signature as the result. Both mirror `dd545d2d`'s shape: the confirmed account matches the verified open params only if this exact open succeeded.

Red-tested per your template: the Python mock RPC now rejects duplicate sends like mainnet and the no-stored-state retry test fails with the fix reverted (`test_open_verifier_rescues_landed_open_on_duplicate_preflight_rejection`); TS equivalent in `session-verify-open-tx.test.ts`. Negative cases pin that a non-matching (or missing) confirmed account keeps the broadcast failure authoritative.

python/src/solana_pay_kit/protocols/mpp/server/session_onchain.py
lgalabru: Fixed in `08c77fb4`, mirroring the TS `34f293f2` rescue exactly: on send failure the transaction's own signature status arbitrates — landed clean continues to confirmation and the post-confirm deposit re-check (which, with the credit lost, still expects `state.deposit + amount`), while a landed-but-FAILED or unknown signature re-raises the original broadcast error.

Red-tested: `test_top_up_verifier_rescues_landed_top_up_on_duplicate_preflight_rejection` fails on the pre-fix code (the mock rejects duplicate sends like mainnet), and the two negative tests pin the strictness (`err` set / no status → original error).

python/src/solana_pay_kit/protocols/mpp/intents/session.py
lgalabru: Agreed with your correction — verify-side only, since all three shipped clients emit the explicit 4102444800 default; the exposure was a spec-conforming external client omitting `expiresAt`. Fixed in `dc30ef65`: Python now encodes an omitted `expiresAt` verbatim as 0 into the 50-byte preimage, the `highest_voucher_expires_at` watermark records 0 so settle replays exactly the bytes the signature covers, and the `VoucherData` docstring documents the correct behavior. The remaining `DEFAULT_SESSION_EXPIRES_AT` uses are explicit signed wire values (operator-use path) or internal delivery-directive expiries — audited, none leak into reconstructed signed bytes.

Frozen cross-SDK: `session-voucher-preimage-no-expiry` pins the exact 50 bytes (expiresAt omitted → trailing zero i64) and runs in the TS/Python/Go/Kotlin/Swift runners, driving each SDK's own never-expires default.

typescript/packages/mpp/src/Methods.ts
lgalabru: Ludovic ruled: follow the spec. Done in `7085443f` across all three SDKs — inner key `data` → `voucher` (voucher action and the voucher nested in close), plus the REQUIRED top-level `channelId` routing key, emitted and parsed. Servers reject a voucher action whose top-level `channelId` diverges from the signed voucher's inner one, and a close whose nested final voucher is bound to another channel — red-tested in each SDK (the check removed → exactly the new reject tests fail).

Frozen so it cannot drift again: `session-wire-action-voucher-frozen` and `session-wire-action-close-frozen` (the close vector also freezes the omitted-`expiresAt` form — no key on the wire), plus `session-wire-action-voucher-legacy-inner-data-reject` pinning that the superseded inner-`data` shape now fails parse. Go's seeded skip allowlist carries the three new ids until its wire migration, which inherits this shape (noted in the PR body).

python/src/solana_pay_kit/protocols/mpp/server/session_method.py
lgalabru: Agreed, both interleavings are real and the channel stays re-drivable, so tracking as a follow-up (added to the PR-body list). Your fix shape matches what I'd do: timer removal only once a signature is recorded or the channel is sealed, and a retry-delay re-touch when a settle resolves `None` while still close-pending with signer/RPC configured.

rust/crates/kit/src/core/store.rs
lgalabru: Agreed — the substring guard was always a pragmatic shortcut and the flattened `extra` round-trip gives it a real false-positive path. Tracked as a follow-up (PR-body list); leaning toward the JSON-aware check over namespacing `extra`, since namespacing would break the old-writer round-trip contract we just shipped.

rust/crates/kit/src/mpp/server/session.rs
lgalabru: Agreed — hard to reach (it needs a settled watermark on a channel the store has never seen), but `settlement == 0` makes adoption airtight and is one line. Tracked as a follow-up (PR-body list).

typescript/packages/mpp/src/server/Session.ts
lgalabru: Verified against §Receipt-Format — real, and purely additive since nothing parses receipts strictly yet. Tracked as a follow-up (PR-body list) across the three SDKs rather than folded in: it touches every receipt constructor and deserves its own vector freeze.

typescript/packages/mpp/src/server/Session.ts
lgalabru: Agreed on all three (the step-7 current-slot window was already on our open list; the non-curve `authorizedSigner` reject and the wire `bump` MUST-reject join it). Exposure is bounded to sponsored fees on guaranteed-to-revert opens since the program fails closed, so tracked as a follow-up (PR-body list).

Issue #261, opened 2026-08-01
ci: gate per-language jobs on a changes filter instead of running all 43 checks on every PR


EfeDurmaz16:
## Problem

Every pull request runs all 9 language workflows and all 43 checks, regardless of what changed. Measured on a recent PR: **42 jobs, 112 minutes of runner time, 9.6 minutes wall clock**. Queueing is not the issue (3 minutes total across 42 jobs), the cost is that everything runs.

A concrete example: PR #259 changed 100 files, none of them under `lua/`, and still paid the full Lua job.

## Why a workflow-level `paths:` filter is the wrong shape

Two reasons, both load-bearing:

1. **The harness legs are cross-language by construction.** `harness-lua` runs a TypeScript client and a Rust client against the Lua server. A filter scoped to `lua/**` would skip it when `typescript/` changes, which is exactly the case it exists to catch. A filter wide enough to be correct would cover most of the repo.
2. **`paths:` skips the check entirely.** If required status checks are ever turned on, a skipped workflow never reports and the PR sits in limbo.

## Proposed shape

One `changes` job that always runs, and every other job gated on its outputs:

```yaml
jobs:
  changes:
    runs-on: ubuntu-latest
    outputs:
      lua: ${{ steps.filter.outputs.lua }}
      harness-lua: ${{ steps.filter.outputs.harness-lua }}
    steps:
      - uses: actions/checkout@v5
      - uses: dorny/paths-filter@v3
        id: filter
        with:
          filters: |
            lua:
              - 'lua/**'
              - '.github/workflows/lua.yml'
            harness-lua:
              - 'lua/**'
              - 'harness/**'
              - 'typescript/**'
              - 'rust/**'
              - '.github/actions/setup-harness*/**'

  test-lua:
    needs: changes
    if: needs.changes.outputs.lua == 'true'
```

The job still exists and still reports, so required checks stay satisfiable. The filter map is per job, not per workflow, which is what lets `test-lua` and `harness-lua` have different trigger sets.

## The constraint that must not be broken

**A harness leg's filter must list every language that participates in that leg**, not just the server language. Getting this wrong silently stops testing a cross-SDK pair, which is the failure mode this repo's harness exists to prevent. Concretely:

| leg | must trigger on |
|---|---|
| `harness-lua` | `lua/`, `harness/`, `typescript/`, `rust/` |
| `harness-swift` | `swift/`, `harness/`, `typescript/`, `rust/` |
[...]

PR #264, opened 2026-08-01
perf(ci): thaw the frozen cargo caches

EfeDurmaz16:
Third PR in the CI series (#260 → #262 → this, stacked on #262; see #261 for the investigation index). Everything here was measured before it was changed and verified warm after; no SDK behaviour is touched.

## The bug

Every cargo cache in the repo keys on `hashFiles('rust/Cargo.lock')` — but `Cargo.lock` is gitignored (`.gitignore:20`), so the hash is always the empty string and each key degenerates to a constant:

```
key: Linux-cargo-rust-            ← nothing after the dash, every run
Cache hit occurred on the primary key Linux-cargo-rust-, not saving cache.
```

A constant key means an exact primary hit on every run, and `actions/cache` never re-saves on an exact hit. So each cache froze at whatever its first run ever stored — artifacts from an older rustc that cargo discards wholesale. The result, measured across one green run (branch `perf/lua-coverage-ci`):

| job | restored | then compiled anyway |
|---|---|---|
| Rust tests | 660 MB | **890 crates, 4m55s** |
| Payment links E2E: Rust | ~500 MB | 489 crates, 3m05s |
| harness.yml build job | ~500 MB | 474 crates, ~103s |
| ci.yml TS harness | ~500 MB | 474 crates, ~101s |
| go / php / kotlin setup-harness | ~400 MB | 251 crates, ~70s each |
| Swift ×3 legs (macOS, 10x billing) | 492-557 MB | ~90s each |

Control experiment, locally, same `cargo llvm-cov` invocation as CI:

```
cold:  890 crates compiled   88s
warm:    2 crates compiled   21.6s   (coverage identical, 91.16%)
```

So `cargo-llvm-cov`'s default clean is not the problem (it only cleans workspace crates); the frozen key is, and `--no-clean` is not needed.

## The fixes

**1. Key on rustc version + manifest hash** (`ci.yml` ×2, `setup-harness`, `setup-harness-leg`)

`rustc -V` goes into the key because artifacts are unusable across rustc bumps anyway; `hashFiles('rust/Cargo.lock', 'rust/**/Cargo.toml')` covers dependency changes, and the lockfile pattern starts contributing automatically the moment #216 commits the lockfile — no conflict, just a better key later. `restore-keys` falls back only within the same rustc version: the old frozen caches are dead weight and not worth their download time. `setup-harness` and `setup-harness-leg` compute the identical expression so the harness build job's save stays an exact hit for the eleven legs (the pairing the leg action was already written for — and the legs are the part that works today, zero `Compiling` lines).

**2. Cache the payment-channels `.so`** (`build-payment-channels/action.yml`)

The action builds a **pinned commit** yet paid the Anza CLI curl-install (~16s), the platform-tools v1.52 download (a 45-60s silent gap before the first `Compiling` line), and `cargo build-sbf`, on every run. The output is fully determined by (ref, program-id, this action file), so the artifact is cached on exactly those and install/patch/build are skipped on a hit. The key omits `runner.os` deliberately: SBF bytecode is host-independent, so the ubuntu-built `.so` serves the macOS legs. Nothing downstream needs the Anza CLI on PATH (verified: consumers read the `.so` path from the action output; the only later match for `solana` in those jobs is a comment).

**3. Prebuild the rust x402 client in `ruby.yml`**

The 99.4-second `x402-exact-basic: rust-x402 client pays ruby server` test — 69% of the harness-ruby job — was not a slow test. `ruby.yml` was the only workflow driving a rust client without passing `cargo-bins` to setup-harness, so the in-test `cargo run --quiet` was the job's first cargo invocation ever: a cold compile of the full dependency tree, silent because of `--quiet`, visible in the raw log as a 101-second gap with zero output. Passing `cargo-bins` like every sibling workflow moves the build into the setup step, under the now-working cache.

## Measured effect

Three runs on this branch: a cold run (new keys miss, build, and save for the first time), then a warm run that exposed one bug in this PR's own first draft (below), then the verification warm run. All 43 checks green on the final run. Cold-vs-warm, same code (the warm trigger was an empty commit):

| job | before (frozen cache) | final warm | |
|---|---|---|---|
| Rust tests | 479s | **146s** | 3.3x |
| Swift x402-upto (macOS, 10x billing) | 340s | **107s** | 3.2x |
| Payment links E2E: Rust | 261s | **95s** | 2.7x |
| Go PayKit harness | 274s | **124s** | 2.2x |
| Python harness matrix | 350s | **175s** | 2.0x |
| harness.yml build job | 160s | **80s** | 2.0x |
| Kotlin x402-upto | 242s | **124s** | 2.0x |
| Harness: Ruby | 183s | **110s** | 1.7x |
| Harness: TypeScript (ci.yml) | 255s | **178s** | 1.4x |

Wall clock for the full 43-check suite: **9.6 min → 4.5 min**. Combined with #260 (Lua 20.3 min → ~4.5 min) the critical path of a PR has gone from ~20.5 minutes at the start of this series to 4.5.

**The save race the first warm run caught:** the first draft of this PR gave `Rust tests` and `Payment links E2E: Rust` a shared `cargo-rust-` key. The coverage job builds instrumented artifacts under `target/llvm-cov-target`; the E2E job builds plain ones, finishes first, and wins the save every time — so the coverage job restored artifacts it could not use and recompiled 1100 crates on a run whose cache reported an exact primary hit. The `cargo-rust-cov-` namespace in the last commit fixes it, and 479→117s is the measured result. Worth keeping in mind for any future cache: two jobs may share a key only if they build with identical flags.

One unrelated flake surfaced during measurement (`charge-basic: typescript client pays python server`, "fee payer pubkey not present in transaction accounts") — it failed on one run and passed on the rerun and on both neighbouring runs of identical code; main's history shows the same leg failing sporadically. Not introduced here.

## What this deliberately does not touch

- ci.yml's duplicate "Harness: TypeScript harness" job (255s of repeated setup) — that is a job-topology question, raised on #261 instead.

### DIFF
--- a/.github/actions/build-payment-channels/action.yml
+++ b/.github/actions/build-payment-channels/action.yml
@@ -32,12 +32,28 @@ runs:
         ref: ${{ inputs.ref }}
         path: _payment-channels
         fetch-depth: 1
+    # The build is fully determined by the pinned ref, the declared program
+    # id, and this action file (which carries the sentinel patch and the
+    # pinned platform-tools version), so the resulting .so can be cached on
+    # exactly those and the Anza CLI install (~16s), the platform-tools v1.52
+    # download (~45-60s of silence before the first Compiling line), and the
+    # build-sbf run are all skipped on a hit. No runner.os in the key: the
+    # artifact is SBF bytecode, host-independent, so the ubuntu-built .so
+    # serves the macOS legs too.
+    - name: Cache built program
+      id: so-cache
+      uses: actions/cache@v5
+      with:
+        path: _payment-channels/target/deploy
+        key: payment-channels-so-${{ inputs.ref }}-${{ inputs.program-id }}-${{ hashFiles('.github/actions/build-payment-channels/action.yml') }}
     - name: Install Anza CLI
+      if: steps.so-cache.outputs.cache-hit != 'true'
       shell: bash
       run: |
         sh -c "$(curl -sSfL https://release.anza.xyz/stable/install)"
         echo "$HOME/.local/share/solana/install/active_release/bin" >> "$GITHUB_PATH"
     - name: Patch payment channels treasury owner
+      if: steps.so-cache.outputs.cache-hit != 'true'
       shell: bash
       working-directory: _payment-channels
       run: |
@@ -59,6 +75,7 @@ runs:
           exit 1
         fi
     - name: Build payment channels program
+      if: steps.so-cache.outputs.cache-hit != 'true'
       shell: bash
       working-directory: _payment-channels/program/payment_channels
       run: cargo build-sbf --tools-version v1.52 --arch v1
--- a/.github/actions/setup-harness-leg/action.yml
+++ b/.github/actions/setup-harness-leg/action.yml
@@ -26,6 +26,13 @@ runs:
     # miss this cache — different runner.os — and compile Rust themselves, as
     # they do today.)
     - uses: dtolnay/rust-toolchain@stable
+    # Same key expression as setup-harness (see the comment there for why the
+    # gitignored-lockfile hash froze this cache): the build job saves under
+    # this key, the legs hit it exactly.
+    - name: Compute cargo cache key
+      id: cargo-key
+      shell: bash
+      run: echo "rustc=$(rustc -V | tr -d ' ()')" >> "$GITHUB_OUTPUT"
     - name: Restore shared cargo cache
       uses: actions/cache@v5
       with:
@@ -33,8 +40,8 @@ runs:
           ~/.cargo/registry
           ~/.cargo/git
           rust/target
-        key: ${{ runner.os }}-cargo-shared-harness-${{ hashFiles('rust/Cargo.lock') }}
-        restore-keys: ${{ runner.os }}-cargo-shared-harness-
+        key: ${{ runner.os }}-cargo-shared-harness-${{ steps.cargo-key.outputs.rustc }}-${{ hashFiles('rust/Cargo.lock', 'rust/**/Cargo.toml') }}
+        restore-keys: ${{ runner.os }}-cargo-shared-harness-${{ steps.cargo-key.outputs.rustc }}-
     - name: Install harness
       working-directory: harness
       shell: bash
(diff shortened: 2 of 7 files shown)


EfeDurmaz16:
Visual summary of the whole series, measured on real runs (final warm run, 43/43 green).

**Wall clock for a full PR check suite:**

```mermaid
flowchart LR
    A["20.5 min<br/>start of series<br/>Lua coverage step: 20 min"] --> B["9.6 min<br/>after #260<br/>cluacov + memoized ATA"]
    B --> C["5.7 min<br/>this PR, first draft<br/>thawed keys + .so cache"]
    C --> D["4.5 min<br/>final<br/>cov namespace + go/py on the action"]
    style A fill:#fae1d7,stroke:#d85a30,color:#4a1b0c
    style B fill:#faeeda,stroke:#eda100,color:#412402
    style C fill:#e6f1fb,stroke:#378add,color:#042c53
    style D fill:#e1f5ee,stroke:#1d9e75,color:#04342c
```

(#262 in the stack additionally cut the Python unit step 182s → 59s and made superseded PR runs cancel — a compute saving rather than a wall-clock one.)

**Per-job, before the series vs the final warm run (seconds):**

```mermaid
xychart-beta horizontal
    title "CI job duration: before (outer) vs after (inner)"
    x-axis ["Lua tests (cov)", "Rust tests", "Python harness", "Swift x402-upto", "Go harness", "E2E Rust", "TS harness", "Kotlin x402-upto", "Python tests", "Ruby harness"]
    y-axis "seconds" 0 --> 1250
    bar [1198, 479, 350, 340, 274, 261, 255, 242, 218, 183]
    bar [235, 146, 175, 107, 124, 95, 178, 124, 80, 110]
```

| job | before | after | |
|---|---|---|---|
| Lua tests (coverage step, #260) | 1198s | 235s | 5.1x |
| Rust tests | 479s | 146s | 3.3x |
| Swift x402-upto (macOS, 10x billing) | 340s | 107s | 3.2x |
| Payment links E2E: Rust | 261s | 95s | 2.7x |
| Go PayKit harness | 274s | 124s | 2.2x |
| Python harness matrix | 350s | 175s | 2.0x |
| Kotlin x402-upto | 242s | 124s | 2.0x |
| Python tests (#262) | 218s | 80s | 2.7x |
| Ruby harness | 183s | 110s | 1.7x |
| TS harness (ci.yml) | 255s | 178s | 1.4x |

The distribution is now flat: the critical path is Lua at 4.4 min with four jobs in the 175-180s band right behind it, so further wall-clock wins come from the changes-filter design in #261 rather than from any single job.

PR #265, opened 2026-08-02
fix(rust/x402)!: fetch blockhash + openSlot via RPC when upto challenge omits hints

lgalabru:
## What

The x402 `upto` client hard-required `extra.recentBlockhash` and `extra.recentSlot`, rejecting spec-valid challenges that omit them — the spec marks both hints optional, and the reference TS server (phdargen `svm-upto`, under review as solana-foundation/x402#5) only embeds them when `SVM_RPC_URL` is set. This was interop blocker #1 for `pay curl` against that implementation.

`build_upto_payload` / `build_upto_header` now take an `RpcClient` (mirroring `x402::client::exact`) and fall back to a single `getLatestBlockhash` call when either hint is absent — the response context carries the current slot, so one round-trip covers both `recentBlockhash` and `openSlot`. Embedded hints still win, and the RPC client is untouched when both are present.

## Notes

- **Breaking** (signature change): callers pass an `RpcClient`, same shape as the `exact` client. Kit consumers construct it lazily — no connection is made unless the fallback fires. Suggest folding into the next breaking bump (0.6.0) alongside the rest of the svm-upto cascade.
- The blocking `RpcClient` uses `block_in_place` + its own runtime internally, so the fallback is safe under a multi-thread tokio runtime — the same invariant the `exact` client already relies on. `pay` drives this path via `block_on` on a `new_multi_thread` runtime, so it only needs to thread its existing `rpc_url` through (it already resolves one for Surfpool auto-fund).
- Hint semantics when exactly one is present: the present hint wins, the fetch fills the missing one. Server blockhash + fetched current slot (or vice versa) are independently valid — the blockhash gates tx recency, the slot seeds the channel PDA window.
- Harness client bin updated (it already constructed an unused `RpcClient` for exactly this).

## Testing

- 4 new tests: hints-present builds without touching RPC (proven with a fail-loud mock), no-hints fallback, slot-only-missing fallback, no-hints + RPC-down errors as `Error::Rpc`.
- Kit lib 839/839 (`--features server,client`), full workspace green, clippy clean vs main, fmt clean.

## Out of scope (remaining svm-upto cascade items)

Strict `Vec<UptoRequirements>` accepts parsing (mixed EVM+SVM envelope kills upto detection), `voucherSignature` wire types, `challenged_slot=None` fail-open, settlement-readiness simulation, auto-refund on handler failure — tracked from the svm-upto branch evaluation.

### DIFF
--- a/rust/crates/harness-bins/src/bin/x402_harness_upto_client.rs
+++ b/rust/crates/harness-bins/src/bin/x402_harness_upto_client.rs
@@ -40,12 +40,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     let requirements = parse_upto_challenge(&first_headers, Some(&first_body))
         .ok_or_else(|| "server did not return a supported x402 upto challenge".to_string())?;
 
-    let _rpc = RpcClient::new(rpc_url);
+    let rpc = RpcClient::new(rpc_url);
     let expires_at = now_unix() + 3600;
     let nonce = format!("upto-{}", now_unix());
-    // `openSlot` (like `recentBlockhash`) rides in the challenge requirements;
-    // the client never fetches its own slot.
-    let payment_header = build_upto_header(&signer, &requirements, expires_at, nonce).await?;
+    // `openSlot` (like `recentBlockhash`) normally rides in the challenge
+    // requirements; the RPC client only fills in whichever hint is absent.
+    let payment_header = build_upto_header(&signer, &rpc, &requirements, expires_at, nonce).await?;
 
     let paid_response = http
         .get(&target_url)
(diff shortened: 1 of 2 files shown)


EfeDurmaz16 (approved):
Clean spec alignment: embedded hints still take precedence with zero RPC traffic (proven by the failing-mock test), a single getLatestBlockhash covers both missing values via the response context, the mixed case keeps the hinted blockhash, and an absent-hints RPC failure fails closed with Error::Rpc rather than degrading, which also matches the fail-closed direction from the pay-kit#259 follow-ups. All four paths are unit-tested and the harness client is updated for the breaking signature. Verified by reading both changed files at 91873441; CI is green.

PR #281, opened 2026-08-04
feat(mpp/charge): support display-unit amounts via optional `decimals` field

bensandler-stripe:
## Summary

Add a schema-level transform to the solana charge method that optionally converts display-unit amounts to atomic units. This aligns with how tempo and evm methods handle amounts in mppx, enabling a single shared amount across all methods in `compose`.

### Problem

In mppx, standard payment methods (tempo, evm) accept amounts in display units (e.g., `"0.01"` for 1 cent USDC) via a `z.pipe()` + `z.transform()` in their schema that converts to atomic at the input boundary. The solana charge method only accepted atomic units (e.g., `"10000"`), forcing callers to manually convert and preventing uniform amounts in `mppx.compose()`:

```ts
// Before: solana requires atomic, others accept display
["tempo/charge",  { amount: "0.01" }]        // display -> schema converts -> atomic on wire
["evm/charge",    { amount: "0.01" }]        // display -> schema converts -> atomic on wire
["solana/charge", { amount: "10000" }]       // atomic only -- caller must convert
```

### Solution

Same pattern as tempo/evm: add an optional `decimals` field to the charge request input schema, and wrap the schema in `z.pipe()` + `z.transform()`. When `decimals` is present, the transform converts display amounts to atomic at the schema boundary. When absent, amounts pass through unchanged.

```ts
// After: opt into display units by passing decimals
["tempo/charge",  { amount: "0.01" }]                     // existing behavior
["evm/charge",    { amount: "0.01" }]                     // existing behavior
["solana/charge", { amount: "0.01", decimals: 6 }]        // new: display -> atomic at schema boundary
["solana/charge", { amount: "10000" }]                    // still works: no decimals = atomic passthrough
```

### Design

- **Zero breaking change**: `decimals` is optional. Existing callers passing atomic amounts (without `decimals`) are completely unaffected.
- **Schema-level transform**: Conversion happens at the input boundary via `z.pipe()` + `z.transform()`, identical to how tempo (`mppx/src/tempo/Methods.ts`) and evm (`mppx/src/evm/Methods.ts`) handle it.
- **Wire format unchanged**: The challenge always contains atomic amounts. Server verify logic, client transaction building, and all downstream code remain unmodified.
- **Minimal diff**: Only `Methods.ts` is changed (1 file). No changes to `Charge.ts`, client code, or tests.

### Changes

- `typescript/packages/mpp/src/Methods.ts`:
  - Add `parseUnits()` helper (display string -> atomic string)
  - Add optional `decimals` field to charge request schema
  - Wrap request in `z.pipe()` + `z.transform()` that converts amount (and split amounts) when `decimals` is present

### DIFF
--- a/typescript/packages/mpp/src/Methods.ts
+++ b/typescript/packages/mpp/src/Methods.ts
@@ -4,13 +4,24 @@ import { Method, z } from 'mppx';
  * Converts a display-unit amount string to its atomic-unit string representation.
  * Used by the schema transform to convert amounts at the input boundary.
  *
+ * Validates that the input is a well-formed non-negative decimal number.
+ * Rejects negative values, non-numeric strings, and malformed input.
+ *
  * @example parseUnits('0.01', 6) → '10000'
  * @example parseUnits('1.5', 9) → '1500000000'
+ * @example parseUnits('100', 6) → '100000000'
  */
 function parseUnits(value: string, decimals: number): string {
-    const [integer = '0', fraction = ''] = value.split('.');
+    if (!/^[0-9]+(?:\.[0-9]+)?$/.test(value)) {
+        throw new Error(
+            `Invalid amount "${value}": must be a non-negative decimal number (e.g., "0.01", "100")`,
+        );
+    }
+    const [integer, fraction = ''] = value.split('.');
     if (fraction.length > decimals) {
-        throw new Error(`Amount "${value}" has more decimal places than allowed (${decimals})`);
+        throw new Error(
+            `Amount "${value}" has more decimal places than allowed (${decimals})`,
+        );
     }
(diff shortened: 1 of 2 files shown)


bensandler-stripe:
Closing this PR. For the display-unit conversion to be useful with the unified `.charge()` API we're building in mppx (where all methods share a single amount), `decimals` needs to be in the method's `defaults` so the schema transform fires automatically. But adding it to defaults is a breaking change for existing callers who pass atomic amounts. Since a non-breaking opt-in version (caller passes decimals explicitly) doesn't help our use case (the whole point is the caller NOT needing to specify per-method details), this needs to be a breaking change in a future major version of `@solana/mpp`. We'll revisit when the package is ready for a major bump.

PR #283, opened 2026-08-11
fix(mpp): align session SDKs with draft-00

lgalabru:
## Summary
- align Rust, TypeScript, and Python session challenges, receipts, accounting, and open validation with draft-solana-session-00
- add Rust server fee sponsorship for session opens and top-ups, including client-signature validation, bounded compute budgets, server co-signing, and replay-safe transaction IDs
- reject stale or future opens, off-curve signers, native SOL challenges, and client-supplied bump fields
- bump Rust to 0.6.0, TypeScript packages to 0.9.0, and Python to 0.7.0

## Vendored x402 submodule -> upstream `main`
The `typescript/external/x402` submodule now tracks
[`x402-foundation/x402`](https://github.com/x402-foundation/x402) `main`
(7d5363a6) instead of `solana-foundation/x402` `feat/svm-upto-scheme`, and the
vendored `@x402/core` + `@x402/svm` tarballs move 2.16.0 -> 2.23.0.

Our fork's `upto` work is upstream now — server-provided recent blockhash
(x402-foundation/x402#2693), svm upto paymentflow (#3094), solana upto in the Go
SDK (#3141), injectable RPC client in the upto facilitator (#3183) — and
upstream `main` has grown past it (epoch-addressed opens, `ChannelSplit` fee
splits, rent cleanup). The fork branch held nothing upstream lacks, so tracking
it only added a sync burden.

Two upstream API changes reach our adapters:
- `UptoSvmScheme` (facilitator) takes one `FacilitatorSvmSigner` plus config
  instead of two raw signers — wrapped with `toFacilitatorSvmSigner`, matching
  the exact adapter.
- `ExactSvmSchemeOptions` gained `maxPriorityFeeMicroLamports`,
  `maxComputeUnits`, and `maxRequiredSignatures`; `X402Options` exposes all
  three (the drift guard in `x402.test.ts` requires exact parity).

## Cascade: the `upto` open-transaction layout (Rust + Python)
Since x402-foundation/x402#3155 the canonical client builds the channel `open`
as `[SetComputeUnitLimit, SetComputeUnitPrice, open, Memo]`. Our Rust and Python
verifiers required **exactly one** instruction, so every open built by
`@x402/svm >= 2.23` was rejected before the operator co-signed it — including
opens from our own TypeScript client, which uses the upstream scheme.

Both verifiers now enforce the spec layout: an optional ComputeBudget prefix
(`SetComputeUnitLimit` before `SetComputeUnitPrice`, each at most once, within
the 400,000 CU / 5,000,000 microlamport ceilings), exactly one payment-channels
`open`, then at most four optional instructions — up to three Phantom/Solflare
Lighthouse assertions and a Memo. Anything else is still rejected (the operator
co-signs as fee payer, so an unrecognized instruction is one it would blindly
authorize), and no wrapper instruction may name the fee payer among its
accounts. **A bare open stays valid, so every existing client keeps working.**

`extra.memo` joins the `upto` challenge surface: a seller that declares it
requires exactly one matching Memo after `open`, so the Rust and Python clients
emit it (capped at 256 bytes) and stay silent otherwise. Our servers never
declare one, so they accept any memo without demanding a match.

**Follow-up (not in this PR):** the Go, Kotlin, Swift, Ruby, PHP, and Lua `upto`
verifiers still enforce the single-instruction rule and need the same cascade.

## Two fixes the bump surfaced
- **`verifyOpen` now hands the facilitator the challenged `recentSlot`.** Upstream's
  facilitator runs the openSlot window check itself and, without the hint, reads
  `getSlot({commitment: 'finalized'})` — which lags the challenge's
  `getLatestBlockhash` context slot by tens of slots and rejected every fresh
  open. The on-chain settlement harness caught it against the real program;
  `verifyOpen` re-observes the slot once and both checks now judge
  `payload.openSlot` in the same frame.
- **The Python conformance runner rejects a malformed session idle timeout.**
  `_receipt_from_canonical` coerced a boolean/non-integer/negative
  `idleTimeoutSeconds` to `None`; `format_receipt` omits a `None` idle timeout and
  `parse_receipt` requires one on a session receipt, so `receipt.format` answered
  with a header our own parser refuses. It now applies the parser's rule and
  fails the op.

## Test Plan
- Rust: `cargo test --features server,client,axum --lib` — 910 passed;
  `cargo fmt --check` clean (also fixes the pre-existing `batch_verify.rs`
  failure on this branch); local `cargo llvm-cov` line coverage 90.06% vs
  89.98% on the branch tip before these commits
- TypeScript: install, build, typecheck, lint, format:check, and the full test
  suite against vendored 2.23.0 — 521 passed
- On-chain: `harness pnpm test:onchain` against a local surfpool mainnet fork
  — 4 passed (fails on the real program without the `recentSlot` fix)
- Python: Ruff and Pyright clean; full suite — 1,239 passed, 1 Surfnet-only
  skip; coverage gate 90.75%


typescript/packages/mpp/package.json
```diff
@@ -31,6 +31,7 @@
     "typecheck": "tsc --noEmit"
   },
```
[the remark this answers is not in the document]
lgalabru: Checked GitHub's own verification API for both this commit and the current HEAD (d053ba45) — both show `verified: true`, `reason: valid` (SSH signature). Nothing to fix here, the commit is signed.

python/src/solana_pay_kit/protocols/mpp/core/headers.py
```diff
@@ -229,14 +229,41 @@ def parse_receipt(header: str) -> Receipt:
...
         timestamp=timestamp,
         reference=str(data["reference"]),
         challenge_id=str(data.get("challengeId", "")),
         external_id=str(data.get("externalId", "")),
+        intent=str(data.get("intent", "")),
+        accepted_cumulative=str(data.get("acceptedCumulative", "")),
+        spent=str(data.get("spent", "")),
+        idle_timeout_seconds=idle_timeout_seconds,
+        tx_hash=str(data.get("txHash", "")),
+        refunded=str(data.get("refunded", "")),
```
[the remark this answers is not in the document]
lgalabru: Fixed in d053ba45: `_receipt_to_canonical`/`_receipt_from_canonical` now round-trip `intent`, `acceptedCumulative`, `spent`, `idleTimeoutSeconds`, `txHash`, `refunded`. Added a subprocess-based regression test (`test_protocol_conformance_runner.py`) driving the runner's real stdin/stdout ABI with a session receipt — verified it fails without the fix. Left the vendored `harness/vectors/mpp-protocol/` vectors alone since those are upstream mpp-tools fixtures for the base IETF scheme, not the Solana session extension.

EfeDurmaz16 (approved):
Open validation, mint-on-wire, bump reject, and session receipt extensions look sound across Rust/Python/TypeScript. CI is green.

- rust/crates/kit/src/mpp/server/session.rs:873 - client voucher path still leaves spent_amount alone (charged/replay model); Python/TS burn challenge amount into spent on every voucher including replay. Receipt spent and replay economics will diverge until the spine matches or the other SDKs document the split.
- typescript/packages/mpp/src/server/Session.ts:683 - open/topUp put the action signature in txHash; Rust/Python only attach settle signatures. Soft field, but clients that treat txHash as settle proof will misread TS receipts.

Verified: full PR diff line-by-line (36 files); adversarial slices on spent accounting, open validation, receipt wire; CI checks all pass including Rust/Python/TS tests and harness matrix.

EfeDurmaz16 (changes requested):
Requesting changes on the spent / voucher accounting. draft-solana-session-00 already picks one model; the three SDKs do not all implement it.

Spec (paymentauth.org / tempoxyz mpp-specs `draft-solana-session-00`):
- `spentAmount` = amount charged for delivered service
- client available: `acceptedCumulative - spentAmount`
- Voucher Update: persist `acceptedCumulative`, then debit `spentAmount += cost`, then 200 + receipt
- idempotent replay MAY return a cached receipt and MUST NOT change channel state or increment `spent` again
- bare replay (`cumulativeAmount <= acceptedCumulative`, no matching idempotent cache) MUST be rejected

Current PR:
- Python/TS debit `cost` on voucher (good direction) but accept the same voucher again and burn `spent` again (TS test: 100 → 200 → reject). That violates the idempotency / bare-replay rules.
- Rust `verify_voucher` / `process_commit` never bump `spent_amount` (only `process_use`). That skips the required debit on Voucher Update / commit.
- Same client against Rust vs Py/TS gets different receipts and accept/reject. Fix before merge: one path that matches the draft in all three.

Also soft but real:
- `typescript/.../Session.ts` puts open/topUp signatures in `txHash`; the receipt table defines `txHash` as the close settlement signature only.
- `session.routes()` still wires symbol `currency` into deliveries while challenges use the mint.

Open hardening, mint-on-wire, bump reject, and receipt codecs look sound. CI is green. Accounting is the merge gate.

lgalabru:
@EfeDurmaz16 pushed b382dd6a addressing the accounting gate:

- **Rust `verify_voucher`**: never persisted `spent_amount` for client-signed vouchers — only `cumulative` moved. Now debits `charged` (0 on idempotent replay) atomically alongside the watermark.
- **Rust `process_commit`**: never touched `spent_amount` at all. Now debits `actual_amount` on the fresh-commit path (the replay path stays untouched, as before).
- **TS/Python `handleVoucher`/`verify_voucher`**: re-debited `spent` on every replay of the already-accepted highest voucher, so replaying doubled the charge and a second replay wrongly rejected with "insufficient authorized voucher availability". Replay now returns the cached `spent`/`acceptedCumulative` with no debit, matching the idempotent-replay rule in draft-solana-session-00.
- **TS open/topUp receipts**: dropped `txHash` — the spec reserves it for the close settlement signature; Rust/Python already only set it there.
- **TS `session.routes()`**: now resolves `currency` to the mint before wiring it into delivery directives, matching what challenges advertise via `currency: resolvedMint`.

Updated the two existing tests that encoded the old (buggy) double-debit/reject-on-replay behavior, and added `spent_amount` assertions to the Rust voucher/commit tests.

Verified: `cargo test -q` (845 passed) + scoped MPP clippy + fmt; TS lint/typecheck/build/vitest (432 passed in `mpp`); Python pytest (1236 passed) + ruff + pyright.

rust/crates/kit/src/mpp/server/session.rs
```diff
@@ -862,6 +1005,10 @@ impl<S: ChannelStore> SessionServer<S> {
                             "Channel disappeared after voucher acceptance".to_string(),
                         )
                     })?;
+                    state.spent_amount =
+                        state.spent_amount.checked_add(charged).ok_or_else(|| {
+                            StoreError::Internal("session spent amount overflow".to_string())
+                        })?;
```
EfeDurmaz16, line 1072: Follow-up: draft says `spentAmount += cost`. TS/Python debit the challenge cost. This still adds `charged` (the voucher jump). Same client can get a different balance. Also fold this into the same store write as the watermark.

python/protocol_conformance_runner.py
```diff
@@ -137,6 +137,12 @@ def _receipt_to_canonical(rcpt: Receipt) -> dict[str, Any]:
         "reference": rcpt.reference,
         "challengeId": rcpt.challenge_id,
         "externalId": rcpt.external_id,
+        "intent": rcpt.intent,
+        "acceptedCumulative": rcpt.accepted_cumulative,
+        "spent": rcpt.spent,
```
EfeDurmaz16, line 142: Follow-up: nice that these round-trip here. Still need a cross-SDK harness case for spent + exact replay + bare reject on Rust, TS, and Python. Unit tests alone are not enough to call draft-00.

rust/crates/kit/src/mpp/server/session.rs
```diff
@@ -1738,7 +1908,37 @@ async fn verify_submit_and_fetch_open(
         ));
     }
 
+    verify_client_transaction_signatures(&tx, fee_payer_signer.map(|signer| signer.pubkey()))?;
+    if let Some(signer) = fee_payer_signer {
+        payment_channels::cosign_fee_payer(signer, &params.rent_payer, &mut tx).await?;
+    }
```
EfeDurmaz16, line 1976: Follow-up: when the channel already exists, this still co-signs and rebroadcasts. A failed open with a matching channel can burn sponsor fees. Skip co-sign/broadcast on existing channel, or dedupe like top-up.

EfeDurmaz16 (dismissed):
I'm approving with follow-ups. It is not full draft-00 yet.

`b382dd6` fixes the double-debit I blocked on. Open checks, bump reject, mint-on-wire, and the `txHash` / routes fixes look good. CI is green.

Still open before calling this draft-00:

1. **Rust spend model** - draft says `spent += cost`. TS/Python do that. Rust still spends the voucher jump. Same client, different balance. Persist accepted + spent in one write.
2. **Sponsored open replay** - Rust still co-signs/rebroadcasts when the channel already exists. Can burn fees. Skip or dedupe like top-up.
3. **Harness** - add one cross-SDK case for spent + exact replay + bare reject. Unit tests are not enough.

Soft later: Python empty `tx_hash`, looser receipt parse, commit `spent` asserts in Py/TS.

lgalabru:
@EfeDurmaz16 thanks for the approve + follow-ups.

**Item 1 (Rust spend model) — fixed in e77ecc0e.** You're right, and I verified it concretely: my earlier fix debited `acceptance.charged` (the voucher's own cumulative jump), while TS/Python debit the fixed per-action `price`/`config.amount` (spec: `spentAmount += cost`, where `cost` comes from the challenged amount, not the voucher's delta). A voucher pre-funding more than one action's worth of `cumulativeAmount` would silently spend more in Rust than in TS/Python for identical wire traffic. Rust now debits `config.amount` — same source `process_use` already uses — with 0 on replay. Test updated (`spent_amount == 25`, the fixture's `config.amount`, not the voucher's 100-unit jump).

**Item 2 (sponsored-open replay) — pushed back, wanted to flag before changing it.** I traced `verify_submit_and_fetch_open` and compared it against Rust's own `process_topup`: top-up does **not** skip-before-submit either — it always calls `verify_submit_and_fetch_topup` and only dedupes the *state credit* inside the atomic mutator afterward (there's an explicit comment there: "A replay may therefore reach RPC again; the confirmed signature status makes that path idempotent before this mutator credits the deposit"). So open and top-up are already consistent with each other in Rust — "dedupe like top-up" doesn't hold as a fix, since top-up has the identical pattern.

What *is* real: TS's `handleOpen`/`handleTopUp` check the local record first and skip the RPC call entirely on a matching replay, where Rust always re-verifies against the RPC before reconciling local state. That's a real perf difference, not a fee-burn one — resubmitting a byte-identical already-landed transaction gets "already processed" back with no second charge (deterministic signatures ⇒ identical tx hash). Rust's current shape is arguably the more conservative one (re-derive truth from chain every time rather than trust a local cache), so swapping it for TS's short-circuit is a real design tradeoff, not a bug fix — want to loop in @ludo before I change it, since it trades paranoia for round-trips on both `open` and `topUp` and I'd want the harness (item 3) to cover the tradeoff either way.

**Item 3 (cross-SDK harness)** — agreed this is the real gap; scoping it now, will follow up separately.

Soft items (Python `tx_hash`, looser receipt parse, commit-`spent` asserts in Py/TS) — noted, will fold into the harness pass rather than one-off patches.

EfeDurmaz16 (dismissed):
Item 1 looks good. Fine leaving 2 and 3 for follow-up.

rust/crates/kit/src/mpp/server/session.rs
lgalabru: Fixed in `e77ecc0e`: `verify_voucher` now debits `config.amount` (the fixed per-action price), not the voucher's raw cumulative jump — matches `process_use` and the TS/Python session servers. Verified: the existing idempotent-replay test now asserts `spent_amount == config.amount` per fresh acceptance, replay stays 0.

rust/crates/kit/src/mpp/server/session.rs
lgalabru: Fixed in `86bee631`: for a non-fresh open, check the confirmed on-chain channel account first — if it already matches this exact open, return the already-landed signature instead of co-signing and rebroadcasting. Added a regression test asserting the mock RPC sees exactly one `sendTransaction` call across a fresh open + its replay.

python/protocol_conformance_runner.py
lgalabru: Agreed, and not done yet — the receipt-field round-trip fix (`d053ba45`) only closes the codec gap Greptile flagged, it doesn't add the harness scenario you're asking for (spent + exact replay + bare reject exercised cross-language on Rust/TS/Python). Leaving this open as real remaining work rather than resolving it; will pick up as a follow-up.

rust/crates/kit/src/mpp/server/session.rs
```diff
@@ -862,6 +1013,10 @@ impl<S: ChannelStore> SessionServer<S> {
                             "Channel disappeared after voucher acceptance".to_string(),
                         )
                     })?;
+                    state.spent_amount =
```
EfeDurmaz16, line 1069: Spec rule ([draft-solana-session-00, Debit Processing](https://github.com/tempoxyz/mpp-specs/blob/main/specs/methods/solana/draft-solana-session-00.md#debit-processing)): available = acceptedCumulative - spentAmount, and if available < cost the server returns 402 instead of serving. [TS](https://github.com/solana-foundation/pay-kit/blob/055b0c8dc0f775668fc7a537a23a5d26f32d93fa/typescript/packages/mpp/src/server/Session.ts#L840-L841) and [Python](https://github.com/solana-foundation/pay-kit/blob/055b0c8dc0f775668fc7a537a23a5d26f32d93fa/python/src/solana_pay_kit/protocols/mpp/server/session.py#L795-L796) implement that reject in this PR; this mutator debits config.amount with no check, so with min_voucher_delta 0 a cumulative +1 voucher buys a full-price serve and close settles only cumulative. Add the reject here before the debit, plus a delta-1 test; the current tests only cover the over-funded direction.

EfeDurmaz16 (changes requested):
e77ecc0e closed the debit-source half of the accounting item; the availability gate is still missing in Rust, one inline comment on it. The TS/Python sponsored-open gaps (unbounded compute budget, TS non-open instruction pass-through) are pre-existing; leaving them to the follow-up PR with the harness item.

rust/crates/kit/src/mpp/server/session.rs
lgalabru: Fixed in `d2114598`.

I enforced the availability gate one level down, inside `accept_voucher`'s atomic closure (the same lock that advances the watermark), rather than in this debit closure. Reason: if the reject fired *after* `accept_voucher` had already advanced the watermark + highest-voucher signature, a retry of that same under-funded voucher would be classified as an idempotent replay and served for free — reopening the same hole from the other side. Gating before the watermark commits makes a rejection leave no trace, matching how the TS/Python single-mutator paths roll back.

Concretely: `accept_voucher` gains a `required_availability` param; the session server passes `config.amount`, and on a fresh acceptance it rejects with `insufficient authorized voucher availability` when `acceptedCumulative - spentAmount < cost` (draft-solana-session-00 Debit Processing). x402 batch settlement passes `0` (it meters via the cumulative watermark, no `spentAmount` debit, so no per-action floor) — behavior there is unchanged.

Test `voucher_below_action_price_is_rejected_and_not_replayable` covers your delta-1 case: with `min_voucher_delta 0` and one prior serve (spent==cumulative==25), a +1 voucher is rejected, the watermark/spent do **not** advance, an immediate retry of the same voucher is rejected again (not served as a replay), and a subsequent voucher that authorizes a full action serves normally.

The cross-SDK harness case (spent + exact replay + bare reject on all three) is still the agreed follow-up in the harness PR — not in this commit.

lgalabru:
Added confirmed-commitment RPC clients for session challenges, opens, and top-ups in dc32a090. Validation: 880/880 solana-pay-kit library tests pass with the server and client features; formatting passes. Pay #435 is pinned to this head and its 34/34 benchmark tests pass.

rust/crates/kit/src/core/batch_verify.rs
```diff
@@ -0,0 +1,168 @@
...
+        // the stricter check for the degenerate case).
+        let job = batch.into_iter().next().expect("len == 1");
+        let ok = job.key.verify_strict(&job.message, &job.signature).is_ok();
+        let _ = job.reply.send(ok);
+        return;
+    }
+    let messages: Vec<&[u8]> = batch.iter().map(|j| j.message.as_slice()).collect();
+    let signatures: Vec<Signature> = batch.iter().map(|j| j.signature).collect();
+    let keys: Vec<VerifyingKey> = batch.iter().map(|j| j.key).collect();
+    if ed25519_dalek::verify_batch(&messages, &signatures, &keys).is_ok() {
```
EfeDurmaz16: `verify_batch` only checks the batch equation. It skips the extra rejections `verify_strict` makes, so once a batch holds two or more jobs the accept set here is wider than the inline path's, on a money decision.

This needs no unusual key. With an ordinary keypair, `R = identity` and `s = H(R||A||M)*a` satisfies the batch equation. Ran against ed25519-dalek 2.2.0:

```
attacker key is_weak (small order)? false
inline path  verify_strict(forged) accepted = false
batch path   verify_batch([honest, forged]) accepted = true
```

`core/session.rs:138` routes every fresh voucher through the batch verifier when the flag is on, and `verify_and_reply` falls back to `verify_strict` only when the equation fails, so a batch-passing forgery is never rechecked. The watermark advances and the server serves. Settlement goes through the Ed25519 precompile and `solana-signature` verifies with `verify_strict` (lib.rs:78), so the voucher never settles: service given away, nothing collectible. The batch-of-one `verify_strict` at line 111 only helps at low load, which is the load where this mode does nothing.

Suggest gating the batch inputs once in `voucher_verify_parts` the way `solana-signature` does (canonical s, R decompresses, R not small order, key not weak). Free on the inline path. The module also has no tests.

rust/crates/kit/src/mpp/server/session.rs
```diff
@@ -1881,27 +2170,34 @@ async fn verify_submit_and_fetch_topup(
...
+    let submission = pipeline.submit_verified(&tx).await;
+    let min_context_slot = submission.as_ref().ok().map(|confirmed| confirmed.slot);
+    let account_data = pipeline
+        .read_account_data(channel, min_context_slot)
+        .await
+        .map_err(|error| Error::Rpc(format!("fetch topped-up channel failed: {error}")))?
+        .ok_or_else(|| Error::Rpc("confirmed topped-up channel account not found".to_string()));
+    let account_data = match (submission, account_data) {
+        (Ok(_), account) => account?,
+        (Err(_), Ok(account)) => account,
```
EfeDurmaz16: This dropped the `get_signature_status(sig) == Ok(Some(Ok(())))` check the old code ran when broadcast failed (still on main at session.rs:1896). On the `(Err(_), Ok(account))` arm the only gate left is `deposit >= existing.deposit + amount`, and the two sides use different baselines: `existing` is read at 1274 before submission, the mutator at 1333 adds `additionalAmount` to whatever the deposit is under the store lock.

Concurrent case: stored and on-chain deposit both 1000, payer holds 100. Send N top-ups of 100 with different blockhashes so the signatures differ and the dedupe does not fire. One lands, on-chain goes to 1100. The rest fail for insufficient funds, each takes this arm, each reads 1100, each compares against its own baseline of 1000, and 1100 >= 1100 passes for all of them. Stored deposit ends at 1000 + N*100 against 1100 of real escrow, and vouchers are honored up to the stored cap.

Either restore the confirmed-signature check before this arm, or clamp the new deposit to the observed on-chain value inside the mutator instead of adding blindly.

typescript/packages/pay-kit/src/adapters/x402-upto.ts
```diff
@@ -175,28 +176,48 @@ export class X402Upto {
...
+        // and then handed to the facilitator, which runs the same window check
+        // inside `verifyOpenTransaction`. Without the hint the facilitator
+        // falls back to `getSlot({commitment: 'finalized'})`, which lags the
+        // challenge's own read by tens of slots and rejects a fresh open.
+        const recentSlot = await this.#observeRecentSlot();
+        this.#assertOpenSlotBoundToChallenge(payload, recentSlot);
 
-        const requirements = this.#requirements(maxPrice);
+        const requirements = this.#requirements(maxPrice, recentSlot);
         const verification = await this.#facilitator.verify(payload, requirements);
```
EfeDurmaz16: In @x402/svm 2.16 the facilitator's `verify()` called `broadcastOpen` (facilitator/index.js:1070). In 2.23 `verify()` is documented as "Read-only preflight: validate the open authorization without broadcasting" and the broadcast moved to `settle()` then `settleDeposit()`. We are pinned to 2.23.0 in package.json, the pnpm-workspace catalog, and the submodule.

This adapter still assumes the 2.16 contract: `verifyOpen()`'s doc says it broadcasts the open, and `settle()` skips `facilitator.settle()` entirely and submits settle-and-seal directly against `payload.channelId`. So the channel is never opened on chain and settlement targets a PDA that does not exist.

CI stays green because the three upto harness jobs all run against the Rust server, and the new adapter test asserts `rejects.toThrow()`, so nothing exercises a successful open.

EfeDurmaz16 (changes requested):
Re-reviewed at 969b28f3. The availability gate is fixed, and moving it inside `accept_voucher`'s closure is the better call: a rejection now leaves the watermark untouched, so the retry cannot come back as a paid replay.

Three blockers in the new work.

lgalabru:
@EfeDurmaz16 all three fixed, pushed at 91495ce2.

**`typescript/packages/pay-kit/src/adapters/x402-upto.ts` — confirmed and fixed (56d9fb80).** You're exactly right about the 2.16 → 2.23 contract change. I confirmed it empirically before fixing: instrumented `submitSettleAndDistribute` and ran the on-chain harness against a real surfpool fork — `exists=false` on the channel account right before building `settle_and_seal`, on every "successful" (200) `upto` payment. `verifyOpen()` now calls `facilitator.settle()` with deposit-shaped requirements (the actual open broadcast in 2.23); `settle()` builds a claim payload (`voucherSignature` spread onto the wire payload, the discriminator `UptoSvmScheme.settle` uses) and calls `facilitator.settle()` again instead of our own hand-rolled `submitSettleAndDistribute` — so the vendored facilitator's simulate-before-broadcast, dedup cache, and channel storage all run instead of being bypassed. Added a new on-chain e2e test asserting the operator's USDC balance actually moves by the exact billed amount (not just HTTP 200) — verified it fails hard (0 transferred) against the pre-fix adapter.

**`rust/crates/kit/src/core/batch_verify.rs` — confirmed and fixed (db67994a).** Reproduced your PoC exactly (`R = identity`, `s = H(R‖A‖M)·a mod L`) in a test — `verify_batch` accepts it, `verify_strict` rejects it, matching your numbers. Added `is_degenerate` as a pre-filter (small-order `R` or key, the same two checks `verify_strict` makes beyond the batch equation) before the batch equation runs, gated inline as you suggested — free on the single-signature path. Six new tests including an end-to-end one through the public `BatchVerifier` API: a forged signature riding a batch alongside a real one is rejected while the real one still verifies.

**`rust/crates/kit/src/mpp/server/session.rs` — confirmed and fixed (91495ce2).** Went with restoring the confirmed-signature check rather than clamping in the mutator: `submit_verified` already does the per-signature check internally (`TxPipeline::confirm` via `getSignatureStatuses`) before returning `Err`, so the fix is just to stop second-guessing it with the aggregate-deposit fallback — `Err` now fails the top-up immediately, full stop. New test (`topup_whose_own_transaction_fails_is_not_credited_from_the_aggregate_deposit`) simulates a channel whose on-chain deposit already looks topped-up (a concurrent top-up) while THIS request's own signature reports an on-chain failure; verified it fails against the pre-fix code (stored deposit inflates to 1,100 despite the transaction never landing).

Full local verification: 923 Rust tests (was 917 + the new ones) + fmt clean, coverage 90.40%; TypeScript 521 tests + lint/format/typecheck clean; on-chain harness 5/5 against a real surfpool fork. CI run in progress on 91495ce2.

EfeDurmaz16 (approved):
All three fixed. Verified each one locally at 91495ce2 rather than taking the summary.

- `batch_verify.rs`: `is_degenerate` covers exactly the two checks `verify_strict` makes beyond the batch equation (weak key, R small order or not decompressable), and it runs before the batch equation, so a forgery cannot ride a passing batch. Re-ran my PoC case: caught. 6/6 new tests pass.
- `session.rs`: the `(Err(_), Ok(account))` fallback is gone, so an unconfirmed top-up now fails instead of borrowing another request's deposit, and the read-back floors on `confirmed.slot`. `topup_whose_own_transaction_fails_is_not_credited_from_the_aggregate_deposit` passes.
- `x402-upto.ts`: `verifyOpen` now goes through `facilitator.settle()`, which is the path that actually broadcasts in 2.23, and `settle()` hands the claim payload back to the facilitator instead of the hand-rolled `submitSettleAndDistribute`. 19/19 pass.

Nice call adding the on-chain balance assertion to the upto e2e test. HTTP 200 was the reason this went unnoticed.

Verified: `cargo test -p solana-pay-kit --features server,client --lib` 898 passed 0 failed; `vitest` x402 suite 19 passed; CI 42/43 green with only the slow Lua job outstanding and no failures.

PR #287, opened 2026-08-14
docs: rename payment channel -> tab in prose

rishinsharma:
Prose and doc-comment rename as part of the org-wide **Payment Channels → Tabs** change.

**No language binding breaks.** Every public API symbol, module path, and file path is untouched — `payment_channels`, `PaymentChannel*`, `paymentChannels`, `PAYMENT_CHANNELS`, `idl/payment-channels.json`, `_payment-channels`, and `build-payment-channels` all have byte-identical occurrence counts before and after.

Error-message strings and their test assertions moved together (Go, Python, TypeScript), including the playground SSE fixture `"A payment channel "` → `"A tab "`. Workflow step `id:`s that other steps reference by name were kept (`id: payment-channels` in kotlin.yml/swift.yml still matches `steps.payment-channels.outputs.program-so`).

### Not done here, deliberately
The CI coupling to the program repo stays on the old paths: the `build-payment-channels` action and `payment-channels-pull-idl` recipe read `program/payment_channels/idl/payment_channels.json` and `payment_channels.so` from the **pinned ref `0c07d575`**, where those paths still exist — so CI keeps passing today. When `payment_channels_ref` is next bumped past the rename PR (solana-foundation/payment-channels#82), the same change must switch those paths to `program/tabs/idl/tabs.json` / `tabs.so`.

### DIFF
--- a/.github/actions/build-payment-channels/action.yml
+++ b/.github/actions/build-payment-channels/action.yml
@@ -1,6 +1,6 @@
-name: Build payment-channels program
+name: Build tabs program
 description: >
-  Check out the canonical payment-channels program, patch the localnet
+  Check out the canonical tabs program, patch the localnet
   treasury-owner sentinel, and build the SBF artifact the embedded surfnet runs
   for x402 `upto` interop. A prebuilt localnet artifact is required because the
   mainnet deployment uses BPF instructions surfnet cannot execute, and streaming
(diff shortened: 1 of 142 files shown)


EfeDurmaz16 (changes requested):
This PR leaves the public APIs and the pinned paths alone, so it does not fix or cause the real break. That lives in payment-channels#82: the next `payment_channels_ref` bump and the next IDL codegen still expect `program/payment_channels` / `payment_channels.so` / `paymentChannels`.

I would not merge a prose rename to "tab" while the session draft, the vendored IDL, and the generated clients still say payment channel. Happy to revisit if there is a written spec/product decision.

PR #288, opened 2026-08-17
feat(mints): add devnet-only USDtest

lgalabru:
## Summary
- stack USDtest support on #283
- register the devnet USDtest Token-2022 mint in Rust
- resolve it in MPP and x402 while rejecting omitted, mainnet, testnet, and localnet networks
- select Token-2022 and recognize the mint across charge, session, and settlement paths

## Test Plan
- `cargo test --workspace --quiet` (891 passed, 5 ignored)
- `cargo clippy -- -D warnings -A clippy::too_many_arguments`

### DIFF
--- a/rust/crates/kit/src/core/mints.rs
+++ b/rust/crates/kit/src/core/mints.rs
@@ -8,6 +8,9 @@
 pub const USDC_MAINNET: &str = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v";
 pub const USDC_DEVNET: &str = "4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU";
 pub const USDC_TESTNET: &str = USDC_DEVNET;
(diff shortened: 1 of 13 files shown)


EfeDurmaz16 (approved):
Devnet gate is applied consistently across every Rust resolution path, constructors fail closed on USDtest off devnet, and Token-2022 selection is right for both the symbol and the raw mint. CI is green.

- rust/crates/kit/src/mpp/server/html.rs:93 - the "USDTEST" display arm never matches the canonical mixed-case USDtest spelling, so the payment page renders "0.10 USDtes" instead of "$0.10"; match on the uppercased currency or add the mixed-case literal.
- USDtest lives only in the Rust registry, so a symbol-denominated challenge from a Rust server is unpayable by TS/Python clients (fails closed). Worth a line in the PR body, or register the mint plus the devnet gate there too.

Verified: full diff of 3076229b against the 055b0c8d base line by line; traced every resolution path (mpp charge and session, client and server, x402 exact, upto, and batch-settlement constructors); adversarial review with skeptic verification per finding; CI green including the harness matrix.

PR #291, opened 2026-08-23
pay-kit: require decimals for SPL candidates and verify decimals

latent-9:
x402 exact and mpp select used unwrap_or(6) to default decimals to 6 when the server omitted it. For non-6-decimal mints this silently produces wrong human_amount ranking (1000x skew for 9-decimal tokens) and wrong TransferChecked decimals.

Require decimals for SPL candidates (return None to skip) and verify that the on-chain TransferChecked decimals byte matches the requirements. Update test helpers to expect decimals instead of defaulting.

Fixes the remaining unwrap_or(6) instances after #285 (Audit #42) which fixed mpp/client/charge.rs.

Verification: POC verified 1000% — wrong decimals 9 vs 6 now fails verify (was Ok), select ranking skew 1000x confirmed, cargo check passes.

### DIFF
--- a/rust/crates/kit/src/select.rs
+++ b/rust/crates/kit/src/select.rs
@@ -572,7 +572,7 @@ mod tests {
             "amount": amount.to_string(),
             "currency": currency,
             "recipient": RECIPIENT,
-            "methodDetails": { "network": "mainnet" },
+            "methodDetails": { "network": "mainnet", "decimals": 6 },
         });
         let challenge = PaymentChallenge::new(
             "id-1",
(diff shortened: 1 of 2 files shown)


EfeDurmaz16 (dismissed):
Good, and it covers the two call sites #285 does not touch. The new `data[9]` read is safe: the guard above pins `data.len() == 10` and the TransferChecked discriminator first.

Non-blocking: `select.rs` uses `decimals?`, so a challenge missing decimals is silently dropped from the candidate list while the other sites return an error. A loud rejection is easier to debug than a payable option quietly disappearing.

EfeDurmaz16:
Could you consolidate this series (#291, #293, #294, #295, #296) into a single PR with signed/verified commits? Same bug class, and the cross-SDK claim is easiest to review and keep on record in one place - happy to prioritize review once it is up.

latent-9:

PR #293, opened 2026-08-24
fix(x402): require decimals for SPL payments in the Go client

latent-9:
## What

The Go x402 exact client defaults `extra.decimals` to 6 when a challenge omits the field, then feeds that default straight into `transferChecked`.

## Why

`parseEntry` fills `Extra.Decimals = 6` with `DecimalsSet = false` when the offer carries no decimals (`x402.go`), and `buildSPLTransfer` consumed it unconditionally. For any non-6-decimal mint (wrapped SOL at 9, for example) the built instruction carries the wrong decimals byte and fails on chain after the fee payer has already signed and paid.

The MPP charge client in this repo already refuses missing decimals for exactly this reason (audit #42, spec §7.2), and the Rust x402 client was fixed the same way in #285.

## Change

`buildSPLTransfer` now returns an error when `DecimalsSet` is unset, mirroring the charge client's wording. The token-program parity test now supplies decimals so it keeps exercising what it verifies, and a new regression test rejects a wrapped-SOL challenge that omits them.

### DIFF
--- a/go/protocols/x402/client/client.go
+++ b/go/protocols/x402/client/client.go
@@ -562,6 +562,13 @@ func buildSPLTransfer(
 	amount uint64,
 	entry *x402.AcceptsEntry,
 ) (solana.Instruction, error) {
+	// #42: decimals are required for SPL payments (spec §7.2 marks them MUST be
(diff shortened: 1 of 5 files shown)


latent-9:

PR #296, opened 2026-08-25
fix(x402): require decimals for SPL payments in the Python client

latent-9:
## What

The Python x402 exact client defaults `extra.decimals` to 6 when an offer omits the field, then feeds that default straight into the SPL `transferChecked` instruction.

## Why

For any non-6-decimal mint (wrapped SOL at 9, for example) the built instruction carries the wrong decimals byte and fails on chain after the fee payer has already signed and paid. The MPP charge client in this repo already refuses missing decimals for exactly this reason (spec §7.2), and the Rust x402 client was fixed the same way in #285.

## Change

The SPL build path now raises `ValueError` when the offer carries no usable `decimals`, instead of falling back to 6. The now-unused `_DEFAULT_DECIMALS` constant is removed.

A regression test rejects a wrapped-SOL challenge without decimals.


latent-9:

PR #297, opened 2026-08-26
fix(x402): require decimals for SPL payments across SDK clients

latent-9:
## What

Consolidates #291, #293, #294, #295 and #296 into one cross-SDK change: every language client (Rust, Go, Swift, Kotlin, Python) now refuses to build an SPL `transferChecked` when the x402 offer carries no `extra.decimals`, instead of silently defaulting to 6.

## Why

For any non-6-decimal mint (wrapped SOL at 9, for example) the defaulted instruction carries the wrong decimals byte and fails on chain after the fee payer has already signed and paid. The MPP charge client in every language already refuses missing decimals for exactly this reason (audit #42, spec §7.2).

## Per language

- **Rust** — `select` requires decimals for SPL candidates and verifies them against the funded mint; the charge fixture in its tests supplies the required field.
- **Go** — `buildSPLTransfer` returns an error when `DecimalsSet` is unset; token-program parity test updated, wrapped-SOL regression test added.
- **Swift** — `\_buildPaymentPayload` throws `PayKitError.invalidTransaction`; unused default constant removed; the two shared build vectors whose USDC offers omitted decimals now carry their correct value explicitly (only the Swift suite consumes that file); regression test added.
- **Kotlin** — `buildPayment` throws `IllegalArgumentException`; unused constant removed; regression test added.
- **Python** — raises `ValueError`; unused constant removed; regression test added.

## Verification

Per-language suites were run locally on this consolidated branch: Rust select tests 11/11, Go x402 packages ok, Python x402 client tests pass (including the two decimals cases), Swift full suite 203/203. All commits are SSH-signed.


EfeDurmaz16 (changes requested):
The direction is right: the blind default to 6 should go, and thanks for consolidating the series. But hard-requiring `extra.decimals` introduces an interop break. The canonical exact SVM spec (x402-foundation/x402, scheme_exact_svm.md) does not define `extra.decimals` at all - its `extra` block is feePayer, memo, recentBlockhash and lastValidBlockHeight - so the field is a pay-kit extension, and a spec-compliant non-pay-kit server that legally omits it would now have its offers rejected by every client in this PR.

The spec already models the right pattern with `recentBlockhash`: use the server-supplied value when present, otherwise the client fetches it itself. Decimals should work the same way: when `extra.decimals` is absent, fetch the mint account (the offer's `asset`) over the RPC connection the client already holds and read the authoritative decimals from chain, instead of erroring. When the field is present, keep using it as an RPC-saving hint. The on-chain TransferChecked still verifies whatever we pass against the mint, so wrong or lying values stay fail-closed either way; note also that `amount` is already in base units, so decimals never touches the amount math - its only consumer is the checked byte.

Happy to re-review quickly once the reject paths become fetch-from-mint fallbacks; the per-language error messages you added stay useful for the case where the RPC fetch itself fails.

latent-9:
Addressed all CHANGES_REQUESTED feedback. Implemented fetch-from-mint fallback for extra.decimals across all five SDK languages, mirroring the existing recentBlockhash pattern.

**What changed (per language):**
- Python: _fetch_mint_decimals(rpc, mint) helper reads byte 44 of the Mint layout via RPC; errors surface the mint address so the caller can recover
- Go: solanatx.ResolveMintDecimals(ctx, rpc, mint) plus plumbing ctx/rpc through buildSPLTransfer; client updated to call it when decimals absent
- Swift: RpcClient.getMintDecimals(pubkeyBase58:) plus fallback in _buildPaymentPayload; tests use URLProtocol stubs with drainBody() for httpBodyStream
- Kotlin: rpcDecimalsProvider: (() -> UByte)? = null param (same pattern as existing rpcBlockhashProvider); absent plus no provider equals clear error
- Rust: verify.rs skip-compare when requirements.decimals is None (on-chain TransferChecked enforces anyway); select.rs Candidate.decimals: Option<u8> with unwrap_or(6) for ranking heuristic only

**Tests:**
- Python 49/49, Go all pass, Swift 10/10, Rust 382 pass
- Kotlin: locally validated at code level; Gradle not available on dev box, CI runs gradle test

**Interop preserved:** when extra.decimals is present, it is still used as an RPC-saving hint (no behavior change for compliant callers). The on-chain TransferChecked instruction verifies the byte against the mint, so lying values stay fail-closed. Happy to re-review quickly.

latent-9:
Hi Efe, the review feedback is addressed. When `extra.decimals` is absent, every client now fetches the authoritative decimals from the mint account over the existing RPC connection (same shape as the `recentBlockhash` pattern), and the per-language error messages remain for when that RPC fetch itself fails. When the field is present it is kept as an RPC-saving hint, and the on-chain TransferChecked still fails closed on wrong values. Applied across Go, Kotlin, Python, Rust and Swift. Also dropped a stray comment-draft file that had ended up in the tree. Happy for a re-review whenever you get a chance.

EfeDurmaz16:
Thanks for the thorough pass across the five SDKs. Two things before this can move.

The Rust side does not compile: `select.rs:186` widens the field to `Option<u8>`, but line 536 still assigns `decimals: 6`.

The bigger one is semantics. This PR and your #285 now ship opposite rules for the same protocol field: #285 rejects a missing `decimals`, this one resolves it from the mint. main already rejects it in `rust/crates/kit/src/mpp/client/charge.rs:771` with "decimals is required for SPL charges (spec 7.2)" from audit #42. The two PRs share no files, so git would merge both and land the contradiction silently.

Can we settle on one rule and align both PRs on it? My read is that reject matches what the repo already decided, but the mint fallback is a real ergonomic argument, so tell me your reasoning before I push either way.

latent-9:
@EfeDurmaz16 the fetch-from-mint fallback is now complete across all five exact clients — could you take another look when you have a moment?

What changed since your review:
- **Rust** (was still `unwrap_or(6)`): `build_spl_instructions` now takes the existing `RpcClient` and resolves absent decimals from mint byte 44; malformed hints are rejected at parse time instead of silently dropped. Also fixed a latent `select.rs` type error the `client` feature exposed.
- **Kotlin** (fallback existed only as an unwired optional lambda): added `X402RpcClient.fetchMintDecimals` + `mintDecimalsProvider`, wired through `PayKitClient.Builder.x402(rpc)` and the interceptor, so the production path fetches by itself. Also annotated a previously dead no-provider test with `@Test`.
- **Go/Swift**: out-of-range present hints are now rejected before conversion (no silent truncation/wrap); absent still fetches. **Python**: dropped the redundant `& 0xFF` mask (range-checked above).

Verified locally: Go suite green, Rust `x402::client::exact` + select suites green (195 tests with `client,server` features), Swift files parse-clean (full `swift build` is broken on my toolchain by a pre-existing strict-concurrency error in `SolanaSigner.swift`, untouched here — CI will confirm). Kotlin/Python rely on CI (no local JDK 17 / Python 3.11 here).

Greptile's P1 on the Python mask should now read clean. 🙏

EfeDurmaz16:
Sure, I'm on it.

EfeDurmaz16:
Thanks, @latent-9! I have carried your work into #324 with co-author credit and addressed the remaining review findings. Closing this in favor of that PR.

PR #299, opened 2026-09-01
docs(skill): modernize Pay Kit protocol references

JasonColapietro:
## Summary
- replace removed mpp-sdk paths with the current Pay Kit Rust, TypeScript, and harness locations
- route x402 exact, upto, and batch-settlement work to the public in-repo implementations instead of a private checkout
- update MPP subscription from unpublished/future guidance to the shipped Rust and TypeScript sources
- make protocol scope target-language-specific and require current harness evidence before marking a matrix cell complete

## Why
The official payment-sdk-implementation skill described x402 as future, said subscriptions had no implementation, and directed contributors to removed paths and a private local checkout. Following it would block valid ports or cause work against the wrong source tree.

## Verification
- strict YAML frontmatter parse; description is 327 characters
- every newly cited in-repo source, test, IDL, workflow, and harness path resolves
- stale-path/status scan found no mpp-sdk root, rust/crates/mpp path, private checkout, or scaffold-only instruction
- git diff --check

The generic quick_validate.py helper was also attempted, but its local Python environment does not include PyYAML; the equivalent metadata checks above passed.


lgalabru (approved):
LGTM, thanks @JasonColapietro!

PR #300, opened 2026-09-01
refactor(signing)!: migrate to solana-keychain 2.x TransactionSigner

ledgicr:
Reworked per @EfeDurmaz16's review. This is PR A from that suggestion: the migration alone, nothing Ledger-specific.

Current head is two commits:

| Commit | What |
|---|---|
| `db71da27` | keychain 2.x `TransactionSigner` migration |
| `e8c81134` | `@toon-format/toon` 2.3.1 audit override, requested by @EfeDurmaz16 |

## What changed since 581cbaa8

**The four review comments, addressed:**

- **`Cargo.toml:85` — git dep.** Gone. Pinned to `solana-keychain = "2.0.0-beta.1"` from crates.io. `cargo publish` resolves.
- **`Cargo.toml:31` — the `ledger` feature.** Removed from this PR entirely. It follows once solana-foundation/solana-keychain#301 ships, as you suggested.
- **`gate.rs:91` — the unwidened configs.** `MppConfig` (`mpp/server/charge.rs`), the subscription `Config` (`mpp/server/subscription.rs`), `confidential_worker`, the three harness binaries and both integration-test suites are all migrated. `--features axum` and the `payment_link_server` example build.
- **`core/signing.rs:66` — the missed call site.** `mpp/client/confidential.rs` now routes through `sign_legacy_transaction`, so `--features confidential` and `--features worker` compile.

**And the fifth, which was the most useful comment in the review:**

- **`upto.rs:96` — the voucher-only signer.** You were right, and it was a real inconsistency: `receiver_authorizer_signer` only ever reaches `sign_message`, so widening it demanded a capability it never uses. It is back to `SolanaSigner`, with a comment saying why. The fee-payer fallback upcasts into it. This also keeps the rule legible for the path a hardware wallet cannot serve yet.

Also rebased onto the batch-settlement SVM rewrite (#301). The three files that touches were taken from it verbatim and the trait change reapplied on top, rather than merged line by line.

The kit is already at `0.6.0` on `main`, so the breaking-API version bump you asked for came along with the rebase.

## Greptile's `recipient_signer` finding, fixed

Greptile was right, and it was the same mistake as `upto.rs:96`. `recipient_signer` reaches exactly one production site, `mpp/server/confidential.rs:156`, which forwards it to `derive_confidential_keys(&dyn SolanaSigner)` to derive the recipient ElGamal key from a signature over the token-account address. It never signs a transaction, so requiring `TransactionSigner` rejected valid message-only signers for no reason.

It is back to `Option<Arc<dyn solana_keychain::SolanaSigner>>`, and there is now a compile-time guard against re-widening it. `recipient_signer_accepts_a_message_only_signer` builds a `Config` with a `MessageOnlySigner` that implements `SolanaSigner` and nothing else, so restoring `TransactionSigner` on that field fails to compile rather than quietly narrowing which signers the confidential path accepts. Confirmed by doing it: widening the field breaks the build at that test.

Which signers widened was decided by the compiler rather than by search-and-replace. Every signer site was reverted to `SolanaSigner`, then widened only where `cargo check --workspace --all-targets` demanded it, iterated to a fixed point. That is what surfaced this: the compiler never asked for `recipient_signer`, `receiver_authorizer_signer`, or the client session signer.

## Four MPP paths that raw-signed transaction bytes

A signer bound is only half the story. `sign_message(&tx.message.serialize())` type-checks against `SolanaSigner` whatever those bytes mean, so the compiler audit above could not see this class of problem at all. Four MPP paths built a transaction, signed its serialized message through `sign_message`, and spliced the 64 bytes into the signature vector by hand:

- `mpp/server/charge.rs`, fee-payer co-signing on the pull path
- `mpp/server/confidential.rs`, gateway co-signing of each bundle transaction
- `mpp/server/confidential.rs` `broadcast_close`, the orphan sweeper's close transaction
- `mpp/server/subscription.rs` `co_sign_as_fee_payer`, activation co-signing

All four now go through the shared `core::signing` helpers, and the manual signature splicing is gone. `charge::Config::fee_payer_signer`, the subscription config's `fee_payer_signer` and the confidential worker's gateway signer widen to `TransactionSigner` accordingly, and `gate.rs` loses the upcast it needed only while MPP's field was narrow.

Leaving these would have made this PR's own claim false. A hardware wallet's `sign_message` wraps its payload in an off-chain-message envelope and signs that, so a spliced signature covers the wrong bytes and the cluster rejects the transaction at preflight.

New `core::signing::cosign_legacy_fee_payer` for the two legacy-`Transaction` paths, the counterpart to the existing `cosign_versioned_fee_payer`.

**This tightened a real weakness.** All four hand-rolled versions searched for the configured sponsor at *any* index and signed there. Both helpers pin the fee-payer signer to account key zero. Under the old form, a crafted transaction could place the sponsor at a non-zero required-signer index, collect a signature there, and leave its actual fee payer unsigned. The slot checks stay at the call sites so those failures keep reporting as `invalid_payload` rather than as a server fault.

Message-only roles stay narrowly typed as `SolanaSigner`: `recipient_signer` (charge and worker), `receiver_authorizer_signer`, the client `ActiveSession` signer, `sign_voucher`, `mpp/client/authenticate`, `x402/siwx` and `derive_confidential_keys`.

The only raw transaction signing left in the crate is in `core/settlement/testkit.rs`, and it stays: that helper takes a concrete `Arc<MemorySigner>`, not a trait object, so no hardware signer can reach it.

**Regression guards, in both directions.** `TransactionOnlySigner` implements `TransactionSigner` and returns an error from `sign_message`, standing in for a hardware wallet. `fee_payer_cosign_never_calls_sign_message` (charge), `activation_cosign_never_calls_sign_message` (subscription) and `legacy_fee_payer_cosign_never_calls_sign_message` (core::signing) each fail if their path reverts to raw signing, which was confirmed by reverting both production paths and watching them fail. Each has a companion test that a sponsor which is not account key zero is rejected with the transaction left unsigned. `charge.rs`'s co-sign moved into `cosign_fee_payer_slot` so it is reachable without a live RPC. The two confidential paths are covered at the helper level rather than at their own call sites, which need an RPC and a real proof bundle.

## The audit override (`e8c81134`)

Done as asked: `'@toon-format/toon': '2.3.1'` under `overrides` in `typescript/pnpm-workspace.yaml`, lockfile regenerated with `pnpm install --lockfile-only`, both files in one commit.

- `pnpm install --frozen-lockfile` passes, using the pinned pnpm 11.13.0 from `typescript`.
- `pnpm audit --production` reports `No known vulnerabilities found`. It was reporting 1 high before.
- The lockfile delta is limited to `@toon-format/toon` 2.1.0 → 2.3.1 and its two `incur` edges: 11 lines, counting the override entry and the new integrity hash. No collateral drift.

Pinned exactly at 2.3.1 rather than floored at `>=2.3.1` like the advisory overrides above it. `@toon-format/toon` has since published a 4.x line, and a pnpm override resolves against the registry rather than against the dependent's declared range, so a floor would select 4.1.1 and carry `incur` across two majors to close a patch-level advisory. 2.3.1 is the lowest patched version on the line `incur` already expects. The advisory reaches us at `packages/pay-kit > mppx > incur > @toon-format/toon`, four levels down, so an override is the only lever.

I have deliberately not touched the `packageManager` pin. npm reports `pnpm@11.13.0` as deprecated ("This is a broken version. Please install pnpm v11.13.1 or newer"), but that is a separate issue: 11.13.0 is what `pnpm/action-setup` resolves for CI and therefore what validates this lockfile, and bumping it is not this PR's call.

## Verification

Local, at head `e8c81134`:

- `just rs-test`: green, **973 tests** across all 12 test binaries
- `cargo check --workspace --all-features`: clean
- `cargo fmt --check`: clean
- `just rs-lint`: reports the same three pre-existing errors as `main` (2x `doc list item without indentation`, 1x `too many arguments`) and no new ones
- `--features` combinations checked individually: `confidential,client`, `confidential,server`, `worker`, `testkit`, and `confidential,worker,server,client,axum,mpp,x402`

**One baseline exception worth stating plainly.** `cargo check --workspace --all-features --all-targets` is *not* green, but it is not green on `main` either. It reaches exactly two errors, both `unresolved import litesvm` at `mpp/protocol/confidential.rs:209` and `:301`, and both are pre-existing: `--all-features` turns on `litesvm-tests`, and your own `Cargo.toml` note says litesvm is deliberately not a dev-dependency because it would break `cargo publish`. I checked `main` at `1aa9333b` with the same command and got the identical two errors. Nothing in this PR adds to that set.

The same goes for the three `just rs-lint` errors. `core/voucher.rs` has a zero-byte diff from the merge base, and `x402/server/upto.rs`'s `validate_open_transaction` has an identical 10-argument signature there; only its line number moved.

**On CI, to be accurate rather than encouraging:** repository Actions require maintainer approval, so the only check that has run against `e8c81134` is Greptile, which passes. The last full Actions run was on an earlier head, where every job was green except `Audit`, which is what `e8c81134` fixes. The Rust suite passed there in 7m26s. The numbers above are local runs, not CI results.

Unrelated to this PR: the `Report` workflow has been failing on `main` since `solana-foundation/surfpool` renamed the branch that `report.yml:80` checks out. Fixed separately so it stays out of this diff.

## Why the migration is needed at all

keychain 2.0 split the signer traits by capability. `SolanaSigner` keeps `pubkey`/`sign_message`/`is_available`; transaction signing moved to `TransactionSigner::sign_transaction` and now takes a `VersionedTransaction`. To see the failure on `main`, bump `solana-keychain` to `2.0.0-beta.1` and run `cargo check`:

```
error[E0599]: no method named `sign_transaction` found for reference `&dyn SolanaSigner`
   --> crates/kit/src/core/payment_channels.rs
error[E0599]: no method named `sign_transaction` found for struct `Arc<dyn SolanaSigner>`
   --> crates/kit/src/core/settlement/worker.rs
```

Satisfyingly, this *removes* a workaround rather than adding one. `core::signing` carried a note saying hardware-wallet support had to stay out until keychain published a transaction-signing API. It has, so those helpers now call `sign_transaction` instead of raw-signing `tx.message.serialize()` through `sign_message`. Every validation they performed is kept and still runs first, since the signer-index and slot-count checks are a security property rather than incidental.

Behaviour is unchanged for software backends: `MemorySigner::sign_transaction` signs exactly the bytes that went through `sign_message` before, so signatures are byte-identical. That is why the existing suite passes with **no assertion changed**, including `sponsored_payment_signs_only_the_payer_slot` and `cosign_accepts_fee_payer`, the two that would catch a slot-placement regression.

## Breaking change

Public functions and config fields that sign transactions now take `dyn TransactionSigner`.

- Concrete keychain signers (`MemorySigner` and friends): no change, they implement `TransactionSigner`.
- keychain's umbrella `Signer` enum: **does** need a change. It implements only `SolanaSigner`; reach the capability via `Signer::as_transaction_signer()`, which returns `Option<&dyn TransactionSigner>` and is `None` for backends that broadcast rather than return signed bytes.
- Your own `dyn SolanaSigner` trait objects: widen them.
- Newly in this set, from the four paths above: `mpp::server::charge::Config::fee_payer_signer`, the subscription config's `fee_payer_signer`, and the confidential worker's gateway signer.


rust/crates/kit/Cargo.toml
```diff
@@ -70,9 +75,14 @@ otel = [
...
-# bound to <3.5, letting signature resolve to 3.3.x.
-solana-keychain = { version = "1.4", default-features = false, features = ["memory", "sdk-v3"] }
+# Signing — solana-keychain 2.x with sdk-v3. 2.0 split transaction signing out
+# of `SolanaSigner` into the `TransactionSigner` capability trait and moved it to
+# `VersionedTransaction`; see core::signing for how the kit adapts.
+#
+# NOTE: git dep because keychain 2.x is not published yet. The branch is the one
+# under review as solana-foundation/solana-keychain#301; repin to the published
+# 2.x version once that lands, before merging this.
+solana-keychain = { git = "https://github.com/ledgicr/solana-keychain", branch = "ledger-signer-v2", default-features = false, features = ["memory", "sdk-v3"] }
```
EfeDurmaz16: Mutable branch on a personal fork with no rev pin cannot land in a payments SDK, even temporarily. solana-keychain 2.0.0-beta.1 has been on crates.io since 2026-08-27 and commit 1 builds and passes the lib tests against it, so pin `version = "2.0.0-beta.1"` and drop the git dep.

rust/crates/kit/Cargo.toml
```diff
@@ -24,6 +24,11 @@ client = ["dep:reqwest"]
 # plus the server side.
 axum = ["mpp", "x402", "server", "dep:axum"]
 gcp_kms = ["solana-keychain/gcp_kms"]
+# Ledger hardware-wallet signing over USB-HID. A pure feature passthrough: the
+# MPP and x402 builders take `&dyn TransactionSigner`, which a Ledger signer
+# satisfies, so there is no code here. Not folded into any aggregate feature --
+# it pulls `hidapi`, which builds a native HID library.
+ledger = ["solana-keychain/ledger"]
```
EfeDurmaz16: This needs solana-keychain#301, which is unmerged. Please move it to a follow-up PR if / once that ships. It is a one-line passthrough and should not hold the migration.

rust/crates/kit/src/gate.rs
```diff
@@ -88,7 +88,7 @@ pub struct PayKitConfig {
     pub challenge_binding_secret: Option<String>,
     /// When set, the server sponsors the network fee: drives MPP fee-sponsored
     /// mode and is used as the x402 fee-payer address.
-    pub fee_payer_signer: Option<Arc<dyn SolanaSigner>>,
+    pub fee_payer_signer: Option<Arc<dyn TransactionSigner>>,
```
EfeDurmaz16, line 98: Widened here, but MppConfig (mpp/server/charge.rs:218) and the subscription Config (mpp/server/subscription.rs:100) still hold `Arc<dyn SolanaSigner>`, so `--features axum` fails at gate.rs:175 and the CI job building the payment_link_server example goes red. The harness bins (x402_harness_upto_server.rs:77) and integration-tests (charge_integration.rs:42) hit the same mismatch.

rust/crates/kit/src/core/signing.rs
```diff
@@ -36,11 +40,39 @@ pub async fn sign_versioned_transaction_slot(
...
+/// `TransactionSigner::sign_transaction` takes a `VersionedTransaction`, while
+/// much of the kit still builds legacy `Transaction`s and serializes them as
+/// such. Converting to a versioned view is lossless in both directions here: a
+/// legacy transaction becomes `VersionedMessage::Legacy`, which leaves the
+/// account keys, the header and therefore the signature slot ordering exactly
+/// as they were, so copying the signature vector back is a faithful round-trip.
+///
+/// This keeps the conversion in one place rather than at each call site, and
+/// leaves callers free to go on serializing the legacy transaction.
+pub async fn sign_legacy_transaction(
```
EfeDurmaz16, line 66: One call site missed this helper: mpp/client/confidential.rs:871 still calls `sign_transaction` on `&dyn SolanaSigner`, so `--features confidential` (and `worker`) does not compile. Widen that param and route through here.

rust/crates/kit/src/x402/server/upto.rs
```diff
@@ -90,10 +90,10 @@ pub struct UptoConfig {
     /// `settle_and_seal` (lifecycle authority: it can always seal with
     /// `has_voucher = 0` to recover rent, but cannot settle a nonzero amount
     /// or redirect funds).
-    pub fee_payer_signer: Arc<dyn SolanaSigner>,
+    pub fee_payer_signer: Arc<dyn TransactionSigner>,
     /// Signer for the Ed25519 settlement voucher (the channel's
     /// `authorized_signer`). Defaults to `fee_payer_signer` when absent.
-    pub receiver_authorizer_signer: Option<Arc<dyn SolanaSigner>>,
+    pub receiver_authorizer_signer: Option<Arc<dyn TransactionSigner>>,
```
EfeDurmaz16: This signer only ever reaches `sign_message` for the voucher (upto.rs:729), so by the PR's own rule it should stay `SolanaSigner`. The fee-payer fallback upcasts fine on current stable. Worth keeping the rule consistent since this is exactly the path a Ledger cannot serve yet.

EfeDurmaz16 (changes requested):
The signing crux in core/signing.rs is sound: MemorySigner 2.x signs the same message bytes as 1.4, so software-key signatures are byte-identical, and the legacy round-trip test pins the property that matters. The PR is not mergeable as is because only the default feature set compiles and the dependency points at a fork branch. I suggest a split: PR A is commit 1 alone, pinned to the published 2.0.0-beta.1 and green under `cargo check --workspace --all-features --all-targets` plus harness-bins and integration-tests, with the kit bumped to 0.6.0 for the breaking API; PR B adds the `ledger` feature if / once solana-keychain#301 ships. Inline comments have the details.

Verified at 581cbaa8: repinned to crates.io 2.0.0-beta.1, `cargo test -p solana-pay-kit --lib` passes 382 tests; the workspace all-features check, harness-bins and integration-tests fail with errors that do not exist on main.

ledgicr:
@EfeDurmaz16 reworked per your review, and pushed. This is PR A: the migration only, no Ledger code.

All five comments addressed. The `upto.rs:96` one was the most useful: you were right that widening the voucher-only signer contradicted the PR's own rule, and it is back to `SolanaSigner` with a comment saying why.

Also rebased onto the batch-settlement rewrite (#301), taking your three rewritten files verbatim and reapplying the trait change rather than merging line by line.

One thing to flag rather than let you find it: `cargo check --workspace --all-features --all-targets` is not green, but it is not green on `main` either. It reaches exactly two `unresolved import litesvm` errors at `mpp/protocol/confidential.rs:209` and `:301`, because `--all-features` enables `litesvm-tests` and litesvm is deliberately not a dev-dependency. Same two errors on `main` at `1aa9333b`. Nothing here adds to that set. `just rs-test` is green at 966 tests, and `cargo check --workspace --all-features` (without `--all-targets`) is clean.

The `ledger` passthrough is on a separate branch and will follow once solana-keychain#301 ships, as you suggested.

EfeDurmaz16:
Audit is failing on [GHSA-p95v-992w-h6c3](https://github.com/advisories/GHSA-p95v-992w-h6c3), which is fixed in `@toon-format/toon@2.3.1`. Please add `'@toon-format/toon': '2.3.1'` under `overrides` in `typescript/pnpm-workspace.yaml`, run `pnpm install --lockfile-only` from `typescript`, and commit both the workspace file and updated lockfile.

ledgicr:
Done in `e8c81134`, workspace file and lockfile in one commit.

```
$ pnpm install --frozen-lockfile
Scope: all 3 workspace projects
Done in 287ms using pnpm v11.13.0

$ pnpm audit --production
No known vulnerabilities found
```

Both run from `typescript` with the pinned pnpm 11.13.0. Lockfile delta is 11 lines and touches nothing but this package: the override entry, the 2.3.1 resolution and integrity hash, and the two `incur` edges.

One deviation worth flagging. I pinned exactly `'2.3.1'` as you wrote it rather than floor it at `>=2.3.1` like the advisory overrides above it, because `@toon-format/toon` has since published a 4.x line. An override resolves against the registry rather than against the dependent's declared range, so a floor would have selected 4.1.1 and carried `incur` across two majors to close a patch-level advisory. Happy to switch to a floor if you would rather keep the file consistent.

I did not touch the `packageManager` pin, though npm now reports `pnpm@11.13.0` as a broken release and suggests 11.13.1 or newer. That seemed like its own change rather than something to slip into this PR.

rust/crates/kit/Cargo.toml
ledgicr: Resolved: `crates/kit/Cargo.toml:76` now pins `solana-keychain = "2.0.0-beta.1"` from crates.io. No git dependency remains in the manifest, so `cargo publish` resolves.

rust/crates/kit/Cargo.toml
ledgicr: Resolved: the `ledger` feature is out of this PR entirely. It moves to the follow-up that lands after solana-foundation/solana-keychain#301.

rust/crates/kit/src/gate.rs
ledgicr: Both are widened now: `mpp/server/charge.rs:218` and `mpp/server/subscription.rs:100` are `Option<Arc<dyn solana_keychain::TransactionSigner>>`. The worker plumbing, the three harness binaries and both integration suites use `TransactionSigner` wherever they sign transactions, so `gate.rs` no longer needs the upcast that was on this line.

rust/crates/kit/src/core/signing.rs
ledgicr: That call site routes through the helper now: `mpp/client/confidential.rs:870` calls `crate::core::signing::sign_legacy_transaction`. The `confidential` and `worker` feature builds pass locally.

rust/crates/kit/src/x402/server/upto.rs
ledgicr: Kept narrow: `receiver_authorizer_signer` is still `Option<Arc<dyn SolanaSigner>>` at `upto.rs:101`, since it only ever signs the voucher message. The fee-payer fallback upcasts into it.

rust/crates/kit/src/mpp/server/charge.rs
```diff
@@ -228,7 +228,7 @@ pub struct Config {
     ///     the proofs), and the recipient reconciles the amount out of band.
     ///
     /// Not used for non-confidential flows.
-    pub recipient_signer: Option<Arc<dyn solana_keychain::SolanaSigner>>,
+    pub recipient_signer: Option<Arc<dyn solana_keychain::TransactionSigner>>,
```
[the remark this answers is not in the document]
ledgicr: Restored to message-only: `recipient_signer` is `Option<Arc<dyn solana_keychain::SolanaSigner>>` at `mpp/server/charge.rs:231`. The `recipient_signer_accepts_a_message_only_signer` test at `charge.rs:4867` builds a config with a signer that implements only `SolanaSigner`, so widening the field again fails to compile.

ledgicr:
Rebase readiness, checked against current `main` so the merge is not a surprise
to either of us. Nothing pushed; the branch has not moved.

I rebased onto `202b0af6` on a scratch branch. **It applies clean.** No
conflicts, and `cargo test` from `rust/` gives **973 passed, 0 failed** — the
same count as this PR's head, so the rebase costs nothing.

Worth stating because I expected the opposite. The delta since this PR's merge
base `1aa9333b` is two commits, #304 and #305, and one rewrites
`x402/server/batch_settlement.rs`, which this PR also touches. They do not
collide: #305's `spawn_blocking` work is at `batch_settlement.rs:1648` and
`:1838`, on the deposit lookup and broadcast, while this PR's change is at
`:2216` in the settlement-signing loop. Different regions of the same file, so
git had nothing to resolve.

I confirmed both sides survive rather than trusting the clean apply:

- ours: `use solana_keychain::TransactionSigner` (line 37),
  `fee_payer_signer: Arc<dyn TransactionSigner>` (188, 208),
  `sign_legacy_transaction(...)` (2216), `impl TransactionSigner for TestSigner`
  (2679);
- theirs: the `spawn_blocking` helpers at 1648 and 1838, and the new
  `core/store.rs` (136 lines, no overlap with this PR at all).

So the delta is mechanical: rebase, run the suite, done.

## One thing to decide before either merges

**#308 adds the identical `@toon-format/toon: '2.3.1'` override**, in the same
`overrides` block, appended after `body-parser` exactly as `e8c81134` does here:

```diff
@@ -28,3 +28,5 @@ overrides:
   body-parser: '>=2.3.0'
+  # GHSA-p95v-992w-h6c3: prototype pollution while decoding untrusted TOON.
+  '@toon-format/toon': '2.3.1'
```

Whichever lands second conflicts on `typescript/pnpm-workspace.yaml` and
`typescript/pnpm-lock.yaml`, and the resolution is to **drop the duplicate**, not
merge two hunks that say the same thing. If #308 goes first I will drop
`e8c81134` from this branch and re-check `pnpm audit --production` against #308's
lockfile. Happy either way — flagging it so nobody hand-merges it.

#308 overlaps this PR on ten files in total:

```
rust/crates/harness-bins/src/bin/x402_harness_upto_server.rs
rust/crates/kit/Cargo.toml
rust/crates/kit/src/core/payment_channels.rs
rust/crates/kit/src/gate.rs
rust/crates/kit/src/mpp/server/session.rs
rust/crates/kit/src/x402/client/batch_settlement/payment.rs
rust/crates/kit/src/x402/server/batch_settlement.rs
rust/crates/kit/src/x402/server/upto.rs
typescript/pnpm-lock.yaml
typescript/pnpm-workspace.yaml
```

Five of those carry this PR's `SolanaSigner` → `TransactionSigner` widening, so a
real conflict there is likely once #308 merges. When it happens, the bound on
each signer should be re-derived by the compiler rather than hand-merged: the
narrow ones are narrow for a reason, and `receiver_authorizer_signer` and
`recipient_signer` must stay `SolanaSigner`.

## CI has never run on this head

Worth saying plainly: repository Actions require maintainer approval, so the only
check that has run against `e8c81134` is Greptile, which passes. The last full
Actions run was on an earlier head, where every job was green except `Audit` —
which `e8c81134` is what fixes. **The 973 above is a local run, not a CI result.**

If you approve Actions on this head, that closes the last gap between what I can
demonstrate and what the repository can verify independently.

## One thing noticed in passing, not touched

At `batch_settlement.rs:2224` the settlement broadcast still calls
`send_and_confirm_transaction` synchronously inside an `async fn`. #305 moved the
deposit broadcast off the runtime for exactly that reason but did not cover this
path. It predates this PR and is upstream's code, so I have left it alone, but it
looks like the same stall #305 set out to remove.

EfeDurmaz16:
@ledgicr Thanks for checking. We’ll handle any conflicts based on the merge order and rerun the checks before merging, so no need to worry about that now :) We’ll also check the Actions approval status.

lgalabru:
@copilot resolve the merge conflicts in this pull request

lgalabru (dismissed):
Looking great thanks @ledgicr!

EfeDurmaz16 (approved):
Approved, thanks!

PR #301, opened 2026-09-01
feat(x402/batch-settlement)!: rewrite the SVM scheme to the current spec

lgalabru:
## Summary

Rewrites the Rust SVM `batch-settlement` scheme against the current specification ([`scheme_batch_settlement_svm.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/batch-settlement/scheme_batch_settlement_svm.md)), covering the client and the resource server.

The module that was here implemented an earlier draft. The spec changed the wire contract incompatibly, so this replaces the scheme rather than extending it. The wire format is kept identical to the in-progress TypeScript implementation on `solana-foundation/x402#feat/svm-batch-settlement` — same field names, same `invalid_batch_settlement_svm_*` codes, same treatment of the cooperative close.

## Breaking changes

| | Before (old draft) | Now (spec) |
|---|---|---|
| `extra` | `channelProgram`, `profiles`, `gracePeriodSeconds`, `suggestedDeposit`, `minimumDeposit`, `minVoucherDelta`, `distributionSplits` | `feePayer`, `withdrawDelay`, `tokenProgram`, `memo`, `receiverAuthorizer`, `paymentFlow`, corrective `channelState` / `voucherState` |
| `channelConfig` | `payee`, `mint`, `authorizedSigner`, `depositAmount`, `recentSlot` | `payer`, `payerAuthorizer`, `receiver`, `receiverAuthorizer`, `token`, `withdrawDelay`, `salt`, `openSlot` |
| voucher | `cumulativeAmount`, `signer`, arbitrary `expiresAt` | `maxClaimableAmount`, `expiresAt` pinned to `0`, signer taken from `channelConfig.payerAuthorizer` |
| `deposit` | flat `transaction` | nested `{amount, transaction}` |
| response | flat `chargedAmount` / `channelState` | nested under `extra`, plus `commitmentId` |

The payment-channels program id is a network constant and is no longer negotiable through `extra` — a resource server must not be able to point a client at a program of its choosing.

Public API: `BatchConfig` fields are renamed (`recipient` → `pay_to`, `grace_period_seconds` → `withdraw_delay`, `operator_signer` → `fee_payer_signer`); `settle_batch`/`distribute` become `claim`/`settle`; `verify_payment` no longer commits and is paired with `settle_payment`.

## What it does

**Sponsor acceptance policy** (`tx_policy.rs`, new). Every client-supplied `open` / `top_up` / `request_close` is validated statically before the fee payer co-signs it: no address lookup tables, a required-signer set of exactly `{feePayer, payer}` with the payer's signature verified first, a bounded ComputeBudget prefix, a required Memo (the declared `extra.memo` or a ≥16-byte hex nonce), exact 14/6/2-account tables, fully decoded arguments with no trailing bytes, a distribution pinned to 100 % `payTo`, and the channel PDA rederived from the instruction's own args. Simulation runs before broadcast but is not relied on — by then the signature has already authorized fee and, for `open`, rent.

**The `authorization` flow, end to end.** `verify_payment` is read-only and reserves the channel; the watermark commits — and any deposit transaction broadcasts — only in `settle_payment`, after the handler succeeds. A failed handler leaves the client uncharged and free to retry the same voucher.

**Metering rules.** Fixed pricing (a fresh voucher must advance by exactly one price), the deposit cap, idempotent replay of an exact repeat, and per-channel serialization held across the handler.

**Corrective 402.** Carries `channelState` plus a `voucherState` proof — the client's own signature at the claimed amount. The client refuses to adopt a `chargedCumulativeAmount` the server cannot prove, so a server cannot walk the cumulative base up and have the client sign away funds it never spent.

**Out-of-band redemption.** `claim` (Ed25519 + `settle`, ≤ 4 channels per legacy transaction), `settle` (`distribute`), `finalize_close` (`seal` + sealed `distribute` once the grace period elapses), `reclaim`, and `getProgramAccounts` discovery by `rent_payer` with PDA revalidation.

## Deliberately not implemented

The immediate cooperative close (`CloseAuthorization` + `settle_and_seal`) is refused with `invalid_batch_settlement_svm_close_authorization`, matching the TypeScript implementation. It is optional in the spec, and honoring it safely needs a facilitator-local trusted binding between a receiver-authorizer key and the server that owns `payTo`, established out of band no later than the first deposit. There is no such registry here, and a key that merely appears in a request is not a trust anchor. The interoperable payer-signed `request_close` path is fully supported.

## Reuse

`core::session::accept_voucher`, `core::voucher`, `core::store` (memory + Redis + CAS), `core::settlement::packing`, and every payment-channels instruction builder are used unchanged — the same wire-agnostic core the MPP `session` intent runs on.

`upto` already carried most of the sponsor policy. Its transaction-layout scanner is extracted to `core::payment_channels::scan_channel_tx_layout` and both schemes now share it, parameterized by a `MemoPolicy` (`upto` never declares a memo; `batch-settlement` requires one). `upto`'s policy suite passes unchanged, and the cross-protocol `x402 -> mpp::server::charge` import for the ComputeBudget decoder is gone.

## Behavior worth reviewing

- **`gate.rs` returns `409` on an idempotent replay.** The spec says return the cached response and do not re-run the handler; this gate owns no response cache (the TypeScript scheme delegates it to `getReplayResponse`), so it reports the conflict rather than double-charge or double-execute. Applications needing true replay should cache on `extra.commitmentId` ahead of the gate.
- **`corrective_challenge` is RPC-free** — no blockhash or slot hint. It answers a client that already has a channel and will retry with a plain `voucher`, which needs neither, and an error path should not fail because the RPC is degraded. A client that then needs a top-up may fetch its own, which the scheme allows.
- **A commit failure after the resource was served is logged, not surfaced.** The response stands; the uncharged request is the server's loss and is worth an alert.

## Testing

- 953 lib tests pass (`--features server,client,axum`), 64 of them new.
- Client and server are tested against each other: the transactions the client builds are asserted to pass the server's own sponsor policy, for `open`, `top_up` and `request_close`.
- Adversarial coverage on the policy: smuggled System transfer, second channel instruction, foreign program, missing or mismatched Memo, unsigned payer slot, wrong fee payer, diverted distribution, stale and future `openSlot`, over-ceiling compute budget.
- `cargo clippy --all-targets -D warnings` clean on every changed file; workspace `--all-targets` and the default-feature (crates.io) build both green.

Not covered by unit tests: the RPC-dependent paths (`settle_payment` broadcast, `claim`/`settle`/`finalize_close`/`reclaim`, discovery). Those want a surfpool integration test — happy to add one in this PR if you'd rather not land it without.


rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
-            let (Some(sig_b58), Some(expires_at)) = (
-                state.highest_voucher_signature.as_ref(),
-                state.highest_voucher_expires_at,
-            ) else {
-                continue; // no voucher accepted yet
-            };
-            if state.cumulative == 0 {
+            // `settle` requires a strictly increasing watermark, so a channel
+            // already claimed at this amount would only fail onchain.
+            if state.cumulative <= state.settled_on_chain {
```
EfeDurmaz16, line 1979: `settled_on_chain` is never updated after a claim, so an already-claimed channel is re-packed at the same cumulative, `settle` fails `voucherWatermarkNotMonotonic`, and the whole batch (and the rest of the call) fails. Read on-chain `settled` here like `settle()` does at L1223, and also skip `close_requested_at`/non-`Open` channels.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
+
+    /// Commit a verified payment. Call only after the resource handler has
+    /// succeeded — this is the step that charges the client.
+    pub async fn settle_payment(
+        &self,
+        outcome: BatchOutcome,
+    ) -> Result<BatchSettlementResponse, Error> {
+        let network = outcome.requirements.network.clone();
+        match &outcome.payload {
+            BatchPayload::Refund { transaction, .. } => {
```
EfeDurmaz16, line 1597: This co-signs `request_close` without claiming the stored voucher first. Once `Closing`, regular `settle` is gone and `finalize_close` seals at the on-chain watermark, so the delta is returned to the payer. We are the `payee`: `settle` before broadcasting, or use `settle_and_seal` in `finalize_close`.

Also, retry is not idempotent: `verify_refund` accepts `Closing` but this rebroadcasts and fails at simulate (502). Spec/TS return the observed `Closing` state without rebroadcast. `record_close` also errors if the store has no record even though the close landed.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
-                    .await
+                let max_claimable = check_voucher(voucher, &config, &channel_id)?;
+                let deposit_amount = deposit.amount()?;
+                // A top-up rides on an existing channel; a first deposit opens
+                // one. The stored record decides which, because the onchain
+                // channel does not exist yet for an `open`.
+                let (form, prior) = match &stored {
+                    Some(state) => {
+                        self.check_channel_open(state)?;
+                        (SetupForm::TopUp, Some(state))
```
EfeDurmaz16: Form is chosen from stored-state presence, so a retry of an `open` that confirmed but failed at `commit_voucher` is validated as `top_up` and rejected. Pick the form from the instruction discriminator.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
+            .first()
+            .ok_or_else(|| Error::Other("transaction has no signature slot".into()))?;
+        // Simulate the exact bytes before they reach the network. The static
+        // policy has already bounded what the sponsor is authorizing; this
+        // catches the rest — an unfunded payer, a frozen or wrong-owner token
+        // account, a settlement path that would not be usable later — while
+        // rejecting is still free.
+        let simulation = self
+            .rpc
+            .simulate_transaction(&tx)
```
EfeDurmaz16, line 1716: Simulating an already-landed tx returns `AlreadyProcessed`, so the signature-status fallback at L955 is unreachable and c98ab84's retry path never completes. Check `processed_topup_signatures` / `get_signature_status` before simulating.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
-    /// configured forced-close grace period: a non-zero voucher expiry must
-    /// outlast it so the voucher can still settle on-chain after the async
-    /// forced-close delay.
-    async fn accept(
+    /// The transaction was already statically validated during verification, so
+    /// this only adds the sponsor signature. A rejection is not authoritative on
+    /// its own: a retry of a transaction whose first submission landed dies at
+    /// preflight with "already processed", so the confirmed signature status
+    /// decides.
+    async fn broadcast_client_transaction(&self, transaction_b64: &str) -> Result<String, Error> {
```
EfeDurmaz16, line 1694: Doc comment says this catches "a settlement path that would not be usable later", but simulating the client's `open` doesn't check the payee/treasury/`payTo` ATAs the spec requires before escrowing.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
+            crate::x402::protocol::schemes::batch_settlement::classify(&error.to_string())
+                == codes::INVALID_CUMULATIVE_AMOUNT_MISMATCH;
+        if is_mismatch {
+            if let Some(channel_id) = self.channel_id_for_header(header) {
+                if let Ok(envelope) = self.corrective_challenge(amount, &channel_id).await {
+                    return Ok((PAYMENT_REQUIRED_HEADER.to_string(), encode_json(&envelope)?));
+                }
+            }
+        }
+        self.payment_required_header(amount)
```
EfeDurmaz16: Only cumulative mismatches get an `error` code; everything else falls back to a plain challenge, so the client can't tell `invalid_voucher_signature` from `duplicate_settlement`. `classify(err)` is already available — put it in the envelope.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +321,1277 @@ impl X402BatchSettlement {
...
+    /// returns `deposit - settled` to the payer, and closes the escrow token
+    /// account. Both run in one transaction per channel.
+    ///
+    /// Channels that are not yet due are skipped, and a channel another crank
+    /// already advanced is treated as success — the terminal onchain state is
+    /// what matters, not which worker got there.
+    ///
+    /// Claim before this runs: a voucher still unclaimed when the watermark
+    /// freezes is value the server forfeits to the payer.
+    pub async fn finalize_close(&self, channel_ids: &[String]) -> Result<Vec<String>, Error> {
```
EfeDurmaz16, line 2146: `finalize_close`/`reclaim` never mark sealed or delete, so `list_channels` keeps feeding finished channels into `claim`/`settle` (`channelNotDistributable`), compounding the L1184 issue.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -78,96 +104,202 @@ pub struct BatchConfig {
...
+    ///
+    /// The store holds the only record of what a client has been charged, so a
+    /// durable store is required in production — see [`Self::with_store`].
     pub fn new(config: BatchConfig) -> Result<Self, Error> {
         Self::with_store(config, Arc::new(MemoryChannelStore::new()))
     }
 
-    /// Build a handler with a caller-provided (e.g. durable) channel store.
+    /// Build a handler over a caller-provided (durable) channel store.
     pub fn with_store(config: BatchConfig, store: Arc<dyn ChannelStore>) -> Result<Self, Error> {
```
EfeDurmaz16, line 320: Validate `withdraw_delay` here; today a bad config advertises challenges nobody can pay.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -78,96 +104,202 @@ pub struct BatchConfig {
...
 
-/// Outcome of verifying a `batch-settlement` payment.
+/// Per-channel serialization guard.
+///
+/// The spec requires the server to serialize all paid-request and close
+/// processing per channel. The guard is held from `verify_payment` through
+/// `settle_payment`, so a second request on the same channel cannot read the
+/// watermark, serve, and commit in between.
+#[derive(Clone, Default)]
+struct InFlight(Arc<Mutex<HashSet<String>>>);
```
EfeDurmaz16, line 176: This guard is process-local, but the store is pluggable (Redis). With two replicas, both can verify the same voucher, both serve, and the second commit lands as a free replay via `accept_voucher`'s CAS. Either document that it only serializes within one process or move the reservation into the store (the TS port keeps a `pendingRequest` in channel state).

rust/crates/kit/src/gate.rs
```diff
@@ -906,46 +929,99 @@ async fn batch_gate_middleware(
...
-            HeaderName::from_bytes(name.as_bytes()),
-            HeaderValue::from_str(&value),
-        ) {
-            resp.headers_mut().insert(n, v);
+    // An idempotent retry of an already-accepted voucher. The spec requires the
+    // cached response for this commitment and forbids re-running the handler;
+    // this gate holds no response cache, so it reports the conflict rather than
+    // charge or execute twice. Applications needing true replay must cache on
+    // `extra.commitmentId` and short-circuit before the gate.
+    if outcome.replay {
```
EfeDurmaz16: 409 carries no `PAYMENT-RESPONSE`, so a client that lost the response re-signs the same voucher and is stuck. Attach the accepted response header so `apply_payment_response` can resync.

rust/crates/kit/src/core/payment_channels.rs
```diff
@@ -768,6 +787,284 @@ pub async fn build_open_payment_channel_tx_with_options(
...
+            lighthouse_count += 1;
+            if lighthouse_count > OPEN_MAX_LIGHTHOUSE_INSTRUCTIONS {
+                return Err(Error::Other(format!(
+                    "{label} transaction allows at most {OPEN_MAX_LIGHTHOUSE_INSTRUCTIONS} Lighthouse instructions after {label}"
+                )));
+            }
+            reject_fee_payer(ix, "Lighthouse")?;
+        } else if program == memo_program {
+            memo_count += 1;
+            if memo_count > 1 {
```
EfeDurmaz16, line 989: Behavior change for `upto`: the old scanner accepted several Memos; this rejects a second one. The `upto` spec only caps at one when `extra.memo` is present, and the PR body says `upto` is unchanged. Fine to keep, but say so in the PR body.

rust/crates/kit/examples/axum_batch.rs
```diff
@@ -47,26 +48,37 @@ async fn main() {
...
                 }
-                match batch.settle_batch(&channels).await {
-                    Ok(sigs) => {
-                        for id in &channels {
-                            let _ = batch.distribute(id).await;
-                        }
-                        tracing::info!(?sigs, "settled batch");
+                // `claim` advances the on-chain watermark (up to four channels
+                // per transaction); `settle` then pays the newly claimed delta.
+                match batch.claim(&channels).await {
```
EfeDurmaz16, line 72: Nit: nothing here runs `finalize_close`/`reclaim`, so refunds never finalize in this example.

EfeDurmaz16 (changes requested):
Wire types and the sponsor tx policy look solid and match the spec. Requesting changes for the redemption/retry paths: `claim()` breaks after its first success, refunds forfeit unclaimed voucher value, and confirmed-deposit retries still fail end to end.

README.md:199 and gate.rs:261 still reference `settle_batch` / `distribute`.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. claim now refreshes each channel from chain, persists the observed settled watermark, and skips closing/non-open channels before packing claims.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. Refunds now claim an unredeemed stored voucher before request_close. A Closing retry observes and records the existing close without rebroadcasting, and the on-chain state is first upserted so record_close has durable state.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. Setup form is now selected from the signed transaction instruction discriminator, then validated strictly in that form. A confirmed open whose voucher commit failed therefore retries as open rather than being misclassified as top_up.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. The transaction signature status is checked before simulation, so a previously confirmed setup transaction is accepted without hitting preflight AlreadyProcessed.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. Every verification failure now returns a machine-readable batch error code in the ordinary challenge; cumulative mismatches retain the authenticated corrective challenge.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. Successful finalize_close marks persisted state sealed, and successful reclaim deletes the terminal record, so future scans no longer re-submit terminal channels.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 7a81054c. Batch handler construction now validates withdraw_delay against the scheme range and max timeout, so invalid configurations cannot advertise unusable challenges.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Agreed. A store-backed reservation is the stronger long-term solution, but it requires extending the ChannelStore transaction contract. 7a81054c now explicitly documents the current process-local guarantee and the required channel-affinity deployment constraint for multiple replicas.

rust/crates/kit/src/gate.rs
lgalabru: Addressed in 7a81054c. The 409 idempotent-retry response now includes the authenticated PAYMENT-RESPONSE header reconstructed from stored channel state, allowing the client to resynchronize after a lost response.

rust/crates/kit/examples/axum_batch.rs
lgalabru: Addressed in 7a81054c. The example worker now invokes finalize_close and reclaim in addition to claim and settle, completing the demonstrated lifecycle.

rust/crates/kit/src/core/payment_channels.rs
lgalabru: Addressed in d8ba705d. README and public gate docs now name the current claim, settle, finalize_close, and reclaim lifecycle APIs.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in d189418e. Before a sponsor co-signs a deposit, the server now verifies the payee, treasury, and receiver settlement ATAs exist and are owned by the advertised token program. This is separate from setup simulation and prevents escrowing into a channel that cannot later distribute.

rust/crates/kit/src/core/payment_channels.rs
lgalabru: Agreed: this intentionally tightens upto to one Memo even when extra.memo is absent, so it is a behavior change despite the earlier summary. I am documenting that explicitly here for reviewers; it is not required for batch-settlement correctness.

lgalabru:
Compatibility note: this PR also intentionally tightens x402 upto transaction policy to reject a second Memo when extra.memo is absent. The previous scanner accepted multiple Memos; this is a behavior change, not a claim that upto is unchanged.

rust/crates/kit/src/x402/protocol/schemes/batch_settlement/types.rs
```diff
@@ -136,220 +178,637 @@ pub struct BatchRequiredEnvelope {
...
+    /// Parse the deposited amount.
+    pub fn amount(&self) -> Result<u64, Error> {
+        parse_amount(&self.amount, "deposit.amount")
+    }
+}
+
+/// The client authorization carried in `PAYMENT-SIGNATURE.payload`: a tagged
 /// union on `type`.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "camelCase")]
```
EfeDurmaz16: `rename_all = "camelCase"` does not rename struct-variant fields. I reproduced `channel_config` on the wire; the spec requires `channelConfig` (and `closeAuthorization`). Use `rename_all_fields = "camelCase"` or explicit renames, and assert the exact JSON keys.

rust/crates/kit/src/gate.rs
```diff
@@ -906,46 +929,103 @@ async fn batch_gate_middleware(
...
+            channel = %channel_id,
+            status = %resp.status(),
+            "handler failed; batch-settlement voucher left uncommitted"
+        );
+        return resp;
+    }
+
+    match batch.settle_payment(outcome).await {
+        Ok(response) => attach_settlement_header(&batch, &response, &mut resp),
+        Err(e) => {
```
EfeDurmaz16, line 1102: If settlement/store commit fails here, the handler has already run but the watermark and replay response are not persisted. Retrying the same voucher is treated as fresh and runs the handler again. Persist a served/pending result or otherwise return the original response without re-executing.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
+            extra: Some(BatchSettlementExtra {
+                commitment_id: Some(format!("{}:{}", outcome.channel_id, outcome.max_claimable)),
+                charged_amount: Some(outcome.charged_amount.to_string()),
+                channel_state: Some(self.snapshot(state)),
+            }),
+        }
+    }
+
+    /// Rebuild the accepted response for an idempotent voucher retry.
+    pub async fn replay_response(
```
EfeDurmaz16, line 1657: `replay_response` still reports `chargedAmount: "0"`, while the Rust client rejects anything other than the advertised amount at `client/batch_settlement/payment.rs:244`. Cache and return the original paid response; this 409 header still cannot resync this crate’s own client.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
-            };
-            if state.cumulative == 0 {
+                .ok_or_else(|| {
+                    batch_err(
+                        codes::INVALID_CHANNEL_STATE,
+                        format!("unknown channel {channel_id}"),
+                    )
+                })?;
+            let channel = pc::parse_pubkey(channel_id)?;
+            let onchain = self.fetch_channel(&channel)?;
```
EfeDurmaz16: A sealed `distribute` can deallocate the PDA. This fetch then aborts `claim`; the same issue exists in `settle` and `finalize_close`, while `reclaim` skips the error without deleting the store record. Treat a classified missing account as terminal and delete it, but not on transient RPC/decode errors.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
+    /// `distribute`. The channel stays open.
+    pub async fn settle(&self, channel_ids: &[String]) -> Result<Vec<String>, Error> {
+        let program_id = self.program_id()?;
+        let mint = self.mint()?;
+        let token_program = self.token_program()?;
+        let receiver = pc::parse_pubkey(&self.config.pay_to)?;
+        let mut groups = Vec::with_capacity(channel_ids.len());
+        for channel_id in channel_ids {
+            let channel = pc::parse_pubkey(channel_id)?;
+            let onchain = self.fetch_channel(&channel)?;
```
EfeDurmaz16: `settle` does not filter by channel status. A payer can move one channel to `Closing`, have it packed into `distribute`, and fail the atomic batch plus later work. Skip states that cannot be distributed before packing.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
-        let channel = self.fetch_channel(&channel_id)?;
-        if channel.status != CHANNEL_STATUS_OPEN {
-            return Err(Error::Other("channel is not open after broadcast".into()));
-        }
-        if pc::from_address(&channel.mint) != expected_mint {
-            return Err(Error::MintMismatch {
-                expected: pc::pubkey_string(&expected_mint),
-                actual: pc::pubkey_string(&pc::from_address(&channel.mint)),
-            });
+    fn check_channel_open(&self, state: &ChannelState) -> Result<(), Error> {
```
EfeDurmaz16, line 803: This only checks the local store. The payer can call `request_close` directly onchain, so the server may keep serving and committing vouchers that `claim` skips once it observes `Closing`. Reconcile and persist onchain close state before accepting another voucher.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
+                })
+            }
+            BatchPayload::Deposit {
+                voucher, deposit, ..
             } => {
-                self.process_voucher(&channel_id, voucher, per_request)
-                    .await
+                let max_claimable = check_voucher(voucher, &config, &channel_id)?;
+                let deposit_amount = deposit.amount()?;
+                let form = setup_form_from_transaction(&deposit.transaction, &program_id)?;
```
EfeDurmaz16, line 681: A confirmed `open` retry still runs fresh `openSlot` validation before `broadcast_client_transaction` can observe the landed signature. At L1011 the status lookup also excludes historical signatures. Check durable confirmation/channel state first and query signature history.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
-                    lifecycle: None,
-                    schema_version: CHANNEL_STATE_SCHEMA_VERSION,
-                    extra: Default::default(),
-                },
-            )
-            .await
-            .map_err(|e| Error::Other(format!("store error: {e}")))?;
+    /// Ensure a channel can later distribute its settled funds before the
+    /// sponsor co-signs the escrow transaction.
+    fn check_settlement_accounts(
```
EfeDurmaz16, line 927: This only checks that the ATA exists and its outer owner is the token program. The spec also requires rejecting frozen, wrong-mint, uninitialized, or otherwise program-unusable accounts before escrowing. Decode the token account and enforce those states/extensions here.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
-                voucher,
-            } => self.process_refund(&channel_id, voucher).await,
         }
     }
 
-    async fn process_deposit(
+    /// The client must echo back the requirements it is answering, so a payload
+    /// built against a different price or asset cannot be replayed onto this
+    /// route.
+    fn check_accepted_matches(
```
EfeDurmaz16, line 778: The spec requires `paymentPayload.accepted == paymentRequirements`, but this omits `maxTimeoutSeconds`, `paymentFlow`, `tokenProgram`, `receiverAuthorizer`, and `memo`; `x402Version` is not checked either. Compare the full normalized requirements and require v2.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
+
+    fn envelope(
+        &self,
+        requirement: BatchRequirements,
+        error: Option<String>,
+    ) -> BatchRequiredEnvelope {
+        BatchRequiredEnvelope {
             x402_version: X402_VERSION_V2,
-            resource,
+            resource: (!self.config.resource.is_empty()).then(|| ResourceInfo {
```
EfeDurmaz16: `PaymentRequired.resource` is required by x402 v2, but the default public path leaves `BatchConfig.resource` empty and omits it here. Require or derive the resource URL instead of emitting a non-conformant challenge.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1396 @@ impl X402BatchSettlement {
...
+                let signature = if channel.status == CHANNEL_STATUS_OPEN {
+                    let state = self
+                        .store
+                        .get_channel(&outcome.channel_id)
+                        .await
+                        .map_err(|e| Error::Other(format!("store error: {e}")))?;
+                    if state.is_some_and(|state| state.cumulative > channel.settlement.settled) {
+                        self.claim(std::slice::from_ref(&outcome.channel_id))
+                            .await?;
+                        channel = self.fetch_channel(&channel_id)?;
```
EfeDurmaz16: Clippy: this assignment is overwritten at L911 before being read. Drop it.

rust/crates/kit/src/mpp/protocol/solana.rs
```diff
@@ -1052,26 +1064,20 @@ pub fn validate_splits(splits: &[Split]) -> Result<(), crate::mpp::error::Error>
         return Err(Error::TooManySplits);
     }
 
-    let mut seen_recipients: HashSet<&str> = HashSet::with_capacity(splits.len());
+    let mut seen_legs: HashSet<(&str, Option<&str>)> = HashSet::with_capacity(splits.len());
```
EfeDurmaz16, line 1078: This unrelated commit should be split out. It also treats `(recipient, None)` and `(recipient, Some(""))` as distinct, while settlement drops empty expected memos; a Rust-issued challenge can then fail with the Rust client. Normalize or reject empty memos and add an end-to-end verify vector in the separate PR.

EfeDurmaz16 (changes requested):
Most first-round redemption and retry fixes landed, and the upto compatibility note matches the current spec. Requesting changes for the remaining wire, replay, and lifecycle gaps: Rust emits a non-conformant batch payload shape, exact retries can still re-execute or fail client resync, and Closing/deallocated channels can poison redemption and distribution. Settlement-account preflight and confirmed-open retry are also incomplete.

rust/crates/kit/src/x402/protocol/schemes/batch_settlement/types.rs
lgalabru: Addressed in 70741ac1. BatchPayload now uses serde rename_all_fields camelCase, with regression assertions for channelConfig and closeAuthorization and against the snake_case wire keys.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Addressed in 2686e71d. Verification now requires x402 v2 and compares maxTimeoutSeconds, paymentFlow, tokenProgram, receiverAuthorizer, and memo in addition to the existing immutable requirement fields.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +324,1225 @@ impl X402BatchSettlement {
...
+                // channel does not exist yet for an `open`.
+                let (form, deposit_after, prior) = match &stored {
+                    Some(state) => {
+                        self.check_channel_open(state)?;
+                        let after = state.deposit.checked_add(deposit_amount).ok_or_else(|| {
+                            batch_err(codes::INVALID_CHANNEL_STATE, "deposit overflow")
+                        })?;
+                        (SetupForm::TopUp, after, Some(state))
+                    }
+                    None => (SetupForm::Open, deposit_amount, None),
```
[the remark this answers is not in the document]
lgalabru: Addressed in d189418e. Setup retries now use the payer transaction signature to recognize an already-recorded top-up, so `deposit_ceiling` retains the confirmed balance instead of adding the same deposit again. `a_confirmed_deposit_is_not_counted_twice_on_retry` covers this case.

rust/crates/kit/src/gate.rs
```diff
@@ -906,46 +929,103 @@ async fn batch_gate_middleware(
...
+            // The resource was already served, so the response stands; the
+            // uncommitted charge is the server's loss, and worth an alert.
+            tracing::error!(
+                channel = %channel_id,
+                error = %e,
+                "batch-settlement commit failed after the resource was served"
+            );
         }
     }
     resp
```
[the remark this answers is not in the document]
lgalabru: Confirmed and fixed in 2d9d51e2. The gate now calls `settle_payment` before invoking the protected handler, as required by the batch-settlement commitment lifecycle. If commitment storage fails, it returns 502 without running the handler; once access is granted, the voucher watermark is already durable and the authorization cannot be replayed.

rust/crates/kit/src/gate.rs
```diff
@@ -906,46 +929,100 @@ async fn batch_gate_middleware(
...
+            }
+        };
+    }
+
+    let channel_id = outcome.channel_id.clone();
+    // `batch-settlement` commits the authorization before granting access. In
+    // particular, do not run the handler until the voucher watermark has been
+    // durably advanced: releasing an uncommitted outcome after a successful
+    // handler would allow the same voucher to authorize another execution.
+    let settlement = match batch.settle_payment(outcome).await {
```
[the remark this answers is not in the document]
lgalabru: This ordering is intentional. The batch-settlement lifecycle requires the payment commitment to be validated and stored before access is granted; committing only after the handler succeeds reintroduces the previous P1, where a successful resource execution can be replayed if durable commitment fails. A handler 5xx therefore does not roll back an already accepted authorization. Returning a cached handler response for exact replay is a separate gate capability (the generic Axum gate does not retain response bodies), but it cannot safely make commitment storage conditional on handler success.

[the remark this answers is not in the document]
lgalabru: Intentional for this scheme: the spec requires committing the payment authorization before access.
Moving commitment after the handler would restore the prior replay vulnerability.

EfeDurmaz16: The resolved flow is `authorization`: verify → resource → settle → respond. This commits before `next.run`, so a handler failure consumes the payment and the retry cannot receive the resource. Please implement the durable post-handler commit and cached-response path instead of precharging the request.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1431 @@ impl X402BatchSettlement {
...
-        let channel = self.fetch_channel(&channel_id)?;
-        if channel.status != CHANNEL_STATUS_OPEN {
-            return Err(Error::Other("channel is not open after broadcast".into()));
-        }
-        if pc::from_address(&channel.mint) != expected_mint {
-            return Err(Error::MintMismatch {
-                expected: pc::pubkey_string(&expected_mint),
-                actual: pc::pubkey_string(&pc::from_address(&channel.mint)),
-            });
+    fn check_channel_open(&self, state: &ChannelState) -> Result<(), Error> {
```
EfeDurmaz16, line 803: This only checks stored state. A payer can call `request_close` directly on-chain, after which the server can still accept and serve a new voucher that `claim` cannot redeem because program `settle` requires `Open`. Reconcile on-chain status before accepting the voucher.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1431 @@ impl X402BatchSettlement {
...
-            };
-            if state.cumulative == 0 {
+                .ok_or_else(|| {
+                    batch_err(
+                        codes::INVALID_CHANNEL_STATE,
+                        format!("unknown channel {channel_id}"),
+                    )
+                })?;
+            let channel = pc::parse_pubkey(channel_id)?;
+            let onchain = self.fetch_channel(&channel)?;
```
EfeDurmaz16: A finalized sealed `distribute` can deallocate the PDA immediately. This fetch then aborts `claim`; `settle` and `finalize_close` do the same, while `reclaim` skips the error without deleting the stored row. Treat confirmed account absence as terminal without collapsing transient RPC or decode errors into absence.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +330,1431 @@ impl X402BatchSettlement {
...
-                    lifecycle: None,
-                    schema_version: CHANNEL_STATE_SCHEMA_VERSION,
-                    extra: Default::default(),
-                },
-            )
-            .await
-            .map_err(|e| Error::Other(format!("store error: {e}")))?;
+    /// Ensure a channel can later distribute its settled funds before the
+    /// sponsor co-signs the escrow transaction.
+    fn check_settlement_accounts(
```
EfeDurmaz16, line 927: This checks only existence and outer owner for three ATAs, and omits the payer return ATA. The spec requires rejecting frozen, uninitialized, wrong-mint, or otherwise unusable settlement accounts before escrow. Decode and validate every settlement ATA here.

EfeDurmaz16 (changes requested):
Wire and retry fixes now look solid. Requesting changes for four remaining runtime paths: the gate commits authorization before the handler, direct on-chain closes can make newly accepted vouchers unredeemable, deallocated channel records can poison lifecycle workers, and setup preflight still does not validate settlement ATA health.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +380,2201 @@ impl X402BatchSettlement {
...
+                    tracing::warn!(channel = %channel_b58, error = %e, age, "serving on a stale channel snapshot");
+                    return Ok(());
+                }
+                return Err(e);
+            }
+        };
+        let closed_at = match onchain {
+            // Confirmed absent: reclaimed, or never opened. Either way no
+            // voucher against it can ever be redeemed.
+            None => {
```
EfeDurmaz16, line 1534: An initial `open` has a store row before its PDA exists. After a crash before `finish_commit`, this expected absence blocks `Resume`; `lookup_for_lifecycle` can also delete the live reservation. Preserve rows with a pending setup or authorization.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -189,794 +380,2201 @@ impl X402BatchSettlement {
...
+/// The fields of an SPL token account this server cares about.
+struct TokenAccount {
+    mint: Pubkey,
+    owner: Pubkey,
+    state: u8,
+}
+
+/// Decode a token account, or `None` when the data is too short to be one —
+/// which is what an uninitialized (or non-token) account looks like.
+fn decode_token_account(data: &[u8]) -> Option<TokenAccount> {
```
EfeDurmaz16, line 2382: This decodes only the base token-account layout, so unsupported Token-2022 extensions pass. The program rejects them for treasury accounts and redirects recipient or payer payouts to treasury. Validate AccountType and TLV extensions under the program policy.

EfeDurmaz16 (changes requested):
Latest recovery commit looks sound. Requesting changes for two remaining runtime paths: a pending open can be mistaken for a deleted channel before post-handler commit, and Token-2022 settlement preflight accepts account extensions the program rejects or redirects.

docs/batch-settlement-durable-authorization-plan.md
```diff
@@ -0,0 +1,163 @@
+# Batch settlement durable authorization plan
```
EfeDurmaz16: This is an internal implementation plan, and it already differs from the landed design. Please leave it out of this PR.

typescript/pnpm-workspace.yaml
```diff
@@ -19,11 +19,11 @@ overrides:
   '@hono/node-server': '>=2.0.10'
   # fast-uri/hono/ip-address floors below track advisories reached via
   # @modelcontextprotocol/sdk (see `pnpm why <pkg>`); bump as new CVEs land.
-  fast-uri: '>=4.1.2'
+  fast-uri: '>=4.1.3'
```
EfeDurmaz16: `fast-uri` 4.1.3 is still affected by both advisories; the patched v4 release is 4.1.4. Please raise the override and lockfile to 4.1.4.

rust/crates/kit/src/gate.rs
```diff
@@ -905,45 +969,168 @@ async fn batch_gate_middleware(
...
+    // failure here is logged and the charge is attempted anyway — otherwise a
+    // store hiccup would serve the request for free and leave the client's
+    // channel stuck at a watermark nothing can advance.
+    if let Err(e) = batch.mark_handler_succeeded(&outcome).await {
+        tracing::error!(
+            channel = %channel_id,
+            error = %e,
+            "batch-settlement could not record a served handler; charging anyway"
+        );
+    }
```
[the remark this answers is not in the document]
lgalabru: This is deliberate, and it is the opposite of the earlier P1 on this same line.

That review asked for expired unmarked reservations to stop being reclaimable, naming this exact case — *"adds a distinct retry path when the process crashes **or the marker write fails after successful execution**"*. They were made terminal in `1e03d9d7`. Resuming the charge on retry, as this comment asks, requires the reservation to be reclaimable again, which reopens that finding: §Phase 5 of the SVM `batch-settlement` spec makes it a MUST that one authorization never runs the resource handler twice.

When the durable store will not accept a write there is no third option — either the authorization stays usable (risking a second execution) or it goes terminal (risking an uncharged served request). The spec settles which way that falls, and the money direction follows it: the payer is never charged for what it did not receive, and the operator absorbs one served request during its own store outage.

Two details in the comment do not hold:

- `handler_succeeded = false` does not block the charge. `commit_authorization` never reads that field — which is precisely why the previous fix (committing even when the marker write fails) works.
- The window needs *both* writes to fail. They go to the same store, so this is a store outage, and nothing can be durably recorded then, marker or otherwise.

There was a real improvement in it, applied in 419407e3: the charge is idempotent — committing an already-committed authorization is a no-op, and a deposit's setup transaction is recovered by its signature rather than broadcast again — so it now retries with backoff instead of surrendering on the first error. A store that blinks costs latency rather than the charge. The residual case (sustained outage, or the process exiting mid-charge) is now written down at the call site rather than implied.

rust/crates/kit/src/x402/server/batch_settlement.rs
EfeDurmaz16: The lifecycle deletion half is addressed, but the original pending-open recovery is still blocked in two places. A retry after `handler_succeeded` but before the initial `open` is broadcast fails in `reconcile_channel` on the expected missing PDA, before `reserve` can return `Resume`. Separately, `lookup_for_lifecycle` reads chain absence and store state non-atomically, so concurrent `finish_commit` can open and commit between them, after which the lifecycle path deletes the newly committed watermark and replay record. Please make pending-open absence resumable and make terminal deletion conditional on unchanged state. @lgalabru

EfeDurmaz16 (approved):
LGTM!

PR #304, opened 2026-09-03
perf(x402/batch): take the batch-settlement hot path to mpp-parity (~42k → ~888k req/s)

lgalabru:
## Summary

Takes the x402 `batch-settlement` server hot path from **~42k → ~888k req/s** (2000-channel closed loop, AVX-512 host) with no protocol or API change. After these three fixes the scheme is **ed25519-verify-bound, at parity with mpp-session** (curve25519 verify ~28% of on-CPU time) — the same ceiling, where it should be.

Built on top of #301 (the SVM-spec rewrite). Three commits, most-impactful last:

| # | commit | change | effect |
|---|--------|--------|--------|
| 1 | `32a59b71` | run the stale-snapshot RPC lookup on `spawn_blocking` instead of the async runtime | correctness/latency hygiene |
| 2 | `d888174a` | shard the per-request in-flight guard (`Mutex<HashSet>` → `DashSet`) | contention hygiene |
| 3 | `5250d824` | **eliminate the per-request `ChannelState` clone** | **the 8x** |

## The bottleneck (measured, not guessed)

A flamegraph of the loaded gateway showed ~33% of CPU in futex/arena-lock traffic under the glibc allocator, not in verification. Root cause: the batch 2-phase `reserve → commit` path plus the read path deep-cloned the **entire** `ChannelState` — `String`s, the `serde_json` `extra` map, and the delivery `Vec`s — roughly **4× per request** (`get_channel` clones on read; `update_channel` clones twice, once for the closure argument and once on insert). Because `committed_deliveries` grows per commit, the clone cost was superlinear over a channel's life.

mpp-session never hit this: its hot path is a single `u64` watermark CAS (`record_committed_watermark` / `advance_cumulative`) with ~0 allocation. Same verify, far lighter state mutation.

> Two earlier hypotheses (the reconcile RPC on the runtime, the in-flight `Mutex`) were fixed first — commits 1 and 2 — but profiling showed **neither** was the bottleneck. They are real hygiene wins and worth keeping; they just weren't the 8x. The clone was.

## The fix

Two new zero-clone accessors on the `ChannelStore` trait, replacing the clone-heavy `get_channel` / `update_channel` on every hot path:

- `read_channel(id, FnOnce(Option<&ChannelState>) -> R)` — borrows in place; callers copy out only the handful of scalars they need.
- `mutate_channel(id, seed: Option<ChannelState>, FnOnce(&mut ChannelState))` — mutates in place, seeding a new entry only on first touch.

Implemented for `MemoryChannelStore` (DashMap `get_mut`, no clone), `RedisChannelStore` (decode → mutate → CAS), and the `Arc<T>` blanket impl. `batch_settlement.rs` reserve / commit / verify (both the Voucher and Refund arms) / reconcile are rewritten to use them and to pass scalars (`check_channel_open` / `check_watermark`) instead of `&ChannelState`.

Once the allocator pressure is gone, `mimalloc` on the gateway side (in the `pay` consumer, not this crate) is what converts the reduced allocation into the full win — but the clone removal is the load-bearing change and stands on its own.

## Testing

- `cargo test -p pay-kit` — 414/414 pass.
- No public API change beyond the two additive trait methods (existing `get_channel`/`update_channel` retained).
- Verified end-to-end against a `pay` gateway (#454) under a 2000-channel closed-loop load: 42k → 888k req/s, p50 191ms → 2.0ms.

### DIFF
--- a/rust/crates/kit/src/core/store.rs
+++ b/rust/crates/kit/src/core/store.rs
@@ -1024,6 +1024,38 @@ pub trait ChannelStore: Send + Sync {
         updater: Box<dyn FnOnce(Option<ChannelState>) -> Result<ChannelState, StoreError> + Send>,
     ) -> Pin<Box<dyn Future<Output = Result<ChannelState, StoreError>> + Send + '_>>;
 
+    /// Read channel state in place, without cloning the record.
+    ///
+    /// The `reader` receives `Option<&ChannelState>` (borrowed under the read
+    /// guard) and extracts only the small values the caller needs, so a hot
+    /// path can inspect a few fields without deep-cloning the whole state
+    /// (String fields, the `extra` map, and the growing delivery Vecs). The
+    /// reader MUST NOT re-enter the store for the same key.
+    fn read_channel(
+        &self,
+        channel_id: &str,
+        reader: Box<dyn FnOnce(Option<&ChannelState>) -> Result<(), StoreError> + Send>,
+    ) -> Pin<Box<dyn Future<Output = Result<(), StoreError>> + Send + '_>>;
+
+    /// Atomically mutate channel state in place, without cloning the record.
+    ///
+    /// This is the allocation-free counterpart to [`Self::update_channel`] for
+    /// the batch-settlement hot path. The `mutator` receives `&mut ChannelState`
+    /// and returns only a small caller-selected result (typically captured out
+    /// through the closure), so an in-memory backend can apply it under the
+    /// shard write guard with ZERO clone of the state. Implementations MUST
+    /// still guarantee the read-modify-write is atomic per channel.
+    ///
+    /// When the channel is absent, `seed` (if provided) is inserted first and
+    /// then mutated in place; if the channel is absent and `seed` is `None`,
+    /// the call fails. The mutator MUST NOT re-enter the store for the same key.
+    fn mutate_channel(
+        &self,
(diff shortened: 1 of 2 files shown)


lgalabru:
Both remaining failures (**Go tests** — `TestPlaygroundSessionE2ESurfpool` / `TestSessionServerE2ESurfpool`, and **Harness: on-chain settlement (mainnet fork)**) are environmental, not caused by this diff:

- Both fail with `429 Too Many Requests` from `mainnet.helius-rpc.com` (Go) and `surfnet_setTokenAccount: Internal error` (fork harness), on the *same* Helius API key.
- That key is currently saturated by an unrelated devnet load-test I'm running for SUNBURST (100k channel-open provisioning, ~200 concurrent). Helius appears to rate-limit the key's plan across mainnet+devnet together.
- Both suites are green on `main` as of this branch's base, and neither exercises Rust — they're Go SDK e2e and a TypeScript on-chain harness. This PR's diff is Rust-only (two additive `ChannelStore` trait methods + the batch hot-path rewrite that uses them); `cargo test -p pay-kit` is 414/414 green.
- Re-running now would just hit the same 429s — the devnet load-test finishes provisioning in ~1-2h and stops calling Helius entirely once it enters its measured window, at which point I'll re-run these two jobs.

EfeDurmaz16 (approved):

PR #307, opened 2026-09-04
fix(x402): use the deployed program's per-cluster treasury owner

lgalabru:
## Summary
- `distribute` validates the treasury token account against `ATA(TREASURY_OWNER, mint, token_program)`, but the on-chain payment-channels program bakes in a **distinct `TREASURY_OWNER` per cluster feature build** (mainnet-beta/devnet/testnet — see the program's `README.md`/`constants.rs`). pay-kit's Rust SDK hardcoded the mainnet-build constant everywhere, so every batch-settlement/upto `distribute` against a devnet deployment failed with `TreasuryAccountMismatch` (0x961).
- Adds `DEVNET_TREASURY_OWNER` and `treasury_owner_for_cluster(cluster)` to `core/payment_channels.rs`, mirroring the TypeScript SDK's existing `getPaymentChannelsTreasuryOwner(network)`. Updates the 5 call sites in `batch_settlement.rs` and `upto.rs` that previously called the unconditional `treasury_owner()`.
- Both `BatchConfig` and `UptoConfig` already carry a `cluster: String` field (`"devnet"` is what the live gateway's `network.slug()` produces), so this is a pure bug fix — no config/API surface change for callers.

## Verification
- `cargo test -p solana-pay-kit --release`: 414/414 passing.
- `cargo fmt -p solana-pay-kit -- --check`: clean.
- Empirically verified against a live devnet payment-channels deployment (program `CHNLxYvVA28MJP9PrFuDXccuoGXAx7jBacfLEkahyGsX`): a real `distribute` call that previously failed with `custom program error: 0x961` now succeeds and fully deallocates the channel PDA (rent returned), using the corrected devnet treasury owner.

### DIFF
--- a/rust/crates/harness-bins/src/bin/x402_harness_upto_server.rs
+++ b/rust/crates/harness-bins/src/bin/x402_harness_upto_server.rs
@@ -70,7 +70,13 @@ fn read_state(
     runtime: Arc<tokio::runtime::Runtime>,
 ) -> Result<UptoHarnessState, Box<dyn std::error::Error + Send + Sync>> {
     let rpc_url = read_required_env("X402_HARNESS_RPC_URL")?;
-    let network = env::var("X402_HARNESS_NETWORK").unwrap_or_else(|_| "devnet".to_string());
+    // This harness runs against a local Surfpool validator whose
+    // payment-channels build has its treasury-owner sentinel patched to the
+    // *mainnet* constant (see .github/actions/build-payment-channels) — it
+    // is not real Solana devnet, so the label must not be "devnet" or
+    // "solana-devnet": treasury_owner_for_cluster() would then look up the
+    // real devnet treasury and mismatch this harness's mainnet-patched one.
+    let network = env::var("X402_HARNESS_NETWORK").unwrap_or_else(|_| "localnet".to_string());
     let mint = env::var("X402_HARNESS_MINT")
         .unwrap_or_else(|_| "4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU".to_string());
     let pay_to = read_required_env("X402_HARNESS_PAY_TO")?;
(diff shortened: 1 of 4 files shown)


lgalabru:
Re: the testnet concern (Confidence 4/5) — pushing back on the severity, the underlying observation is correct but doesn't block merge:

**`treasury_owner_for_cluster` falling through to mainnet for testnet matches the existing TypeScript reference SDK exactly**, not a gap unique to this PR. `getPaymentChannelsTreasuryOwner` in `mechanisms/svm/src/payment-channels/onchain.ts` only special-cases devnet (`SOLANA_DEVNET_CAIP2` / `"solana-devnet"`) and falls through to `MAINNET_TREASURY_OWNER` for everything else, testnet included. This Rust change has parity with that shipped behavior, not a regression from it.

**More importantly, a testnet build of the on-chain program cannot compile today**, let alone be deployed — `program/payment_channels/src/constants.rs` still has testnet's `TREASURY_OWNER` set to `TREASURY_OWNER_SENTINEL` (`// TODO: real testnet owner`), and the crate has a build-time assertion that any `devnet`/`testnet`/`mainnet-beta` feature build with the sentinel still in place fails to compile. So there is no real testnet deployment this could mismatch against — the failure mode described ("will fail on-chain validation") can't occur because there's nothing to validate against yet.

Devnet's constant made it into this PR because I found and empirically verified a real deployed value (the TS SDK's `DEVNET_TREASURY_OWNER`, confirmed against a live devnet payment-channels deployment). No equivalent testnet value exists anywhere I can find — not in this SDK, the TS SDK, or the on-chain program repo. Adding one would mean guessing a constant with no way to verify it's correct, which is worse than the current explicit mainnet fallback.

This PR's scope (per its title/description) is the devnet fix specifically. When a real testnet deployment exists with a documented treasury owner, adding `TESTNET_TREASURY_OWNER` + a match arm is a one-line follow-up using the exact same pattern established here.

lgalabru:
Superseded by #308. PR #308 contains the equivalent treasury-owner and concurrent-submission changes, and now also includes this PR’s unique local Surfpool harness-label fix in commit 268bc2ee.

PR #308, opened 2026-09-04
feat(x402/batch): harden replay and settlement lifecycle

lgalabru:
## Summary

- Cache bounded upstream responses so an idempotent batch-settlement replay returns the purchased representation without rerunning the handler or charging twice.
- Make batch channel opens, top-ups, voucher commits, lifecycle reconciliation, settlement, close, and reclaim resumable and safe across ambiguous RPC outcomes.
- Reuse deployment-specific treasury owners, batch lifecycle account reads, and bounded concurrent settlement submission.
- Preserve distributed watermarks and surface simulation logs for actionable failures.

PR #307 is fully consolidated here, including its unique local Surfpool harness network fix.

## Review fixes

- Preserve replayable representation headers alongside status, content type, and body.
- Fetch a fresh blockhash per concurrently submitted settlement transaction and drain every started submission outcome.
- Bind new-channel outcomes to the configured program ID.
- Poll ambiguous deposit signatures through delayed RPC visibility, including retry simulation failures.
- Allow an absent on-chain channel only for an unexpired setup that actually opens it.

## Test Plan

- `cargo test -p solana-pay-kit --lib --features server,client` (955 passed)
- `cargo test -p solana-pay-kit --lib --features server,client x402::server::batch_settlement::tests` (22 passed)
- Full GitHub Actions matrix


rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -2033,8 +2328,9 @@ impl X402BatchSettlement {
     /// Returns the confirmed transaction signatures.
     pub async fn claim(&self, channel_ids: &[String]) -> Result<Vec<String>, Error> {
         let program_id = self.program_id()?;
+        let onchain_channels = self.lookup_channels_for_lifecycle(channel_ids).await?;
         let mut groups = Vec::with_capacity(channel_ids.len());
-        for channel_id in channel_ids {
+        for (channel_id, onchain) in channel_ids.iter().zip(onchain_channels) {
```
EfeDurmaz16, line 2365: Could we move the `None` guard above the local store lookup here? `lookup_channels_for_lifecycle()` now deletes records for absent onchain accounts, so `get_channel()` fails before the existing guard can skip them. One reclaimed channel therefore aborts the whole `claim()` call, including valid peers whose transactions have not been submitted yet.

The fix is just moving the existing guard:

```diff
 for (channel_id, onchain) in channel_ids.iter().zip(onchain_channels) {
+    let Some(onchain) = onchain else {
+        continue;
+    };
     let mut state = self
         .store
         .get_channel(channel_id)
         // ... existing error handling ...
     let channel = pc::parse_pubkey(channel_id)?;
-    let Some(onchain) = onchain else {
-        continue;
-    };
```

Verified with a mocked absent-account regression test: it fails with `unknown channel` on this head and passes after this reorder. A fresh channel listing can recover on the next tick, but retrying the same ID list keeps failing. Worth keeping that test alongside the fix.

EfeDurmaz16 (dismissed):
Approved. The inline claim cleanup fix can be handled in a follow-up.

rust/crates/kit/src/x402/server/batch_settlement.rs
```diff
@@ -767,7 +959,7 @@ impl X402BatchSettlement {
                         )
                     })?;
                 self.check_channel_open(sealed, close_requested)?;
-                let max_claimable = check_voucher(voucher, &config, &channel_id)?;
+                let max_claimable = check_voucher_batched(voucher, &config, &channel_id).await?;
```
EfeDurmaz16, line 962: With `PAY_VERIFY_BATCH=1`, mixed-order `R` signatures can pass this batch path even though `verify_strict` rejects them. I reproduced this with a regression test; checking `point.is_small_order() || !point.is_torsion_free()` closes the gap and keeps the full test suite green.

lgalabru: Fixed in 24fc7051. The batch prefilter now rejects any R that is small-order or not torsion-free. Added a regression that constructs a mixed-order R, finds a transcript coefficient for which ed25519_dalek::verify_batch accepts it, confirms verify_strict rejects it, and confirms our guard rejects it. cargo fmt --check and all 419 solana-pay-kit library tests pass.

rust/crates/kit/src/x402/server/batch_settlement.rs
lgalabru: Fixed in f456cf1c: confirmed-absent/reclaimed channels are skipped before the local store lookup, so they cannot abort claims for valid peers. cargo fmt --check and all 419 solana-pay-kit library tests pass.

rust/crates/kit/src/core/store.rs
```diff
@@ -952,15 +1028,19 @@ impl ChannelState {
...
-            self.committed_deliveries
-                .sort_by_key(|entry| entry.cumulative);
-            self.committed_deliveries.drain(..excess);
+        while self
+            .committed_deliveries
+            .front()
+            .is_some_and(|entry| entry.retain_until != 0 && entry.retain_until <= now)
+        {
+            self.committed_deliveries.pop_front();
+        }
```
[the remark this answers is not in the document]
lgalabru: Already fixed in 5cc2529c. commit_authorization clamps each retain_until to the previous deque tail, making expiries non-decreasing, and committed_authorization_expiry_remains_ordered tests the invariant and prefix pruning. The final Greptile review also confirms expiry remains ordered and bounded.

EfeDurmaz16 (approved):
LGTM, thanks.

PR #309, opened 2026-09-04
fix(ci): follow renamed Surfpool report-action branch

ledgicr:
Repairs a pre-existing failure on `main`. Unrelated to #300, and kept out of that diff deliberately.

## What is broken

`report.yml:80` checks out `solana-foundation/surfpool` to fetch the reusable `test-report` action the workflow runs at line 99. It pins `ref: feat/sdk`, and that branch no longer exists upstream. It was renamed to `feat/sdk-report`. So the Report job dies at the checkout step before it can post anything:

```
A branch or tag with the name 'feat/sdk' could not be found
```

All six of the most recent Report runs failed there: [33924654188](https://github.com/solana-foundation/pay-kit/actions/runs/33924654188), [33920537042](https://github.com/solana-foundation/pay-kit/actions/runs/33920537042), [33919781322](https://github.com/solana-foundation/pay-kit/actions/runs/33919781322), [33919650125](https://github.com/solana-foundation/pay-kit/actions/runs/33919650125), [33919589352](https://github.com/solana-foundation/pay-kit/actions/runs/33919589352), [33918248350](https://github.com/solana-foundation/pay-kit/actions/runs/33918248350). Each carries that same annotation, so this started when surfpool renamed the branch rather than with any change here.

## The fix

One line:

```diff
-          ref: feat/sdk
+          ref: feat/sdk-report
```

## Why that ref and not a stable one

- `solana-foundation/surfpool` has no `feat/sdk`. The branches API returns 404.
- `feat/sdk-report` exists, currently at `e0204c21`.
- `.github/actions/test-report/action.yml` is present on `feat/sdk-report` and absent from surfpool's `main`, so the workflow genuinely needs the feature branch. Pointing it at `main` would swap one missing-path failure for another.

Pinning a moving branch in another repository is what made this break, and it will break again the next time that branch is renamed or deleted. Worth pinning to a tag or a commit SHA, or vendoring the action, but that is a bigger call than this fix and I did not want to bundle it in. Happy to follow up if you want it.

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -77,7 +77,13 @@ jobs:
       - uses: actions/checkout@v5
         with:
           repository: solana-foundation/surfpool
-          ref: feat/sdk
+          # Pinned to a commit, not a branch. This workflow broke because it
+          # followed `feat/sdk`, which surfpool renamed to `feat/sdk-report`;
+          # following the new branch would leave the same failure mode in place,
+          # and a moving ref in another repository is also an unreviewed input to
+          # our CI. e0204c21 is the head of `feat/sdk-report` as of 2026-09-05.
+          # Bump deliberately when that branch advances.
+          ref: e0204c21c59e225fab81075a91ffc8d9c0b20d02
           sparse-checkout: .github/actions
           path: _surfpool
 


ledgicr:
Hardened to a commit pin rather than the renamed branch.

Following `feat/sdk-report` would fix today's symptom and leave the failure mode intact: the next rename or deletion breaks CI again, with no change here to review. A moving ref in another repository is also an unreviewed input to a workflow step that runs their code. `e0204c21` is the head of `feat/sdk-report` as of 2026-09-05, and the branch it came from is recorded in a comment above the pin so the next person to touch it does not helpfully simplify it back to a name.

EfeDurmaz16 (approved):
Looks good, thanks!

PR #314, opened 2026-09-09
fix(mpp): harden payment verification

lgalabru:
## Summary

- enforce strict subscription activation instruction and account allowlists in Rust and TypeScript, including binding the transfer receiver to the challenged recipient
- add challenge-bound pending/confirmed replay state for TypeScript charge and subscription settlement, with idempotent receipt recovery
- require Lua x402 confirmation before success and preserve safe pending/confirmed recovery state
- keep Python x402 replay markers durable after ambiguous post-broadcast confirmation failures
- refresh the subscriptions IDL and Codama-generated Rust client from upstream v0.5.0, and use the latest official TypeScript account decoder
- bump each SDK incrementally for the next release: TypeScript 0.10.0, Python 0.8.0, and Rust 0.7.0
- refresh the harness lockfile and raise the Hono override to its patched 4.13.5 release

Compatibility note: the v0.5.0 TypeScript subscriptions package declares a Kit 7 peer while MPP remains on Kit 6.10. This PR only uses its isolated account decoder; instruction-builder adoption should land with the Kit 7 migration.

## Test Plan

- TypeScript: 528 tests; lint, typecheck, formatting, and both package builds
- Rust: 73 subscription-focused tests with client and server features
- Python: 1242 passed, 1 skipped; Pyright and Ruff checks clean for the touched files
- Lua: lint clean; 603 passed, 1 skipped
- TypeScript harness: typecheck and server-package startup import
- production dependency audit: no known vulnerabilities
- existing Go, Kotlin, PHP, Ruby, and Swift CI suites

### DIFF
--- a/lua/pay_kit/protocols/x402/init.lua
+++ b/lua/pay_kit/protocols/x402/init.lua
@@ -31,6 +31,7 @@ local tx_cosign  = require('pay_kit.solana.tx_cosign')
 local x402_verify = require('pay_kit.protocols.x402.exact.verify')
 local tx_mod     = require('pay_kit.solana.transaction')
 local network_check = require('pay_kit.solana.network_check')
+local json        = require('pay_kit.util.json')
 
 local M = {}
 local Adapter = {}
@@ -63,6 +64,8 @@ local LEGACY_PAYMENT_RESPONSE_HEADER = 'x-payment-response'
 local DEFAULT_FIXTURE_SETTLEMENT_HEADER = 'x-payment-settlement-signature'
 local DEFAULT_MAX_TIMEOUT_SECONDS = 60
 local DEFAULT_DECIMALS = 6
+local DEFAULT_CONFIRMATION_ATTEMPTS = 40
+local DEFAULT_CONFIRMATION_DELAY_SECONDS = 0.25
 local TOKEN_PROGRAM_BASE58 = 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA'
 
 local CAIP2_MAINNET = 'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp'
@@ -370,6 +373,39 @@ local function consume_signature(store, signature)
   return true
 end
 
+local function sleep_seconds(seconds)
+  local ngx_global = rawget(_G, 'ngx')
+  if ngx_global and type(ngx_global.sleep) == 'function' then
+    return ngx_global.sleep(seconds)
+  end
+  local ok, socket = pcall(require, 'socket')
+  if ok and socket and type(socket.sleep) == 'function' then
+    return socket.sleep(seconds)
+  end
+  local deadline = os.clock() + seconds
+  while os.clock() < deadline do end
+end
+
+local function await_confirmation(rpc, signature)
+  for attempt = 1, DEFAULT_CONFIRMATION_ATTEMPTS do
+    local statuses = rpc:signature_statuses({signature})
+    local status = statuses and statuses[1]
+    if type(status) == 'table' then
+      if status.err ~= nil and status.err ~= json.null then
+        return nil, 'transaction failed on-chain: ' .. tostring(status.err)
+      end
+      if status.confirmationStatus == 'confirmed' or
+         status.confirmationStatus == 'finalized' then
+        return true
+      end
+    end
+    if attempt < DEFAULT_CONFIRMATION_ATTEMPTS then
+      sleep_seconds(DEFAULT_CONFIRMATION_DELAY_SECONDS)
+    end
+  end
+  return nil, 'timed out waiting for transaction confirmation'
+end
+
 -- --- public API -----------------------------------------------------
 
 function M.new(opts)
@@ -581,6 +617,14 @@ function Adapter:verify_and_settle(gate, req)
     return nil, errors.SIGNATURE_CONSUMED
   end
 
+  local call_ok, confirmed, confirm_err = pcall(await_confirmation, rpc, signature)
+  if not call_ok then
+    return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirmed)
+  end
+  if not confirmed then
+    return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirm_err)
+  end
+
   -- Settlement response. v1 emits X-PAYMENT-RESPONSE carrying the plain SVM
   -- network slug and the payer pubkey (rust v1 settlement shape
   -- { success, transaction, network, payer }); v2 emits PAYMENT-RESPONSE
@@ -630,6 +674,7 @@ M._private = {
   PAYMENT_IDENTIFIER_KEY       = PAYMENT_IDENTIFIER_KEY,
   caip2_network_for_cluster    = caip2_network_for_cluster,
   legacy_network_slug          = legacy_network_slug,
+  await_confirmation           = await_confirmation,
   LEGACY_PAYMENT_HEADER          = LEGACY_PAYMENT_HEADER,
   LEGACY_PAYMENT_RESPONSE_HEADER = LEGACY_PAYMENT_RESPONSE_HEADER,
 }
--- a/lua/pay_kit/protocols/x402/init.lua
+++ b/lua/pay_kit/protocols/x402/init.lua
@@ -364,13 +364,27 @@ local function build_rpc(config)
   return rpc_mod.new({url = config.rpc_url, transport = rpc_transport.new()})
 end
 
-local function consume_signature(store, signature)
-  if not store then return true end
-  local key = 'x402-svm-exact:consumed:' .. signature
-  if store.put_if_absent then
-    return store:put_if_absent(key)
+local function claim_signature(store, signature)
+  if not store or not store.put_if_absent then return 'claimed' end
+  local prefix = 'x402-svm-exact:' .. signature
+  if store:get(prefix .. ':confirmed') then return 'confirmed' end
+
+  -- A short lease prevents concurrent requests from both returning success,
+  -- while allowing a later request to recover after a crashed or timed-out
+  -- confirmation attempt. The durable consumed marker still records that the
+  -- transaction has been broadcast.
+  if store:put_if_absent(prefix .. ':pending', 15) then
+    store:put_if_absent(prefix .. ':consumed')
+    return 'claimed'
   end
-  return true
+  return 'pending'
+end
+
+local function confirm_signature(store, signature)
+  if not store or not store.put_if_absent then return end
+  local prefix = 'x402-svm-exact:' .. signature
+  store:put_if_absent(prefix .. ':confirmed')
+  store:delete(prefix .. ':pending')
 end
 
 local function sleep_seconds(seconds)
@@ -613,16 +627,20 @@ function Adapter:verify_and_settle(gate, req)
     return nil, errors.INVALID_PROOF .. ': empty broadcast result'
   end
 
-  if not consume_signature(self._store, signature) then
-    return nil, errors.SIGNATURE_CONSUMED
+  local replay_state = claim_signature(self._store, signature)
+  if replay_state == 'pending' then
+    return nil, errors.SIGNATURE_CONSUMED .. ': settlement confirmation is pending'
   end
 
-  local call_ok, confirmed, confirm_err = pcall(await_confirmation, rpc, signature)
-  if not call_ok then
-    return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirmed)
-  end
-  if not confirmed then
-    return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirm_err)
+  if replay_state ~= 'confirmed' then
+    local call_ok, confirmed, confirm_err = pcall(await_confirmation, rpc, signature)
+    if not call_ok then
+      return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirmed)
+    end
+    if not confirmed then
+      return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirm_err)
+    end
+    confirm_signature(self._store, signature)
   end
 
   -- Settlement response. v1 emits X-PAYMENT-RESPONSE carrying the plain SVM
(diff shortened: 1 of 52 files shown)


rust/crates/kit/src/mpp/server/subscription.rs
```diff
@@ -479,9 +487,27 @@ impl SubscriptionServer {
...
+                    let inserted = self
+                        .store
+                        .put_if_absent(&key, serde_json::Value::Bool(true))
+                        .await
+                        .map_err(|e| {
+                            VerificationError::new(format!(
+                                "Failed to reserve activation signature: {e}"
+                            ))
+                        })?;
+                    if !inserted {
```
EfeDurmaz16: P1: This reservation breaks the existing recovery path. If the delegation fetch fails after confirmation, retrying finds the landed activation but rejects its consumed signature. Preserve challenge-bound settlement state so verification can resume without duplicate fulfillment.

lua/pay_kit/protocols/x402/init.lua
```diff
@@ -581,6 +617,14 @@ function Adapter:verify_and_settle(gate, req)
     return nil, errors.SIGNATURE_CONSUMED
   end
 
+  local call_ok, confirmed, confirm_err = pcall(await_confirmation, rpc, signature)
+  if not call_ok then
+    return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirmed)
+  end
+  if not confirmed then
+    return nil, errors.INVALID_PROOF .. ': confirmation failed: ' .. tostring(confirm_err)
```
EfeDurmaz16: P2: The signature is consumed before confirmation. A timeout or RPC error can therefore leave a paid request rejected, with retries returning `SIGNATURE_CONSUMED`. Preserve pending settlement state so retries can resolve the outcome safely.

typescript/packages/mpp/package.json
```diff
@@ -31,11 +31,12 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@solana/addresses": "^6.10.0",
     "@solana-program/compute-budget": "^0.15.0",
     "@solana-program/system": "^0.12.0",
     "@solana-program/token": "^0.11.0",
+    "@solana/addresses": "^6.10.0",
     "@solana/program-client-core": "^6.10.0",
+    "@solana/subscriptions": "^0.5.0",
```
EfeDurmaz16, line 39: P2: Please refresh `harness/pnpm-lock.yaml` for this dependency. Its file-package snapshot omits `@solana/subscriptions`, and the TypeScript harness currently fails at startup with `ERR_MODULE_NOT_FOUND`.

typescript/packages/mpp/src/server/Subscription.ts
```diff
@@ -262,18 +264,22 @@ async function settleActivation(
...
         if (signer) {
+            if (challenge.methodDetails.feePayerKey !== signer.address) {
+                throw new Error('Configured fee-payer signer does not match challenge feePayerKey');
+            }
             txToSend = await coSignBase64Transaction(signer, clientTxBase64);
         }
 
         await simulateTransaction(rpcUrl, txToSend);
         const signature = await broadcastTransaction(rpcUrl, txToSend);
+        if (!(await reserveReplayKey(store, `solana-subscription:consumed:${signature}`))) {
```
EfeDurmaz16: P2: Reserving the signature before confirmation blocks recovery if confirmation or the later delegation fetch fails. A paid retry then returns `Activation signature already consumed`. Preserve pending settlement state so verification can resume safely. The transaction path in `Charge.ts` has the same issue.

EfeDurmaz16 (changes requested):
Payment recovery needs fixing before merge: the new replay checks can reject retries after payment lands but verification fails. The harness also cannot load the new dependency.

Greptile’s recipient-binding concern remains unresolved in both implementations, although it predates this PR. The transfer recipient should be checked against the challenge; delegation amounts alone do not establish who received payment.

typescript/packages/mpp/src/server/Subscription.ts
```diff
@@ -262,18 +264,22 @@ async function settleActivation(
...
+            throw new Error('Activation signature already consumed');
+        }
         await waitForConfirmation(rpcUrl, signature);
```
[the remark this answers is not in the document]
lgalabru: Fixed in dbd06d12. Transaction-mode activation now stores challenge-bound pending/confirmed replay state. Same-challenge retries can rebuild the receipt after confirmation, while different challenges conflict and an active pending lease prevents concurrent duplicate fulfillment. Added recovery coverage.

rust/crates/kit/src/mpp/server/subscription.rs
```diff
@@ -479,9 +487,27 @@ impl SubscriptionServer {
...
+                        return Err(VerificationError::signature_consumed(
+                            "Activation signature already consumed",
+                        ));
+                    }
```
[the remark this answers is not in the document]
lgalabru: Fixed in dbd06d12. Rust activation reservations are now bound to the challenge ID, so an already-landed activation for the same challenge follows the existing delegation recovery path and rebuilds its receipt; reuse from another challenge remains rejected.

rust/crates/kit/src/mpp/server/subscription.rs
lgalabru: Addressed in dbd06d12: Rust replay state is challenge-bound, allowing the landed-activation recovery path to resume for the original challenge while rejecting cross-challenge reuse.

lua/pay_kit/protocols/x402/init.lua
lgalabru: Addressed in dbd06d12. Lua now uses a short pending lease plus a durable consumed marker and confirmed marker. Concurrent duplicates are rejected while pending; timed-out/crashed attempts become recoverable after the lease; confirmed retries rebuild the same settlement response.

typescript/packages/mpp/src/server/Subscription.ts
lgalabru: Fixed in dbd06d12 for both Subscription and Charge transaction mode. Replay records are challenge-bound and distinguish pending from confirmed; active concurrent claims are rejected, confirmed retries can rebuild receipts, and cross-challenge signature reuse remains rejected.

typescript/packages/mpp/src/server/Subscription.ts
```diff
@@ -370,19 +381,77 @@ function validateActivationInstructions(clientTxBase64: string, challenge: Chall
...
+                throw new Error('Multiple initialize_subscription_authority instructions found');
+            sawInitializeAuthority = true;
+        } else if (ix.data[0] === SUBSCRIPTIONS_SUBSCRIBE_DISCRIMINATOR) {
             if (sawSubscribe) throw new Error('Multiple subscribe instructions found');
             sawSubscribe = true;
             subscribeIndex = index;
         } else if (ix.data[0] === SUBSCRIPTIONS_TRANSFER_DISCRIMINATOR) {
             if (sawTransferSubscription) throw new Error('Multiple transfer_subscription instructions found');
             sawTransferSubscription = true;
             transferIndex = index;
```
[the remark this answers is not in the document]
lgalabru: Fixed in dbd06d12. The TypeScript verifier now derives the challenged recipient ATA and requires the transfer receiver account to match it for both the current Codama v0.5 ten-account layout and the previous nine-account client layout. For v0.5 it also binds amount, delegator, and mint from TransferData. Added a wrong-recipient regression test.

typescript/packages/mpp/package.json
lgalabru: Fixed in dbd06d12. The harness lockfile is refreshed and now contains @solana/subscriptions 0.5.0; the harness TypeScript typecheck and @solana/mpp/server startup import both pass locally.

python/src/solana_pay_kit/protocols/x402/__init__.py
```diff
@@ -245,14 +245,15 @@ async def verify_and_settle(self, gate: Gate, request: Any) -> Payment:
...
+            # would let the same signed payment race a later request and produce
+            # multiple fulfillments. This mirrors the durable post-broadcast
+            # reservation in the MPP charge flow.
             try:
                 await rpc.await_confirmation(signature)
             except Exception as exc:  # noqa: BLE001
-                await self._store.delete(replay_key)
                 raise InvalidProofError(
                     f"solana_pay_kit: invalid proof: confirmation failed: {exc}", code="payment_invalid"
                 ) from exc
```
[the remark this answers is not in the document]
lgalabru: Fixed in e6c8762c. Python x402 replay records now distinguish challenge-bound pending and confirmed settlements. Active pending attempts reject concurrent duplicates, an expired pending lease can be atomically claimed by one recovery attempt, and confirmed retries rebuild the same receipt. Added timeout recovery, confirmed retry, and cross-transaction signature-conflict coverage; the full Python suite passes.

python/src/solana_pay_kit/protocols/x402/__init__.py
```diff
@@ -230,13 +233,47 @@ async def verify_and_settle(self, gate: Gate, request: Any) -> Payment:
...
+                            "solana_pay_kit: signature_consumed: settlement confirmation is pending",
+                            code="signature_consumed",
+                        )
+                    recovery_key = f"{replay_key}:recovery:{lease_until}"
+                    if not await self._store.put_if_absent(recovery_key, True):
+                        raise InvalidProofError(
+                            "solana_pay_kit: signature_consumed: settlement recovery is in progress",
+                            code="signature_consumed",
+                        )
+                    await self._store.put(replay_key, replay_record)
```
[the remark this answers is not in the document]
lgalabru: Fixed in 2cedc6b9. The recovery lease is now 21 minutes, longer than the complete worst-case 40 x 30-second RPC confirmation window plus polling delays. After atomically claiming expired recovery, the retry re-reads settlement state and returns an already-confirmed result without overwriting it; only the unchanged expired pending record can be renewed. Added a regression that simulates confirmation racing the recovery claim, plus timeout takeover coverage. Focused tests, Pyright, and Ruff pass.

typescript/packages/mpp/src/server/Subscription.ts
```diff
@@ -254,27 +262,35 @@ async function settleActivation(
...
+                throw new Error('Configured fee-payer signer does not match challenge feePayerKey');
+            }
             txToSend = await coSignBase64Transaction(signer, clientTxBase64);
         }
 
         await simulateTransaction(rpcUrl, txToSend);
         const signature = await broadcastTransaction(rpcUrl, txToSend);
+        const key = `solana-subscription:consumed:${signature}`;
+        const binding = JSON.stringify({ challengeId: credential.challenge.id ?? null, request: challenge });
+        const replayClaim = await claimReplayKey(store, key, binding);
```
EfeDurmaz16: Recovery still runs after simulation and broadcast. Once activation lands, simulating `Subscribe` again fails with `AlreadySubscribed`, before the replay record is checked. Derive the signature and recover the existing settlement first. `Charge.ts` has the same ordering issue when the blockhash expires or the first payment reduces the available balance.

python/src/solana_pay_kit/protocols/x402/__init__.py
```diff
@@ -230,13 +237,64 @@ async def verify_and_settle(self, gate: Gate, request: Any) -> Payment:
...
+                current_record = cast("dict[str, object]", current)
+                if current_record.get("binding") != binding:
+                    raise InvalidProofError("solana_pay_kit: signature_consumed", code="signature_consumed")
+                if current_record.get("state") == "confirmed":
+                    skip_confirmation = True
+                elif current_record.get("state") == "pending" and isinstance(
+                    current_record.get("leaseUntil"), int | float
+                ):
+                    lease_until = float(cast("int | float", current_record["leaseUntil"]))
+                    if lease_until > now:
```
EfeDurmaz16: Every retry rebroadcasts with preflight before checking settlement state. Recovery waits 21 minutes, by which point a normal recent-blockhash transaction has expired, so a landed payment cannot reach recovery. Derive the signature and check the existing settlement before rebroadcasting.

EfeDurmaz16 (changes requested):
Thanks for the fixes, Ludo. The recipient checks and harness dependency issue look addressed. Two recovery issues remain in the inline comments; I’d fix those before merging.

Also, could we revisit the CI series: #260, #262, #264, and #267? They’ve been open since early August, and you already approved #260. The measured full-suite time dropped from about 20.5 to 4.5 minutes. Landing these would save us time on every review cycle. I can refresh the stack against current main if needed.

python/src/solana_pay_kit/protocols/x402/__init__.py
lgalabru: Fixed in 5592bdc5. Python now derives the first transaction signature locally from the exact cosigned wire and checks the replay record before constructing an RPC client or calling send_raw_transaction. Confirmed records return without settlement RPCs; expired records claim the recovery lease and resume confirmation only. The timeout recovery regression now asserts send_calls == 0, and the confirmed-retry test asserts only one RPC instance/broadcast across both requests.

typescript/packages/mpp/src/server/Subscription.ts
lgalabru: Fixed in 5592bdc5. Both Subscription and Charge now derive the deterministic signature from the fully signed wire and inspect challenge-bound settlement state before simulation or broadcast. Confirmed retries skip simulation/broadcast and rebuild the receipt from on-chain state; expired pending retries atomically claim recovery and only resume confirmation. Regression tests assert an identical settled retry performs exactly one total simulate/send across both attempts.

typescript/packages/mpp/src/server/replay.ts
```diff
@@ -0,0 +1,98 @@
...
+type ReplayRecord = {
+    binding: string;
+    leaseUntil: number;
+    state: 'confirmed' | 'pending';
+};
+
+export type ReplayClaim = 'conflict' | 'pending' | 'reserved' | 'retry';
+export type ReplayStatus = 'available' | 'conflict' | 'expired' | 'pending' | 'retry';
+
+const PENDING_LEASE_MS = 45_000;
```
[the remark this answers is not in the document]
lgalabru: Fixed in 3f460511. The TypeScript pending lease is now 21 minutes, matching Python and covering the complete confirmation plus post-confirmation verification window. Focused charge, subscription, and replay tests pass (148/148), along with typecheck, ESLint, and Prettier.

EfeDurmaz16 (approved):
Approved.

Issue #284, opened 2026-08-12
mpp: track the RFC 3339 `expires` divergences, 98 per-tree failures across six SDKs


michael-moffett:
PR #282 adds a shared RFC 3339 `expires` conformance corpus (`harness/vectors/mpp-protocol/expires.json`, 124 vectors) and drives the `expires` parse path in seven of pay-kit's nine SDK trees against it; kotlin and swift carry no runner. Six of the seven trees diverge from the corpus, 98 per-tree failures over 49 distinct vectors; Rust diverges on none. The conformance document reports the same measurement as 120 of 992 cells, counting TypeScript's two call sites separately; this issue counts one row per tree, so TypeScript's 44 assertions appear once as 22. This issue tracks closing them: the per-SDK checklist lists which vectors fail in each tree, and the root-cause section lists the parser fix or policy decision that clears each group. Measurement, parse-path locations and reproduction commands are in [`docs/mpp-expires-rfc3339-conformance.md`](https://github.com/solana-foundation/pay-kit/blob/aed26381/docs/mpp-expires-rfc3339-conformance.md).

## Per-SDK checklist

One checkbox per divergent vector per tree: 98 rows over 49 distinct vectors. Counts are 9 / 16 / 16 / 17 / 18 / 22 / 0 for Python, Go, Lua, PHP, Ruby, TypeScript and Rust. Each row carries the vector name, the input, the direction of the disagreement, and the root cause it belongs to.

`RC-n` is numbering local to this issue, derived bottom-up from the 49 divergent vectors. Every vector lands in exactly one cause, and each tree's rows sum to its measured failure count. It is a different axis from the nine numbered findings in the conformance document and does not map onto them one for one.

### Python: 9 divergent vectors

Parse path: `_parse_rfc3339` (`python/src/solana_pay_kit/protocols/mpp/core/types.py`). Corpus test: `python/tests/test_expires.py`.

- [ ] **RC-1** `jsts_date_time_005` (`1998-12-31T23:59:60Z`): corpus ACCEPT, Python REJECT
- [ ] **RC-1** `jsts_date_time_006` (`1998-12-31T15:59:60.123-08:00`): corpus ACCEPT, Python REJECT
- [ ] **RC-4** `jsts_date_time_015` (`1990-12-31T10:00:00+10:60`): corpus REJECT, Python ACCEPT
[...]

PR #315, opened 2026-09-09
fix(mpp): validate Python expires as RFC 3339

michael-moffett:
## Summary

Addresses the Python issues in #284, landing the same rule as #313 does for TypeScript. `_parse_rfc3339` accepted three things RFC 3339 does not, and rejected one it does:

- **`23:59:60` is now accepted at a UTC month end** and clamped to `23:59:59.999999`, matching the leap-second handling in #313. A leap second anywhere else (`23:58:60`, a non-month-end day) raises.
- **Offset bounds are checked on the regex groups before `fromisoformat`.** This matters: `fromisoformat` normalises `+10:60` to `+11:00` rather than rejecting it, so the lexical check has to come first. `+24:00` and `+00:60` now raise.
- **The `year_0000` vector stays red on Python.** `datetime` cannot represent year 0, so it raises — an unparseable expiry reads as expired, fail-closed. Pinned by a test, so it's a decision rather than a gap. #313 asserts the opposite for TypeScript (`rfc3339.test.ts:44`), so that row of #284 will not agree across the two SDKs.

One function edited in place — no new module, no call-site churn.

## Test Plan

- `pytest python/tests` — **1255 passed, 1 skipped**
- `ruff check src tests` — all checks passed; `ruff format --check` — clean
- `pyright` — 0 errors
- 11 new cases in `TestStrictRFC3339`: leap second accepted at month end (6 offsets), rejected away from one, rejected when it spills past `datetime`'s range, offset out of range (3), and year `0000`.

### DIFF
--- a/python/src/solana_pay_kit/protocols/mpp/core/types.py
+++ b/python/src/solana_pay_kit/protocols/mpp/core/types.py
@@ -53,7 +53,10 @@ def _parse_rfc3339(value: str) -> datetime:
     parsed = datetime.fromisoformat(normalized)
     if leap_second:
         # RFC 3339 §5.7: a leap second only ever ends a UTC month.
-        utc = parsed.astimezone(UTC)
+        try:
+            utc = parsed.astimezone(UTC)
+        except OverflowError as exc:
+            raise ValueError(f"leap second not representable in UTC: {value!r}") from exc
         if (utc.hour, utc.minute) != (23, 59) or utc.day != calendar.monthrange(utc.year, utc.month)[1]:
             raise ValueError(f"leap second not at a UTC month end: {value!r}")
     return parsed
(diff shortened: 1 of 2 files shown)


michael-moffett:
Hey @EfeDurmaz16 Python for the #284 list, same shape as #313.

One row won't line up across the two SDKs: `year_0000` stays red on Python, since `datetime` has no year 0. Pinned by a test.

koriyoshi2041 (approved):
Checked the RFC 3339 boundary changes at 50f6f258. The pre-parse offset guard avoids datetime's normalization of invalid offsets, and the leap-second path preserves fail-closed behavior at both datetime year bounds by converting OverflowError to ValueError.

Local receipts on Python 3.11.14: 33/33 focused expiry tests, changed-file Pyright, Ruff lint, changed-file format, and diff hygiene all pass.

EfeDurmaz16:
@michael-moffett Looks good. The Audit failure comes from the older Hono dependency and is fixed on main by #314. Please rebase onto main and this PR is good to go.

michael-moffett:
@EfeDurmaz16 rebased to main. Thanks for the heads up on the Hono dependency and for the review!

EfeDurmaz16 (approved):

PR #282, opened 2026-08-10
test(mpp): RFC 3339 expires conformance vectors across seven SDKs

michael-moffett:
RFC 3339 `expires` parse vectors for #111: one file, `harness/vectors/mpp-protocol/expires.json`, 124
scenarios. 27 are imported byte-for-byte from the JSON Schema Test Suite (MIT), 27 from Go's
`src/time/format_test.go` (BSD-3-Clause), 70 derived from the RFC text with the deciding section recorded
per row. Licence files sit beside the JSON, same shape as `LICENSE.mpp-tools`.

Every scenario is an `expires` verdict; nothing to filter, nothing to skip. Seven trees, eight parse paths
(typescript has two call sites): 120 of 992 cells diverge, 12.1%.

| parse path | failed of 124 |
|---|---|
| rust | 0 |
| python | 9 |
| go | 16 |
| lua | 16 |
| php | 17 |
| ruby | 18 |
| typescript, charge call site | 22 |
| typescript, session call site | 22 |

Findings, with per-vector evidence in `docs/mpp-expires-rfc3339-conformance.md`:

- `Charge.ts:402` and `Session.ts:356` hand the string to V8's date parser (ECMA-262, not RFC 3339). An
  input with no offset resolves against host local time; six of the seven trees reject it.
- `challenge.go:111-113` returns not-expired for an empty or whitespace-only `expires`; `time.Parse` at
  `:114` is never reached. Suggested fix: `NewChallengeWithSecret` (`wire/challenge.go:63-64`) hardcodes
  `expires: ""` where the defaulting lives in its callers.
- PHP accepts 6 fractional digits, silently truncates 7 to 9, rejects 10 or more.

Wiring: `expires.parse` in the harness loader and one test per SDK tree, each asserting the count it ran
equals the count in the file. Rust's sits in the existing `#[cfg(test)] mod tests` in `challenge.rs`; no
line outside that block changed. The `error_type` on failure, `parse_error`, is a proposal.

The TypeScript test imports `parseRfc3339` from #313 and needs that PR merged first; until then it does not
load. Branch carries main at `c143bfab`. #219 touches the same loader; this change is additive and rebases
cleanly either way. We cannot set labels, so `m2-followup` is left for you.

Prepared with AI assistance. Every imported verdict was independently re-derived from the RFC text by a
reviewer who did not build the corpus: 169 of 170 agreed; the one that did not is recorded in the README as
unresolved provenance and worked through in the conformance doc, not shipped as a vector.

PR #317, opened 2026-09-10
fix(mpp): validate Go expires as RFC 3339

michael-moffett:
Addresses the Go issues in #284.

## Summary

`PaymentChallenge.IsExpired` parsed `expires` with `time.Parse(time.RFC3339, …)`, which is
looser than RFC 3339 §5.6 in two places and stricter in one; against the shared harness
corpus Go diverged on 16 vectors. This closes 15 by putting a §5.6 grammar in front of
`time.Parse`, the same shape as #313 (TypeScript) and #315 (Python):

- accepts lowercase `t` and `z` (§5.6 note)
- rejects a comma secfrac separator
- rejects offsets past ±23:59, which `time.Parse` let through
- accepts `:60` only at a UTC month end (§5.7), clamped to the last instant of `:59`,
  as TypeScript and Python do

Anything refused reads as expired — fail-closed, as before.

**Not changed: an empty `expires` still reads as not expired.** The corpus rejects the
empty string, but `client/charge.go` says the protocol allows omitting it, and flipping
it fails 11 client signing tests. Left as-is; flagged on #284 for a decision.

## Test Plan

- `TestIsExpiredRFC3339Corpus`: 15 corpus vectors + 2 branch rows. On `main` 15/15 fail; here 0.
- `go test ./...` 21 ok · `golangci-lint run` 0 issues · `gofmt -l` empty · coverage 92.1% (floor 91)
- No vector file added; `expires.json` lands with #282.

### DIFF
--- a/go/protocols/mpp/wire/challenge.go
+++ b/go/protocols/mpp/wire/challenge.go
@@ -13,9 +13,15 @@ import (
 	"crypto/sha256"
 	"crypto/subtle"
 	"encoding/json"
+	"fmt"
+	"regexp"
 	"strings"
 	"time"
 )
+
+// time.Parse alone is looser than RFC 3339 §5.6 (comma secfrac, offsets past
+// 23:59) and stricter (no leap second); the grammar gate closes that gap.
+var rfc3339DateTime = regexp.MustCompile(`^\d{4}-\d{2}-\d{2}[Tt]\d{2}:\d{2}:(\d{2})(?:\.\d+)?(?:[Zz]|[+-](\d{2}):(\d{2}))$`)
 
 // PaymentChallenge is sent by a server via WWW-Authenticate.
 type PaymentChallenge struct {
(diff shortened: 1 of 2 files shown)


michael-moffett:
Go for the #284 list, same shape as #313 and #315. 

Empty string left as is: charge.go says omitting expires is allowed, and flipping it breaks the client signing tests. Happy to change it if you'd rather it reject like the shared corpus.

🙏🏻

michael-moffett:
Pushed a58c78e for the leap-second boundary Greptile flagged. - Sorry for missing it the first time. 
Six rows pin 1990-12-31T23:59:59.999999999Z from both sides, Z and -08:00 forms. Test only, +29/−0.

🙏🏻

EfeDurmaz16 (approved):

koriyoshi2041 (approved):
Validated `a58c78ef`: `go test ./protocols/mpp/wire` passes, the touched files are gofmt-clean, and diff hygiene passes. The grammar gate, UTC month-end leap-second check, and nanosecond boundary cases match the TypeScript/Python contract while preserving the intentional empty-expiry behavior.

PR #318, opened 2026-09-11
fix(mpp): validate Ruby expires as RFC 3339

michael-moffett:
Addresses the Ruby issues in #284.

## Summary

`PayCore::Rfc3339Parser.parse` diverged from the shared corpus on 18 vectors. Same rules as #313, #315 and #317:

- `time-secfrac` is `"." 1*DIGIT` — no digit cap; past nanoseconds, truncated not rounded.
- `:60` only at a UTC month end (§5.7), offset applied first, clamped to the last instant of `:59`.
- Offsets past ±23:59 rejected before `Time.iso8601`, which reads `+00:60` as `+01:00`.

Unparseable stays fail-closed. `headers.rb:85` raises on a bad receipt timestamp, so the two rejections above now raise there.

## Test Plan

18 corpus vectors under their corpus names, 0/18 on main → 18/18 here. `bundle exec ruby -Itest test/run.rb`: 452 runs, 0 failures. standardrb and bundle-audit clean.

### DIFF
--- a/ruby/lib/pay_core/rfc3339_parser.rb
+++ b/ruby/lib/pay_core/rfc3339_parser.rb
@@ -8,15 +8,15 @@ module PayCore
   #
   # @see https://datatracker.ietf.org/doc/html/rfc3339 RFC 3339 Date and Time on the Internet
   module Rfc3339Parser
-    # Strict RFC 3339 date-time (sec 5.6) without leap-second support
-    # at the parse layer. Year is exactly 4 digits; T literal accepted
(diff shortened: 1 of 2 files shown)


michael-moffett:
Hi @EfeDurmaz16 Ruby in #318, same three rules as #313/#315/#317, 18/18 on the corpus. I'll take on Lua and PHP next.

EfeDurmaz16 (approved):
Approved, thanks!

PR #319, opened 2026-09-11
feat(tx)!: one transaction path, Transaction V1 first class, legacy dropped

lgalabru:
## Summary

One change across the kit: every Solana transaction the kit builds, signs, encodes, decodes or broadcasts now goes through `core::tx`, legacy transactions are gone, and Transaction V1 ([SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md)) is a first-class message version for servers and clients. Spec side: solana-foundation/x402#11 and the MPP charge/session drafts.

| Commit | What |
|---|---|
| `build(deps)` | SDK 4.x line (solana-message ~4.4, solana-transaction ~4.1, hash 4.4, rpc-client 4.1), keychain `sdk-v4` from keychain main as a git dependency the kit carries itself (so path and git consumers inherit it), spl-token-2022 11, solana-account ^4, litesvm 0.16, surfpool main. Dependency-only; wire bytes unchanged. |
| `fix(clients)` | Every Go, Kotlin, Python and Swift client path builds a v0 message (the harnesses caught x402 upto in Go/Kotlin/Python and MPP charge in Swift still on legacy). |
| `fix(servers)!` | Go, TypeScript, Python, Ruby, PHP and Lua servers reject legacy transactions like the Rust server does; fixtures and conformance runners build v0. |
| `fix(review)` | Greptile findings: skip-preflight fallback through the canonical encoder; settlement packing measures with the configured compute budget; signature credentials are held to the server's accepted versions after `getTransaction`; integration-tests crate migrated to Token-2022 11. Plus `max_tx_version` on every client builder so a signer that cannot sign version 1 (Ledger) negotiates down. |
| `refactor(tx)` | `core::tx` (version, budget, wincode wire, build, envelope policy, canonical RPC send). Every build/sign/encode/decode/broadcast site moved onto it. Legacy dropped: clients build v0, verifiers reject legacy before any instruction check. `transactionVersions` on every challenge/requirements struct. |
| `feat(confidential)` | When the challenge advertises V1, the confidential charge settles in **one transaction**: the proofs are verified inline by the ZK ElGamal Proof program (three for `transfer`, five for `transfer_with_fee`) and Token-2022 reads them by instruction offset. No context-state or record accounts, so nothing to close or sweep. LiteSVM: plain transfer 2,499 bytes / ~238k CU (client requests 400k); fee-bearing transfer_with_fee 3,351 bytes / ~455k CU (five proofs cost 410,300 CU fixed; client requests 700k). Both paths have LiteSVM execution tests. The server allow-list admits context-less ZK verifies and sizes its base64 cap by the largest accepted version. v0 keeps the multi-transaction bundle. Also fixes `getTransaction` read-backs to pass `maxSupportedTransactionVersion: 1` (the bare call was legacy-only). |
| `feat(tx)` | `TxV1Mode { Auto, On, Off }` with a gate probe; servers accept/advertise/build V1 via `with_tx_v1`; `PayKitConfig::tx_v1` (default `Auto`); verifiers hold the V1 header config to the same caps as ComputeBudget instructions; per-version packing caps (18 voucher settlements and 62 reclaims per V1 tx vs 4 and 28); clients build the highest advertised version; LiteSVM end-to-end test. |

## Breaking changes

- **Legacy transactions are rejected by every server in every SDK** (Rust, Go, TypeScript, Python, Ruby, PHP, Lua), each at one shared decode boundary with the same message, and no client in any SDK builds them (Go, Kotlin, Python, Swift and TS all emit v0).
- `Broadcaster::send`, `TxPipeline::submit_verified` / `broadcast_verified`, `PreparedCharge::transaction` take/return `VersionedTransaction`.
- `settlement::packing::{tx_size, would_overflow_tx, pack}` take a `TxVersion` first; `SettlementConfig` gains `tx_version` / `compute_budget` (`with_tx_version` recalibrates the cap).
- `OpenTxOptions` gains `version` and `compute_budget` (derive `Default`).
- `core::signing::{sign_legacy_transaction, cosign_legacy_fee_payer}` removed; the two versioned helpers remain.
- `mpp::protocol::solana::check_transaction_packet_size` → `check_transaction_size(&VersionedTransaction)`.
- `TransactionExpectations` (batch-settlement policy) gains `accepted_versions`; `validate_open_instruction` (upto) takes it too.
- `verify_exact_transaction` takes `&VersionedTransaction`.
- `packing::{tx_size, would_overflow_tx, pack}` take the `Option<&ComputeBudget>` the transaction is built with, so batch boundaries account for the v0 ComputeBudget prefix.
- Client builders take a signer cap: `max_tx_version: Option<TxVersion>` on `BuildChargeTransactionOptions`, `PaymentChannelOpenOptions`, `BuildSubscriptionActivationOptions`, and as a parameter on x402 `build_payment[_header[_v1]]`, `build_upto_payload/header`, `resolve_terms[_with_token_program]`. `core::tx::negotiate` picks the highest version the challenge accepts that the signer can sign; `None` means no cap. The Ledger Solana app signs v0 but not yet v1, so a Ledger-backed client passes `Some(TxVersion::V0)`.

## Behaviour

- Servers default to **version 0 only, no RPC call**. Hosts opt in with `with_tx_v1(TxV1Mode::Auto)` (probe the `enable_tx_v1` account once) or `On`. The gate config defaults to `Auto`. Advertisement (`transactionVersions: [0, 1]`) is emitted only when V1 is accepted, so v0-only servers keep their exact wire shape.
- Operator-signed transactions use the highest accepted version. Client-built transactions use the highest version the challenge advertises, `[0]` when absent.
- **Every existing sanitization check stays.** V1 adds: envelope first (version, no lookup tables, size), header `computeUnitLimit` / per-unit price bounded by the same caps as the v0 instructions, `computeUnitLimit` required, ComputeBudget instructions inside a V1 message rejected.
- Wire encoding is wincode (canonical for all versions; bincode cannot encode V1). Decode rejects non-canonical bytes and trailing data. RPC sends hand the node the base64 string directly because solana-rpc-client still serializes with bincode.
- Confidential charges: one V1 transaction when the challenge advertises `[0, 1]`, the record-chunked v0 bundle otherwise. `ConfidentialTransferParams` gains `tx_version`.

## Publishing caveat

solana-keychain is a git dependency pinned to the commit on keychain main that carries the wincode 0.5 bump. Path and git consumers (pay) inherit it. `cargo publish` strips the git source and resolves crates.io beta.1, which does not compile against SDK 4.1+, so publishing the kit waits for keychain 2.0.0-beta.2.

## Verification

- `cargo test --workspace`: green, including the new LiteSVM V1 test (a 40-transfer V1 transaction that v0 cannot carry builds, signs, round-trips and executes).
- `cargo test -p solana-pay-kit --lib` with every feature except `litesvm-tests`: 1031 pass (the Redis integration test needs `PAY_KIT_TEST_REDIS_URL`).
- `cargo check --workspace --all-targets` and `--all-features`: clean.
- `cargo test -p solana-pay-kit --lib --features confidential,server,client,worker,litesvm-tests`: 1013 pass, including the LiteSVM confidential lifecycle test that now also executes the single V1 transaction and checks it against the server allow-list.
- Test fixtures that built legacy transactions now build v0. No assertion changed except the new legacy-rejection and V1 cases.


lgalabru:
@greptile-apps review

lgalabru:
@greptile-apps please review

lgalabru:
@greptile-apps review

rust/crates/kit/src/mpp/client/confidential.rs
```diff
@@ -538,6 +589,32 @@ async fn build_confidential_transfer_with_fee_bundle(
...
+        let ixs = inline_transfer_with_fee_instructions(
+            token_program,
+            sender_token_account,
+            params.mint,
+            recipient_token_account,
+            &sender_pubkey,
+            &proof_data,
+            &new_decryptable,
+        )?;
+        let budget = ComputeBudget::new(CONFIDENTIAL_INLINE_TRANSFER_COMPUTE_UNIT_LIMIT, 0);
```
EfeDurmaz16: These five proofs alone consume 410,300 CU, before the Token-2022 transfer, so the shared 400,000 limit always fails simulation here. Could we increase the budget and add a fee-bearing V1 execution test?

[Runtime costs](https://github.com/anza-xyz/agave/blob/v4.2.1/programs/zk-elgamal-proof/src/lib.rs#L19): 6,400 + 16,400 + 6,500 + 13,000 + 368,000.

typescript/packages/mpp/src/server/Charge.ts
```diff
@@ -1294,7 +1302,7 @@ async function fetchTransaction(rpcUrl: string, signature: string): Promise<Pars
                 {
                     commitment: 'confirmed',
                     encoding: 'jsonParsed',
-                    maxSupportedTransactionVersion: 0,
+                    maxSupportedTransactionVersion: 1,
```
EfeDurmaz16, line 1309: Non-blocking follow-up, also noted by the bot: signature credentials still accept confirmed legacy transactions in TypeScript, Go, and Python, while transaction-byte credentials reject them. Could we check the returned transaction version before consuming the signature or issuing a receipt, and cover signature-mode legacy rejection?

EfeDurmaz16 (commented):
Thanks! One fix is needed for the fee-bearing confidential V1 path below. The legacy signature check is a non-blocking follow-up.

Also, could we remove `harness/go-server/go-server`? It is a 15 MB compiled macOS binary that appears to have been committed by accident.

rust/crates/kit/src/mpp/client/confidential.rs
lgalabru: Good catch, thanks. Measured on LiteSVM: the fee-bearing single transaction is 455,393 CU / 3,351 bytes (plain: 238,057 / 2,499). The two paths now have separate limits (400k plain, 700k with fee) and there is a fee-bearing LiteSVM execution test next to the plain one (`inline_v1_confidential_transfer_with_fee_executes_in_litesvm`), which also checks the recipient recovers amount minus fee.

typescript/packages/mpp/src/server/Charge.ts
lgalabru: Done in the same push: TypeScript, Go and Python now check the reported version before consuming the signature or issuing a receipt (legacy and missing versions rejected, 0/1 accepted, same messages as Rust's `check_reported_version`), with signature-mode legacy rejection tests per path. Go's fetch also asks for maxSupportedTransactionVersion 1 now. The stray go-server binary is removed in the commit before this one.

lgalabru:
@greptile-apps review

EfeDurmaz16 (approved):
Looks good, thanks! Non-blocking follow-up: could we enable litesvm-tests in CI so the new fee-bearing V1 execution test runs there too?

PR #320, opened 2026-09-11
fix(mpp): validate PHP expires as RFC 3339

michael-moffett:
## Summary

Conform `Rfc3339Parser` to the #284 vectors (17 added, all passing).

- Leap second: `:60` parses only at the last second of a UTC month (RFC 3339 §5.7); previously every `:60` was downshifted to `:59`. Offsets are converted to UTC before the check.
- Fractional seconds: accept `1*DIGIT` instead of capping at 9; truncated to microseconds as before.
- `year_0000` is rejected (already caught by `checkdate`, now pinned by a test).
- The three `createFromFormat` attempts move into one `createFromAtom()` helper.

## Test Plan

- `composer run lint` — syntax, format, phpstan clean
- `composer run test:coverage` — 448 tests pass, statement coverage 92.3 % (floor 91)
- Vectors mirror the Go/Ruby/Python rows from #284


michael-moffett:
@EfeDurmaz16 PHP row of #284 up as #320. Lua next
🙏🏻

EfeDurmaz16 (approved):
Looks good, thanks!

PR #321, opened 2026-09-11
fix(mpp): validate Lua expires as RFC 3339

michael-moffett:
## Summary

Conform `expires.parse_rfc3339` to the #284 vectors (16 added, all passing).

- Leap second: `:60` parses only at 23:59:60 UTC on a month's last day (RFC 3339 §5.7); the offset is applied before the check, so offset forms of a real leap instant pass and non-instants are rejected.
- Fractional seconds: accept `1*DIGIT` instead of capping at 9; truncated to whole seconds as before.
- One leap rule at one site; no other parser behaviour changed.

## Test Plan

- `luacheck pay_kit/ plugins/ tests/` — 0 warnings / 0 errors
- `busted` via `tests/run.lua` — 623 pass / 1 skip / 0 fail, line coverage 90.75 % (floor 90)
- Vectors mirror the Go/Ruby/Python/PHP rows from #284

### DIFF
--- a/lua/pay_kit/protocols/mpp/expires.lua
+++ b/lua/pay_kit/protocols/mpp/expires.lua
@@ -21,33 +21,43 @@ local function days_in_month(year, month)
   return 31
 end
 
+-- RFC 3339 sec 5.7: `:60` is inserted only at 23:59:60 UTC on a month's last day. Offset first, so the check reads the instant denoted.
+local function is_leap_second_instant(year, month, day, hour, min, offset_secs)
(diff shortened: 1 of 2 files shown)


michael-moffett:
@EfeDurmaz16 Lua row of #284 is up as #321, same shape as Go/Ruby/PHP. 
I think that closes the language rows on #284.

Thanks for the all the quick views and approvals!

EfeDurmaz16 (approved):
Looks good, thanks!

PR #328, opened 2026-09-19
fix(mpp): send preflightCommitment with sendTransaction so confirmed-only blockhashes are accepted

seekdaseek:
`broadcastTransaction` sends `sendTransaction` without `preflightCommitment`, so preflight runs against the finalized bank while the challenge blockhash is minted at `confirmed`. For roughly the first 12 seconds of a blockhash's life the preflight bank has never seen it, and the RPC rejects the transaction with `Blockhash not found`.

The failure is deterministic and inverted from what you would guess: a client that pays **fast** always fails, and one that dawdles past finality succeeds.

## The mismatch

Three RPC calls in one settlement path, one of which disagrees:

| step | call | commitment | file |
|---|---|---|---|
| challenge blockhash | `getLatestBlockhash` | `confirmed` | `typescript/packages/mpp/src/server/Charge.ts:182` |
| pre-broadcast check | `simulateTransaction` | `confirmed` | `typescript/packages/mpp/src/server/Charge.ts:1327` |
| broadcast | `sendTransaction` | **unset** | `typescript/packages/mpp/src/server/Charge.ts:1352` |

Solana's RPC reference for `sendTransaction`: *"preflightCommitment — Commitment level to use for preflight when `skipPreflight` is `false`. Default `finalized`."* `skipPreflight` itself defaults to `false`. In agave, `rpc/src/rpc.rs` passes `None` when `preflight_commitment` is unset, `bank()` resolves it with `commitment.unwrap_or_default()`, and `solana-sdk`'s `CommitmentLevel` derives `#[default] Finalized`.

The same chain in agave on `master`, by function name: `fn send_transaction`
passes `commitment: None` into `get_bank_with_config` when `preflight_commitment`
is unset and `skip_preflight` is `false`; `fn get_bank_with_config` calls
`self.bank(commitment)`; `fn bank` does `commitment.unwrap_or_default()`; and in
`anza-xyz/solana-sdk`, `commitment-config/src/lib.rs`, `CommitmentLevel` derives
`Default` with `#[default]` on `Finalized`.

Because `simulateTransaction` is explicitly at `confirmed` (`Charge.ts:1327`) it passes, so the mismatch never surfaces until the send step.

It is also invisible in sandbox: Surfpool stamps a synthetic always-valid blockhash (`SURFPOOL_BLOCKHASH_PREFIX = 'SURFNETxSAFEHASH'`, `typescript/packages/mpp/src/server/network-check.ts:25`) which has no finality semantics, so the confirmed/finalized split does not exist there.

## Every affected call site

A later measurement corrected this section. The first version of this table
listed all seven sites as broken; that overstated it. `@solana/kit` 6.10.0's
`createSolanaRpc` injects `preflightCommitment: 'confirmed'` itself, so any
site that goes through a kit client was already sending it. Measured against a
capturing HTTP listener on kit 6.10.0, before any change in this PR:

```
caller passed omitted    -> wire: {"encoding":"base64","preflightCommitment":"confirmed"}
caller passed processed  -> wire: {"encoding":"base64","preflightCommitment":"processed"}
caller passed confirmed  -> wire: {"encoding":"base64","preflightCommitment":"confirmed"}
caller passed finalized  -> wire: {"encoding":"base64"}
```

So the sites divide into three groups.

**Genuinely broken — build the JSON-RPC body by hand, no kit involved:**

| file | site | blockhash minted at |
|---|---|---|
| `server/Charge.ts:1352` | `broadcastTransaction` | `confirmed` (`Charge.ts:181`), simulates at `confirmed` (`:1327`) |
| `server/Subscription.ts:1048` | `broadcastTransaction` | `confirmed` (`Subscription.ts:187`), simulates at `confirmed` (`:1023`) |

These two are the fix. They are also the two the added test assertions cover.

**Defensive — the RPC client is injected, so the default is the caller's:**

| file | site |
|---|---|
| `server/session/on-chain.ts:790` | `submitTopUpTx` |
| `server/session/on-chain.ts:944` | `submitOpenTx` |
| `server/session/on-chain.ts:1055` | `submitSettleAndDistribute` |

All three take `args.rpc`, typed `RpcLike` (`Session.ts:1437`), which is
either a `createSolanaRpc` client *or* a custom `SubmitOpenRpc` duck-type. A
kit client already defaults; a custom one need not. They also build on
`buildAndSignWireTransaction`, which mints at `confirmed`
(`session/wire-tx.ts:37`), so the mismatch is reachable with a
non-kit client.

**No-op, kept as explicit intent — already defaulted by kit:**

| file | site |
|---|---|
| `client/Charge.ts:128` | push mode broadcast |
| `client/Subscription.ts:222, :487` | activation, authority init |

These construct their own client with `createSolanaRpc` (`client/Charge.ts:117,189`,
`client/Subscription.ts:217,282,452,506`), so `preflightCommitment: 'confirmed'`
was already on the wire. Writing it out changes no behaviour; it is kept so the
requirement is visible at the call site rather than inherited silently.

`Session.ts` itself only declares the `sendTransaction` shape and duck-types an
injected client, so it reads as unaffected — but the calls it delegates to are
the three in `session/on-chain.ts` above.

`skipPreflight` is unchanged everywhere. `submitSettleAndDistribute` passed no
`skipPreflight` at all and still does not: the default is `false`, so preflight
was already on there and only needed aiming.

## Why it is duplicated

The TypeScript implementation inlines `simulateTransaction` and
`broadcastTransaction` per module — `server/Charge.ts` and
`server/Subscription.ts` carry near byte-identical copies — so one omission
became seven. `python/src/solana_pay_kit/_paycore/rpc.py` and
`go/paycore/solanatx/solanatx.go` keep the equivalent in a shared RPC helper,
which is why neither carries it twice. Folding these into one helper looks
worthwhile as a follow-up; deliberately not done here, to keep this PR to the
one-line shape.

## Measured

Same transaction, same freshly minted `confirmed` blockhash, simulated against mainnet at both commitments:

```
commitment=confirmed   value.err=null
commitment=finalized   value.err="BlockhashNotFound"

confirmed slot : 448354281
finalized slot : 448354250
finality lag   : 31 slots (~12.4s)
```

## Proof on mainnet

Found while integrating `@solana/mpp` 0.7.0 into **x402.ochinimus.app**, a live paid API, so the 402 can be verified directly rather than taken on trust.

Before the change, a mainnet payment was rejected 3.7 s after the challenge was issued — inside the 12.4 s window — with nothing wrong with the credential: `transferChecked`, 1000 base units of USDC (`EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`), the correct derived recipient ATA, validly signed, and using the server-supplied `recentBlockhash` as §7.2 recommends.

With this one-line change applied to the installed package, the same flow settled:

```
signature : 3jZnZQ4Qxx1oGrWv3jj2sKz5oGW6EKSgBBwVmthfpo48jVkSMdrW1yphpaapEMHtYcAzbMEmgQbZFUn446xYxCg3
slot      : 448360973
blockTime : 2026-09-19T09:01:29Z
fee       : 5001 lamports, paid by the payer (feePayer:false)
status    : finalized, meta.err = null
```

Verified from a public RPC independent of the one the server used.

## The kit already does this in its other implementations

TypeScript is the only one that omits it:

| language | file | line |
|---|---|---|
| PHP | `php/src/Protocols/Mpp/Server/SolanaChargeHandler.php` | 283 |
| Python | `python/src/solana_pay_kit/_paycore/rpc.py` | 107 |
| Go | `go/paycore/solanatx/solanatx.go` | 317 |
| **TypeScript** | `typescript/packages/mpp/src/server/Charge.ts` | **1352** |

The PHP one is the MPP charge handler itself, so this is an inconsistency inside one protocol implementation rather than a difference between unrelated components.

The Rust spine does it deliberately too: `rust/crates/kit/src/core/rpc.rs:64`,
`pub fn preflight_config(rpc: &RpcClient)` sets
`preflight_commitment: Some(rpc.commitment().commitment)`, and
`send_and_confirm_transaction` uses it (`:117`). The module header names the
same failure this PR fixes — flows that "can hit false-negative preflight
simulations when the RPC bank lags a confirmed dependency" (`:9-13`).

## Why not `skipPreflight: true`

That is the obvious workaround for this failure, and it is worth naming so it is not reached for: it discards the simulation the spec relies on to catch invalid transactions before fees are spent. §11.1 makes that simulation mandatory in fee-payer mode — *"If feePayer is true, simulate the transaction using the simulateTransaction RPC method."* — and only a SHOULD otherwise, so the cost of skipping it lands hardest exactly where the server is paying. This change keeps `skipPreflight: false` and aims the preflight at the bank the blockhash actually came from.

## Affected versions

| | version | published |
|---|---|---|
| npm `latest` | 0.7.0 | 2026-07-13 |
| repo HEAD | 0.11.0 | unpublished |

`npm view @solana/mpp versions` lists `0.0.0` through `0.7.0`; **0.8 through 0.11 were never published**. So the published package predates `draft-solana-charge-00` (published 19 September 2026) and carries the omission, and HEAD carries it too — the line is identical in `dist/server/Charge.js:982` of 0.7.0 and in `src/server/Charge.ts:1352` at HEAD.

`gh search issues` and `gh search prs` for `preflightCommitment` in this repo both return zero results, so this does not appear to have been reported.

## Spec alignment

§12.5, on the pull path this function serves: *"For type="transaction" credentials, the server controls the broadcast and confirmation process. Servers MUST wait for at least confirmed commitment before returning the receipt."*

`confirmed` is therefore the commitment the settlement path is already specified around. This change makes the three RPC calls agree with each other and with that clause rather than loosening anything.

An alternative fix would be to mint the challenge blockhash at `finalized`, but that is strictly worse: it spends ~12 s of the blockhash's ~60 s validity before the client ever sees it.

## Verification

Run in this repo with the change applied, from `typescript/`:

```
pnpm install --frozen-lockfile     ok
pnpm --filter @solana/mpp build    ok
pnpm lint                          ok
pnpm format:check                  All matched files use Prettier code style!
pnpm typecheck                     ok (packages/mpp, packages/pay-kit)
pnpm vitest run packages/mpp       30 files, 468 tests passed
```

The surfpool integration leg was not run locally: it needs `SURFPOOL_DATASOURCE_RPC_URL`, which CI supplies as a secret.


typescript/packages/mpp/src/server/Charge.ts
```diff
@@ -1349,7 +1349,7 @@ async function broadcastTransaction(rpcUrl: string, base64Tx: string): Promise<s
             id: 1,
             jsonrpc: '2.0',
             method: 'sendTransaction',
-            params: [base64Tx, { encoding: 'base64', skipPreflight: false }],
+            params: [base64Tx, { encoding: 'base64', preflightCommitment: 'confirmed', skipPreflight: false }],
```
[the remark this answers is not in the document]
seekdaseek: Confirmed — the subscription path carries the same mismatch: `server/Subscription.ts` mints its blockhash at `confirmed` (`:187`), simulates at `confirmed` (`:1023`), then broadcasts with no `preflightCommitment` (`:1048`). It is now fixed in this PR.

Sweeping the rest of the package turned up four more sites with the same shape, all fixed in the same commit: `submitTopUpTx`, `submitOpenTx` and `submitSettleAndDistribute` in `server/session/on-chain.ts`, which build on `buildAndSignWireTransaction` (`session/wire-tx.ts:37`, also `confirmed`), plus the client broadcasts in `client/Charge.ts` and `client/Subscription.ts`. `Session.ts` only declares the `sendTransaction` shape and duck-types an injected RPC client, so it reads as unaffected — but the calls it delegates to are those three in `session/on-chain.ts`.

`skipPreflight` is unchanged everywhere. Verified with `pnpm --filter @solana/mpp build`, `pnpm lint`, `pnpm format:check`, `pnpm typecheck` and `pnpm test` — 42 files, 561 tests, all passing.

EfeDurmaz16 (commented):
The fix is correct. The RPC node falls back to `finalized` when `preflightCommitment` is unset, while the charge and subscription paths fetch the blockhash, simulate, confirm and verify at `confirmed`, so preflight was the only outlier. It also matches the Rust spine (`preflight_config` in `rust/crates/kit/src/core/rpc.rs`). Two notes inline.

Both commits are unsigned, so GitHub shows them as Unverified, and `main` has a required-signatures rule. Please sign both and force-push.

Verified against the agave RPC source and by running the installed @solana/kit 6.10.0. Reproduced on mainnet: a fresh `confirmed` blockhash is valid at `confirmed` and not valid at `finalized` (`isBlockhashValid`), finalized was 31 slots behind. CI has not run yet, the workflows are waiting for maintainer approval.

typescript/packages/mpp/src/server/Charge.ts
EfeDurmaz16, line 1352: Nothing pins this option, and Surfpool cannot see the bug because its blockhash is synthetic, so a unit assertion is the only guard against it being dropped again. The `sendTransaction` branch of `mockServerBroadcastFetch` in `charge.test.ts` (:1776) can capture `body.params[1]` and assert `preflightCommitment` is `confirmed`. The `sendTransaction` case in `subscription-server.test.ts` (:715) can do the same for the subscription broadcaster.

typescript/packages/mpp/src/client/Charge.ts
```diff
@@ -127,6 +127,7 @@ export function charge(parameters: charge.Parameters) {
                 const signature = await rpc
                     .sendTransaction(encodedTx, {
                         encoding: 'base64',
+                        preflightCommitment: 'confirmed',
```
EfeDurmaz16, line 130: This edit does not change the request. `createSolanaRpc` in @solana/kit 6.10.0 already adds `preflightCommitment: 'confirmed'` to `sendTransaction` through its default-commitment transformer. I ran it with a capturing transport and the old params go out as `{"encoding":"base64","skipPreflight":false,"preflightCommitment":"confirmed"}`. So the client sites never hit the bug, only the two raw `fetch` broadcasters did, and the "Every affected call site" table in the description overstates it. Fine to keep as explicit intent. The three `session/on-chain.ts` sites are different: `args.rpc` is injected, so there the explicit option protects an rpc that does not default to `confirmed`.

seekdaseek: Measured it rather than taking the client sites on trust, and you are right. Against a capturing HTTP listener on @solana/kit 6.10.0, `createSolanaRpc` injects `preflightCommitment` regardless of what the caller passes: omitted goes out as `confirmed`, `processed` passes through, `confirmed` passes through, and `finalized` is dropped entirely. The client sites build their own client (`client/Charge.ts:117,189`, `client/Subscription.ts:217,282,452,506`), so they were never broken.

The description's table is corrected into three groups: the two raw `fetch` broadcasters that were genuinely broken, the three `session/on-chain.ts` sites where `args.rpc` is injected as `RpcLike` and may be a non-kit duck-type so the option is defensive, and the client sites where it is a no-op kept as explicit intent. The body says the earlier table overstated it and gives the measurement that corrected it. Code unchanged, per your note that keeping it is fine.

Assertions added at both sites you named. `charge.test.ts` and `subscription-server.test.ts` now capture `params[1]` and assert `preflightCommitment: 'confirmed'` with `skipPreflight: false`. Removing the option fails 12 tests in the first and 2 in the second.

Also confirmed the Rust spine by opening it: `preflight_config` at `rust/crates/kit/src/core/rpc.rs:64` sets `preflight_commitment: Some(rpc.commitment().commitment)`, used by `send_and_confirm_transaction` at `:117`, and the module header at `:9-13` names the same failure — flows that "can hit false-negative preflight simulations when the RPC bank lags a confirmed dependency".

All three commits are now signed and show Verified. Full suite green: build, lint, format:check, typecheck and 561 tests across 42 files.

seekdaseek:
Following up with a live mainnet settlement through the patched path.

**Signature:** [`3mbbU85h3JLxnuQbq2YvXr3hjvLsixYyvUYuPgrJj3jWidkVJyWdVBGE7iUcN6CN6VsMUqUDhWk9ag1GuSY9KD2g`](https://solscan.io/tx/3mbbU85h3JLxnuQbq2YvXr3hjvLsixYyvUYuPgrJj3jWidkVJyWdVBGE7iUcN6CN6VsMUqUDhWk9ag1GuSY9KD2g)

Server-side `@solana/mpp@0.7.0` with this patch applied, `feePayer: false`, `type="transaction"` credential so the server broadcasts. Verified on chain through the public RPC rather than the server's own: slot 448534662, `err: null`, fee 5001 lamports paid by the payer, USDC 0.099 → 0.098 on the payer and 0.033532 → 0.034532 on the recipient. HTTP 200 on the retry, and a replay of the same signature is refused with `verification-failed` / "Signature already consumed".

**On what this does and does not prove, since I would rather be precise than overstate the PR.**

It does not prove the broadcast took the patched line. I did not capture the outgoing `sendTransaction` params, and by the time I went looking the challenge blockhash had already expired, so the finalized-bank state at broadcast time cannot be reconstructed retrospectively. What I can say is narrower: the file on the server carries the patch, the process loaded that file, and the payment succeeded roughly 2.9s after the challenge was minted at `confirmed` — which is inside the window this PR is about.

I also tried to build the negative control properly and could not, for a reason worth recording. Reverting the patch in a scratch tree does not produce a failing payment, because the layer's own boot assertion refuses to start at all on an unpatched dependency, so no challenge is served and there is nothing to pay. Bypassing that assertion does serve a live unpatched challenge, but the scratch host then advertises an `https://localhost:<port>` resource URL with no TLS listener, so the client cannot complete against it either. That is a property of tunnelling a scratch port, not of the patch.

So: a real settlement landed, and the timing is consistent with the fix mattering, but the direct observation is still missing. If a maintainer wants the airtight version, the cheapest shape is probably a unit-level test that calls `broadcastTransaction` twice with one freshly-signed transaction — once patched, once not — against a blockhash minted at `confirmed`, and asserts the unpatched call returns "Blockhash not found".

seekdaseek:
Second settlement, this time through the `pay` CLI itself rather than a hand-rolled client, against the other gated route.

**Signature:** [`2TFZpJd4ZijoFvNJbBjEohpAxQegHk27pCFmcQCcqHk3TPnad397iTK2DH3qLnyxuTjVvV84bUnT5RMv3R9VfGdD`](https://solscan.io/tx/2TFZpJd4ZijoFvNJbBjEohpAxQegHk27pCFmcQCcqHk3TPnad397iTK2DH3qLnyxuTjVvV84bUnT5RMv3R9VfGdD)

`pay fetch --account <mainnet account> https://x402.ochinimus.app/api/btc-price`. Verified through the public RPC: slot 448535554, `err: null`, fee 5001 lamports paid by the payer, USDC 0.098 → 0.097 on the payer and 0.034532 → 0.035532 on the recipient. Server recorded it `settled` with the signature in its audit row.

Worth recording for anyone else automating the CLI: a mainnet account with `auth_required: true` blocks in `pay_keystore::macos::TouchId::AuthGate::authenticate`, which shells out to `~/.cache/pay/pay.sh authenticate` and polls for a biometric approval. In a headless or background context that never resolves and the process holds zero TCP sockets, so it looks like a network hang but never reaches the network at all. `sample <pid>` names it immediately. With the prompt approved interactively the same command completes in seconds.

The caveat in my previous comment stands unchanged: neither settlement demonstrates the broadcast took the patched line, only that payments in this window succeed against a patched server.

seekdaseek:
@EfeDurmaz16 both of your notes are in as of 04a386b — the three commits are signed and show Verified, and the assertions sit at the two raw broadcasters you named, capturing `params[1]` and asserting `preflightCommitment: 'confirmed'` with `skipPreflight` still `false`.

This is now blocked on two things that need write access rather than on code: 9 workflow runs are still awaiting maintainer approval, so CI has never run against this branch, and one approving review. If you are able to release the runs, the checks can make the case instead of my screenshots.

One sequencing note for whoever picks it up: solana-foundation/surfpool#813 and this PR are the same defect at two layers, and that description says it "hid a deterministic mainnet failure from pay-kit's Surfpool CI" — which is why this repo's own sandbox cannot currently reproduce the bug. Once #813 lands I am happy to add the surfpool integration test here. It would fail without this patch and pass with it, which is the negative control I could not build against a live host.

EfeDurmaz16:
LGTM, thanks! Could you shorten the PR description to the problem, fix, and tests? It’s quite so long for this change. I opened #813 already and will merge it when I get a chance :)

EfeDurmaz16 (approved):

EfeDurmaz16:
Surfpool https://github.com/solana-foundation/surfpool/pull/813 is now merged. I ran all 8 MPP integration tests against that exact commit (`dc825309`), and they passed. Removing this fix makes the pull-payment test fail with `Blockhash not found`, so the regression is now reproducible and covered. Thanks!

